PR #2, opened 2016-03-08
Improve check_linux_ulimit

hasegeli:


check_linux_ulimit.py
```diff
@@ -0,0 +1,150 @@
...
+
+        elif soft_limit <= num_fds:
+            state = 2
+            procname = get_proc_name(pid)
+            msg = 'PID {0} [{1}] reached its soft limit (open: {2}, limit {3})\n'.format(
+                pid, procname, num_fds, soft_limit
+            )
+
+        elif (soft_limit * warning / 100) <= num_fds:
+            if state != 2:
```
datenbrille: What is the intention of this if statement?

hasegeli: The intention is not to decrease state to warning when it is already critical.  It could have been written nicer, though.  I rewrote this function, too.  How is it looking now?

datenbrille:
Yeah, more readable with the Enums. Thanks

PR #6, opened 2016-05-26
increases thresholds after changing the usage to persistent beanstalk connections

lacunoc:
we are now using persistent beanstalk connections between web -> db and therefore we got many more connections.


hasegeli:
We made the the script configurable.

PR #14, opened 2016-09-08
Fix warning on check_linux_ulimit.py

hasegeli:


juergenthomann:
Now it fails if it should give a warning or error message:
`
Traceback (most recent call last):
  File "/usr/lib/nagios/igmonplugins/check_linux_ulimit.py", line 189, in <module>
    main()
  File "/usr/lib/nagios/igmonplugins/check_linux_ulimit.py", line 75, in main
    exit(*get_state(float(args.warning) / 100.0))
  File "/usr/lib/nagios/igmonplugins/check_linux_ulimit.py", line 110, in get_state
    .format(pid, get_proc_name(pid), num_fds)
  File "/usr/lib/nagios/igmonplugins/check_linux_ulimit.py", line 151, in get_proc_name
    return cmdline.split('\x00')[0] if cmdline else 'unknown'
AttributeError: 'list' object has no attribute 'split'
`

hasegeli:
Fixed

juergenthomann:
Now it works.

PR #15, opened 2016-10-19
Add nagios check for private repositories (not forks) in bitbucket 

TheFRedFox:
Add nagios check for private repositories (not forks) in bitbucket server

 * OK if no private repositories (not forks) were found
 * WARNING if private repositories (not forks) were found

### DIFF
--- a/bitbucket_private_repository.py
+++ b/bitbucket_private_repository.py
@@ -87,7 +87,7 @@ def main(args):
     repositories = fetch_repositories(base_url, auth)
     private_repos = [
         repo for repo in repositories
-        if 'origin' not in repo and repo['project']['type'] != 'PERSONAL'
+        if 'origin' not in repo and repo['project']['type'] == 'PERSONAL'
         ]
 
     if not private_repos:
@@ -133,19 +133,18 @@ def fetch_repositories(base_url, auth=None):
         :rtype: list of dict
     """
     endpoint = '/rest/api/1.0/repos'
-    limit = 1000
+    limit = 1000  # right now bitbucket limit the request with a default of 1000
     start = 0
     params = {'limit': limit, 'start': start}
-    repositories = []
     last_page = False
     while not last_page:
         response = do_request('get', base_url, endpoint, params=params,
                               auth=auth)
         response = json.loads(response.text)
-        repositories.extend(response['values'])
+        for repository in response['values']:
+            yield repository
         params['start'] += limit
         last_page = response['isLastPage']
-    return repositories
 
 
 def do_request(method, base_url, endpoint, params={}, auth=None):
--- a/bitbucket_private_repository.py
+++ b/bitbucket_private_repository.py
@@ -1,0 +1,176 @@
+#!/usr/bin/env python
+#
+# InnoGames Monitoring Plugins - bitbucket_private_repository.py
+#
+# This is a Nagios script which checks, if there are any private repositories
+# in a Bitbucket system which are not forks. The script uses the Bitbucket
+# Server Rest Api and it is possible to use Basic or two-legged OAuth
+# authentication.
+# The script will exit with:
+#  - 0 (OK) if there are no private (none fork) repositories
+#  - 1 (WARNING) if there are private (none fork) repositories
+#
+# Copyright (c) 2016, InnoGames GmbH
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+
+# Needed modules which are imported below
+#
+# For OAuth authentication, lines 150-151
+# pycrypto, pip install pycrypto
+# requests_oauthlib, pip install requests requests_oauthlib
+#
+
+from __future__ import print_function
+
+import json
+from argparse import ArgumentParser
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--base-url', dest='base_url',
+                        help='the base url of the application you want to '
+                             'check (e.g. https://sub.example.com)')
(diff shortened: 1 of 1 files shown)


bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+    parser = ArgumentParser()
+    parser.add_argument('--base-url', dest='base_url',
+                        help='the base url of the application you want to '
+                             'check (e.g. https://sub.example.com)')
+    parser.add_argument('--auth', dest='auth', choices=['basic', 'oauth'],
+                        help=('authentication mode to use. basic uses '
+                              'username and password. oauth uses private key '
+                              '(with/out passphrase) and consumer key. '
+                              'without --auth script will try an anonymous '
+                              'access.')
```
hasegeli: You haven't put parentheses around the previous help.  They are not necessary in here either.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+import requests
+from requests.auth import HTTPBasicAuth
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--base-url', dest='base_url',
+                        help='the base url of the application you want to '
+                             'check (e.g. https://sub.example.com)')
+    parser.add_argument('--auth', dest='auth', choices=['basic', 'oauth'],
```
hasegeli: `choices` can be a tuple.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+                        help=('authentication mode to use. basic uses '
+                              'username and password. oauth uses private key '
+                              '(with/out passphrase) and consumer key. '
+                              'without --auth script will try an anonymous '
+                              'access.')
+                        )
+    parser.add_argument('--username', dest='username',
```
hasegeli: The `dest` is unnecessary.  `argparse` makes the variable names from the argument names.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+
+    auth = parse_auth_argument(args)
+
+    repositories = fetch_repositories(base_url, auth)
+    private_repos = []
+    for repository in repositories:
+        if 'origin' in repository:
+            continue
+
+        if not repository['project']['type'] == 'PERSONAL':
```
hasegeli: Better use `!=`

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+            continue
+
+        private_repos.append(repository)
+        # print(json.dumps(repository))
```
hasegeli: I would write this like:
```python
private_repos = [
    r for r in repositories
    if 'origin' not in r and r['project']['type'] != 'PERSONAL'
]
```

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+        exit(0)
+
+    if args.format:
+        format_string = args.format
+    else:
+        format_string = ('{repo[project][owner][name]}: {repo[name]} '
+                         '({repo[links][self][0][href]})\n')
+    string = ''
+    for repository in private_repos:
+        string += format_string.format(repo=repository)
```
hasegeli: I would write this like
```python
''.join(format_string.format(repo=r) for r in private_repos)
```

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+          .format(string, amount=len(private_repos)))
+    exit(2)
+
+
+def parse_auth_argument(args):
+    auth = args.auth
+    if auth == 'basic':
+        if not (args.username and args.password):
+            print(('For basic authentication, \'username\' and \'password\' '
+                   'parameter are needed'))
```
hasegeli: Extra paranheses

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+    print('WARNING: {amount} private repositories found:\n{0}'
+          .format(string, amount=len(private_repos)))
+    exit(2)
+
+
+def parse_auth_argument(args):
+    auth = args.auth
+    if auth == 'basic':
+        if not (args.username and args.password):
+            print(('For basic authentication, \'username\' and \'password\' '
```
hasegeli: PEP 8 says to use double quotes when there are single quotes inside to avoid escaping.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+    repositories = []
+    last_page = False
+    while not last_page:
+        response = do_request('get', base_url, endpoint, params=params,
+                              auth=auth)
+        response = json.loads(response.text)
+        repositories.extend(response['values'])
+        params['start'] += limit
+        last_page = response['isLastPage']
+    return repositories
```
hasegeli: It is a good idea to to `yield` them instead of materialising in an array as you are bothering to fetch them in batches.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+
+def do_request(method, base_url, endpoint, params={}, auth=None):
+    return requests.request(method, base_url + endpoint, auth=auth,
+                            params=params)
+
+
+def get_oauth1session(consumer_key, consumer_secret, private_key,
+                      passphrase):
+    from Crypto.PublicKey import RSA
+    from requests_oauthlib import OAuth1
```
hasegeli: It is better to `import` things on the beginning of the file.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+    f = open(private_key, 'r')
```
hasegeli: `with open(private_key) as fd:`

bitbucket_private_repository.py
```diff
@@ -0,0 +1,176 @@
...
+    rsa_key = RSA.importKey(f.read(), passphrase)
+
+    return OAuth1(client_key=consumer_key, client_secret=consumer_secret,
+                  signature_method='RSA-SHA1', rsa_key=rsa_key)
+
+
+if __name__ == "__main__":
```
hasegeli: Why double quotes only there?

bitbucket_private_repository.py
TheFRedFox: Ha, thought they are necessary because of mult-line-string. ^^

hasegeli: Not when they are already inside parentheses.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,167 @@
...
+    parser.add_argument('--consumer-key',
+                        help="consumer key for oauth authentication")
+    parser.add_argument('--consumer_secret',
+                        help="consumer secret for oauth authentication")
+    parser.add_argument('--private-key', help="private key for oauth")
+    parser.add_argument('--passphrase',
+                        help="possible passphrase for the private key")
+    parser.add_argument('--format',
+                        help=("the format of the final print out. for the "
+                              "format the repository object will be passed"))
```
hasegeli: Also extra parentheses

bitbucket_private_repository.py
```diff
@@ -0,0 +1,167 @@
...
+    parser.add_argument('--delete', action="store_true",
+                        help="if found repositories should be deleted")
```
hasegeli: Randomly mixed double - single quotes.  Not that it is really a problem.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,167 @@
...
+    if auth == 'basic':
+        if not (args.username and args.password):
+            print("For basic authentication, 'username' and 'password' "
+                  "parameter are needed")
+            exit(3)
+        auth = HTTPBasicAuth(args.username, args.password)
+    elif auth == 'oauth':
+        if not (args.consumer_key and args.private_key):
+            print(('For oauth authentication, \'consumer-key\' '
+                   'and \'private-key\' parameter are needed'))
```
hasegeli: Better use double quotes there, too.  Also extra parentheses.

bitbucket_private_repository.py
```diff
@@ -0,0 +1,167 @@
...
+    auth = parse_auth_argument(args)
+
+    repositories = fetch_repositories(base_url, auth)
+    private_repos = [
+        repo for repo in repositories
+        if 'origin' not in repo and repo['project']['type'] == 'PERSONAL'
+        ]
+
+    if not private_repos:
+        # no private repos were found
```
hasegeli: Obsolete comment

PR #16, opened 2016-10-24
Add a script to check processes

hasegeli:


check_process_list.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env python
+"""InnoGames Monitoring Plugins - check_process_list.py
+
+This is a monitoring check that calls the "ps" binary, and parses
+the results.  It is compatible with BSD and GNU "ps", though they
+provide different set of variables.  It is possible to use any
+supported variable.  See you man pages for the list of them.
```
TheFRedFox: Also there are 2 spaces after the dot ... TWO SPACES!!!

PR #17, opened 2016-11-07
Add nagios script to check for unprotected default branches

TheFRedFox:
Add nagios script to check for unprotected default branches (mostly masters) in GitLab

 * OK if no projects have an unprotected default branch
 * WARNING if at least one project has an unprotected default branch


puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+
+    unprotected_masters = [
+        (project, branch.json()) for project, branch in master_branches
+        if branch.status_code != 200 or
+        (
+            not branch.json()['protected'] or
+            branch.json()['developers_can_merge'] or
+            branch.json()['developers_can_push']
+        )
```
hasegeli: This would call `branch.json()` 4 times.  We should instead do the filtering later.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+    parser.add_argument('--username',
+                        help='the username for basic authentication')
+    parser.add_argument('--password',
+                        help='the password for basic authentication')
+    parser.add_argument('--token',
+                        help='the private/access token with which the script '
+                             'can authenticate as the user behind the token')
+    parser.add_argument('--format',
+                        help='the format of the final print out. for the '
+                             'format the repository object will be passed')
```
hasegeli: What is the use of this argument?

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+    parser.add_argument('--branch', default='master',
+                        help='the branch which should be protected')
+    parser.add_argument('--protect', action='store_true',
+                        help='if found unprotected master branches should be '
+                             'automatically be protected')
```
hasegeli: Monitoring scripts shouldn't do such things.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+
+    if args.format:
+        format_string = args.format
+    else:
+        format_string = '{project[path_with_namespace]}: {project[web_url]}\n'
+
+    string = ''.join(format_string.format(project=project, branch=branch)
+                     for project, branch in unprotected_masters)
+    print('WARNING: '
+          '{amount} projects with unprotected default branches found\n{0}'
```
hasegeli: Chances of people seeing the detailed information on their monitoring system is much higher, if we don't print it after a new line.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def __init__(self, token_name, token):
+        self.token_name = token_name
+        self.token = token
+
+    def __eq__(self, other):
+        return all([
+            self.token_name == getattr(other, 'token_name', None),
+            self.token == getattr(other, 'token', None)
+        ])
```
hasegeli: Lets just say `and`.

hasegeli: Third argument of `getattr()` is unnecessary.  Most things are defaulted to `None` on Python.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def __ne__(self, other):
+        return not self == other
```
hasegeli: `!=`

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def __call__(self, r):
```
hasegeli: PEP 8 dissoutes single letter variables.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    endpoint = '/api/v3/projects/all'
+    params = {}
+    if search:
+        params['search'] = search
+    response = do_request('get', base_url, endpoint, params=params,
+                          auth=auth)
+    response = json.loads(response.text)
+    for project in response:
+        yield project
```
hasegeli: There is no benefit of making this a generator.  `response` is already materialised.  You can just return it.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env python
+#
+# InnoGames Monitoring Plugins - puppet_protected_default_branch
```
TheFRedFox: Not for every repository it makes sense to check it. What about private forks? What about some temporary repositories? Also repositories which shall be completely open? There are some and it is intended.
However I am open for the `--project` param.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
...
+def main(args):
+    base_url = args.base_url
+
+    auth = parse_auth_argument(args)
+
+    projects = fetch_projects(base_url, 'puppet', auth=auth)
+    puppet_projects = [
+        project for project in projects
+        if project['namespace']['name'] == 'puppet'
+        ]
```
TheFRedFox: Me too .. I don't know why, but PyCharm is doing it on its own. I didn't find out why it is doing so. I will have a second look at it.

puppet_protected_default_branch.py
TheFRedFox: I'm not sure, if you are talking about the `new line` after 'WARNING: ' (which won't be printed) or the new line after `... branches found`.
What is your suggestion then how to print it?

puppet_protected_default_branch.py
TheFRedFox: I had a look at the `HTTPBasicAuth` class and basically reused the code there. I already saw that the third argument is default to `None`, but I thought `HTTPBasicAuth` class will be fine enough. ^^

puppet_protected_default_branch.py
TheFRedFox: Like said above (reusing `HTTPBasicAuth` class code): I am not sure if a `__ne__` function is really needed here. Is `!=` maybe calling `not object == other` on it's own without overriding `__ne__` function?

puppet_protected_default_branch.py
hasegeli: No, I meant `\n` before `{0}`.  I would print something like: `WARNING: project with unprotected default branch: puppet/ig, puppet/admin, puppet/ndco`.

puppet_protected_default_branch.py
TheFRedFox: Typo for `repository` here, should be `project and branch`.
With this you can change the format of the lines in the outprint of the found branches.
See lines 121 to 126.

puppet_protected_default_branch.py
hasegeli: I see.  Sorry for not paying enough attention before.  We can write `not self.__eq__(other)` to make it clear.  `not x == y` is going to annoy everybody.

puppet_protected_default_branch.py
TheFRedFox: Well I built the output based on Flexi's wishes as it will be a check for him. And you have a direct different wish for the output .. so it seems it can be useful.
However, why not giving the choice? It is already implemented.

puppet_protected_default_branch.py
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env python
+#
+# InnoGames Monitoring Plugins - puppet_protected_default_branch
+#
+# This is a Nagios script which checks, if there are any puppet projects with
+# unprotected default branches (mostly master branches). The branches must be
+# strictly protected, means that no developer must not be allowed to push or
+# merge into the branch. The script uses the GitLab Server Rest Api and it is
```
TheFRedFox: The script is actually doing so. I actually think that the two negatives are correct in english as two negatives are one "huge" negative and not one positive expression. However I can rewrite it to be not confusing. ^^

gitlab_protected_branch.py
```diff
@@ -0,0 +1,265 @@
...
+    if args.group:
+        projects = fetch_group_projects(base_url, args.group, args.project,
+                                        auth=auth)
+    elif (args.project and
+          # if args.project represents an int or is like 'group/project'
+          ('/' in args.project or str(int(args.project)) == args.project)):
+        projects = [fetch_project(base_url, args.project, auth=auth)]
+    else:
+        projects = fetch_projects(base_url, args.project, auth=auth)
+    print(projects)
```
hasegeli: Is this intentional?

PR #21, opened 2016-11-10
Set default path on check_ssl_expiry.py

hasegeli:


check_ssl_expiry.py
```diff
@@ -49,12 +49,12 @@
 
 def parse_args():
     parser = ArgumentParser()
-    parser.add_argument('--cert-dir', dest='cert_dir',
+    parser.add_argument('--cert-dir', default='/etc/ssl/private',
```
eviscares: Why do you remove the "dest" here?
Also, I thought about defaulting that, but I think I'd rather have it set in the check, because that is more obvious for people looking at nagios configs.

check_ssl_expiry.py
```diff
@@ -49,12 +49,12 @@
...
                         help='The folder containing the certificates to check')
-    parser.add_argument('--warning-days', dest='warning_days', default=30,
+    parser.add_argument('--warning-days', default=30,
```
eviscares: And here?

check_ssl_expiry.py
```diff
@@ -49,12 +49,12 @@
...
                         help='Days before expiry of a cert that a warning is '
                         'triggered. Will be thirty if left unspecified')
-    parser.add_argument('--crit-days', dest='crit_days', default=7,
```
eviscares: And here?

check_ssl_expiry.py
hasegeli: Arguments starting with `--` are optional by design.  Failing with `TypeError: coercing to Unicode: need string or buffer, NoneType found`, is not a better default.

TheFRedFox:
This PR is obsolete after https://github.com/innogames/igmonplugins/pull/23?

hasegeli:
Yes

PR #22, opened 2016-11-10
Feature/check ssl expiry

eviscares:


TheFRedFox:
Could you please squash the last two commits. Git will recognize the rename by it's own and it will be possible to actually see the real diff. That would be really cool.

Additionally: What is going on here now? ^^ Why two more pull requests for changes from you and @hasegeli now? ^^

PR #23, opened 2016-11-11
Feature/check ssl expiry

eviscares:
[CHANGE]Renamed check_ssl_expiry to check_ssl_dir

Renamed script to check_ssl_dir to clarify intent and also to leave intent clear if functionality gets extended.


check_ssl_dir.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python
+
+# InnoGames Monitoring Plugins - check_ssl_expiry.py
```
hasegeli: Could you please change the name also?

PR #26, opened 2016-11-30
Add authentication check script to check for anonymous, basic, oauth, header and form authentication

TheFRedFox:
This is a Nagios script which checks, if the authentication to a specific url is possible or not. Supported authentication checks are: anonymous,; basic, oauth, header (e.g. token) and form authentication. Two Factor; Authentication is supported too by passing your TOTP Secret Key with the; parameter `totp`. Every single occurrence of `{totp}` in the header and input values will be replaced with the actual generated TOTP value. However right now it will just works with one single request.

The script will exit with:
- 0 (OK)         if authentication request returns a http status code
                     without error (means not 400 or above)
- 2 (CRITICAL)   if authentication request fails either because of the
                         request or the authentication

Use it like that:
`python authentication.py sub.example.com/login --auth basic --inputs username Bob password SECRET`
`python authentication.py sub.example.com/login --auth header --inputs ACCESS-TOKEN SECRET`
`python authentication.py sub.example.com/login --auth oauth --inputs private-key path/to/file passphrase SECRET consumer-key CONSUMER_KEY consumer-secret CONSUMER_SECRET`
`python authentication.py sub.example.com/login --auth form --inputs username Bob password SECRET code {totp} --totp SECRET_TOTP`
or as AJAX call
`python authentication.py sub.example.com/login --auth form --headers X-Requested-With XMLHttpRequest --inputs username Bob password SECRET code {totp} --totp SECRET_TOTP`


authentication.py
```diff
@@ -0,0 +1,242 @@
...
+    parser.add_argument('--format-fail',
+                        help='the format which will be used if the request was '
+                             'not successful due to authentication failure. '
+                             'the response as `response` and all arguments as '
+                             '`args` will be passed in.')
+    parser.add_argument('--format-error',
+                        help='the format which will be used if the request '
+                             'was not successful due to some other errors. '
+                             'all arguments as `args` and the raised error as '
+                             '`error` will be passed in.')
```
hasegeli: Do we need those?  It is not common for checks to customise this part.

authentication.py
```diff
@@ -0,0 +1,242 @@
...
+    return parser.parse_args()
+
+
+def main(args):
+    if args.totp:
+        import pyotp
+        args.totp = pyotp.TOTP(args.totp).now()
```
hasegeli: I don't like changing the original arguments.

authentication.py
```diff
@@ -0,0 +1,242 @@
...
+
+    headers = {}
+    if args.headers:
+        if len(args.headers) % 2 != 0:
+            print("The parameter 'headers' needs an even amount of "
+                  "elements")
+            sys.exit(3)
+
+        for name, value in zip(args.headers[0::2], args.headers[1::2]):
+            if '<totp>' in value:
```
hasegeli: It would be nicer to accept it as `{totp}`.  Then we could do: `value.format(totp=args.totp)`.

authentication.py
```diff
@@ -0,0 +1,242 @@
...
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('url',
+                        help='the base url of the application you want to '
+                             'check (e.g. https://sub.example.com)')
+    parser.add_argument('-m', '--method', default='get',
+                        help='the http method to use e.g. get, post, put, etc.')
+    parser.add_argument('-a', '--auth',
+                        choices=('none', 'basic', 'oauth', 'header', 'form'),
+                        default='none',
```
hasegeli: You can use normal `None` on the default, or better you can just not specify it.  `None` is the implicit default.

authentication.py
```diff
@@ -0,0 +1,242 @@
...
+            sys.exit(3)
+        auth = HTTPBasicAuth(args.username, args.password)
+    elif auth == 'header':
+        if not args.headers:
+            print("For header authentication the 'headers' parameter is needed")
+            sys.exit(3)
+
+        if len(args.headers) % 2 != 0:
+            print(
+                "The parameter 'headers' needs an even amount of elements")
```
hasegeli: Why do we check the same thing again?

authentication.py
```diff
@@ -0,0 +1,242 @@
...
+def parse_auth_argument(args):
+    auth = args.auth
+    if not auth or auth == 'none':
+        return None
+
+    if auth == 'basic':
+        if not (args.username and args.password):
+            print("For basic authentication, 'username' and 'password' "
+                  "parameter are needed")
+            sys.exit(3)
```
hasegeli: It is not nice to `exit()` from functions.  Raising an exception would be better.

authentication.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class HTTPFormAuth(requests.auth.AuthBase):
+    """Attaches HTTP Token Authentication to the given Request object."""
+
+    def __init__(self, inputs):
+        """
+        :param dict inputs:
+        """
+        self.inputs = inputs
```
hasegeli: Don't we need to call the parent `__init__()`?

authentication.py
```diff
@@ -0,0 +1,242 @@
...
+    def __call__(self, request):
+        """
+        :param requests.models.PreparedRequest request:
+        """
+        request.prepare_body(self.inputs, None)
+        return request
+
+
+if __name__ == '__main__':
+    main(parse_args())
```
hasegeli: `main()` could as well call `parse_args()`.  I don't think it is better to have function which accepts a custom arguments objects.

TheFRedFox: What is the better way? To pass all arguments like `def main(auth, username, password, ...)`? When I call `parse_args()` inside the `main` function, is it possible to use the function in another script then?

authentication.py
TheFRedFox: Copied it from `HttpBasicAuth` but I think it would be better to also call `__init__()`. I guess it is not necessary as there is no `__init__()` function in the `AuthBase` class.

authentication.py
TheFRedFox: I would rather offer the possibility than rewriting or duplicate the script for a small difference, because two persons want different output formats.

hasegeli: Who wants different format?  What is the use case?

authentication.py
hasegeli: I think it is okay to call `parse_args()` from `main()` we don't really expect it to be called elsewhere anyway.

authentication.py
TheFRedFox: Maybe one does not want the status code in the output, or maybe one does want the response text. But another one does not want it. I would like to offer the possibility to do so. What are the reasons to no offer it?

authentication.py
TheFRedFox: Discussed for a Best Practice and stick to calling `parse_args()` inside the `main` function.

src/authentication.py
```diff
@@ -0,0 +1,283 @@
...
+    parser.add_argument('-a', '--auth',
+                        choices=('basic', 'oauth', 'header', 'form'),
+                        help='The auth method to use, you can choose between '
+                             'basic, oauth, header and form authentication. '
+                             'If nothing is passed it will try an anonymous '
+                             'request.')
+    # Basic Auth
+    # Form Auth
+    # Header Auth
+    # OAuth
```
hasegeli: This is rather unusual comment style.  What do you meant exactly.

src/authentication.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def collect_pair_values(values):
+    if not values:
+        return {}
+
+    if len(values) % 2 != 0:
+        raise ValueError("An even amount of elements are needed.")
+
+    return zip(values[0::2], values[1::2])
```
hasegeli: `zip(*((iter(values), ) * 2))` would work more efficiently.  No, I am not offering to change it.

src/authentication.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def __call__(self, request):
+        """
+        :param requests.models.PreparedRequest request:
+        """
+        if not self.headers:
+            return request
+
+        # for name, value in self.headers.items():
+        #     request.headers[name] = value
```
hasegeli: Is these left intentionally?

src/authentication.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class HTTPFormAuth(requests.auth.AuthBase):
+    """Attaches HTTP Token Authentication to the given Request object."""
+
+    def __init__(self, inputs):
+        """
+        :param dict inputs:
+        """
+        super(requests.auth.AuthBase, self).__init__()
```
hasegeli: `super()` should take the actual class as the first argument.  This is calling the grandparent.

src/authentication.py
```diff
@@ -0,0 +1,283 @@
...
+
+class HTTPHeaderAuth(requests.auth.AuthBase):
+    """Attaches HTTP Header Authentication to the given Request object."""
+
+    def __init__(self, headers):
+        super(requests.auth.AuthBase, self).__init__()
+        self.headers = headers
+
+    def __eq__(self, other):
+        return self.headers == getattr(other, 'headers')
```
hasegeli: Why not `other.headers`?  If you want it to not cause an error, you need to use `hasattr()`.

src/authentication.py
```diff
@@ -0,0 +1,283 @@
...
+    with open(private_key, 'rb') as fd:
+        rsa_key = RSA.importKey(fd.read(), passphrase)
+
+    return OAuth1(client_key=consumer_key, client_secret=consumer_secret,
+                  signature_method='RSA-SHA1', rsa_key=rsa_key)
+
+
+def get_new_totp(totp):
+    import pyotp
+    return pyotp.TOTP(totp).now()
```
hasegeli: I would `from pyotp import TOTP`.

src/authentication.py
TheFRedFox: Now there are too few arguments left to need such comments to mark them.

src/authentication.py
TheFRedFox: Actually I changed it and commented it, that the trick is to use the exact same iterator twice to go through the list in parallel.

PR #28, opened 2016-11-30
Add missing script and fix another one.

TuxPowered42:


src/check_puppetd
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+#/////////////////////////////////////////////////////
+#// check_puppetd
+#/////////////////////////////////////////////////////
```
hasegeli: Can you rename it to `check_puppetd.sh`.

PR #30, opened 2016-12-09
Simplify the script, use multiline output

TuxPowered42:


src/check_bgp_peers.sh
```diff
@@ -6,32 +6,28 @@
...
-done
+	routers="${routers}${router}: ${status}
+"
+done < /tmp/bgpstatus
 
 if [ -n "$CRIT" ]; then
-	echo "No connection to routers $dead!"
+	echo "No connection to some routers!"
 else
 	echo "All routers are fine."
```
TheFRedFox: Is it necessary to continue and print out all routers if everything is fine? If it is not necessary one could directly exit here.

src/check_bgp_peers.sh
```diff
@@ -6,32 +6,28 @@
...
 	status=${proto#*:}
+    status=${status# *}
 
 	if [ "$status" != "Established" ]; then
 		CRIT="yes"
-		dead="$dead $router"
 	fi
-done
+	routers="${routers}${router}: ${status}
+"
```
TheFRedFox: I am not sure about shell, but can't we use `\n` in the string? Like `"...: ${status}\n"`. IMO it would look a bit nicer that way.

src/check_bgp_peers.sh
```diff
@@ -6,32 +6,28 @@
 #2   CRITICAL
 #3   UNKNOWN
 
-protos=`birdc 'show protocols' | awk '$2=="BGP" { print $1":"$6}'`
 
-for proto in $protos; do
-	router=${proto%:*}
+
+birdc 'show protocols' | awk '$2=="BGP" { ORS=" "; print $1":"; for (i=6; i<=NF; i++) print $i; ORS="\n"; print ""}' > /tmp/bgpstatus
```
TheFRedFox: Is it necessary and especially is it safe to use a file here? IMO I would prefer not using files if we can avoid it.

src/check_bgp_peers.sh
```diff
@@ -6,32 +6,28 @@
 #2   CRITICAL
 #3   UNKNOWN
 
-protos=`birdc 'show protocols' | awk '$2=="BGP" { print $1":"$6}'`
 
-for proto in $protos; do
-	router=${proto%:*}
+
```
TheFRedFox: Dunno if there are any code styles for shell, but I think 3 blank lines are not needed. ^^

src/check_bgp_peers.sh
```diff
@@ -6,32 +6,28 @@
...
-		dead="$dead $router"
 	fi
-done
+	routers="${routers}${router}: ${status}
+"
+done < /tmp/bgpstatus
 
 if [ -n "$CRIT" ]; then
-	echo "No connection to routers $dead!"
+	echo "No connection to some routers!"
```
TheFRedFox: Some are preferring the result inside the subject string, so: `"CRITICIAL: No connection to some routers!"`. I don't have a mind there, but I would say to stick to one consistent way. And right now the consistent way is to print the result.
So please use `"CRITICIAL: No ..."`
and below
`"OK: All routers ..."`

src/check_bgp_peers.sh
TuxPowered42: Otherwise I have to run a subshell or use bashisms. First one makes very hard to get variables out, the 2nd one will not work on BSD with no Bash. Interwebz are suggesting to use a pipe, which is even uglier.

src/check_bgp_peers.sh
TuxPowered42: No, in sh we can not.

src/check_bgp_peers.sh
TuxPowered42: That is useless, duplicate information and we already removed it from quite many checks and I hope more will come.

src/check_bgp_peers.sh
TuxPowered42: Yes, I want to *always* print the list of known BGP peers so that even if all is fine to be able to see if really proper ones get checked. And it in fact simplifies the code, as the list is just printed always.

src/check_bgp_peers.sh
TuxPowered42: I think you made this comment in wrong line, but I fixed empty lines after the 1st comment.

src/check_bgp_peers.sh
TuxPowered42: OK, I found a way to do it.

src/check_bgp_peers.sh
TheFRedFox: Then the team has different opinions right now for that. As I said, I don't have an opinion on that, I don't mind both ways. But I would like to have one way to do so for the team.

src/check_bgp_peers.sh
TheFRedFox: Ok .. if there is no quite easy way to do, files are fair enough, I guess. However I would suggest if we will know an easy way to do so, to rather use this than files. 🙂

src/check_bgp_peers.sh
TheFRedFox: Cool 👍

src/check_bgp_peers.sh
TuxPowered42: I guess it would have helped much more if there was more lines quoted, because your suggestion is about the *next* block of code. Yes, that I can simplify.

TheFRedFox: Well it is about changing this block here, too. However I thought it is safe enough to understand, what I mean. 😋

PR #31, opened 2016-12-20
Add check to check kvm domain memory allocation

kofrezo:


src/check_kvm_memory.py
```diff
@@ -0,0 +1,113 @@
...
+            if args.verbose:
+                print('{}: {} MiB'.format(domain_name, domain_memory))
+        print('')
+        print('usable: {} MiB'.format(hypervisor_memory))
+        print('used: {} MiB'.format(domain_allocated))
+        print('free: {} MiB'.format(hypervisor_free))
+        print('reserved incl.: {} MiB'.format(args.reserved))
+        print('overhead incl.: {} MiB'.format(args.overhead))
+
+    exit(code)
```
hasegeli: Lets put this to `main()`.

src/check_kvm_memory.py
```diff
@@ -0,0 +1,113 @@
...
+# Copyright (c) 2016, InnoGames GmbH
+#
+
+from argparse import ArgumentParser
+
+from os import sysconf
+
+from libvirt import openReadOnly
+
+from sys import exit
```
hasegeli: We normally use empty lines between imports to separate the ones coming from different sources.  You could maybe only separate `libvirt` from others, because it is not part of the standard library.

src/check_kvm_memory.py
```diff
@@ -0,0 +1,113 @@
...
+    try:
+        result = dict()
+        con = openReadOnly(None)
+        domains = con.listAllDomains()
+
+        for domain in domains:
+            result[domain.name()] = float(domain.maxMemory()) / 1024.0
+            result[domain.name()] += args.overhead
+    finally:
+        con.close()
```
hasegeli: Nice

src/check_kvm_memory.py
```diff
@@ -0,0 +1,113 @@
...
+
+    memory_bytes = sysconf('SC_PAGE_SIZE') * sysconf('SC_PHYS_PAGES')
+    return float(memory_bytes) / 1024.0**2.0 - args.reserved
+
+
+def get_domain_memory():
+    """Get memory allocated by domains in MiB -> dict"""
+
+    try:
+        result = dict()
```
hasegeli: This line can go outside.

src/check_kvm_memory.py
```diff
@@ -0,0 +1,113 @@
...
+        con = openReadOnly(None)
+        domains = con.listAllDomains()
+
+        for domain in domains:
+            result[domain.name()] = float(domain.maxMemory()) / 1024.0
+            result[domain.name()] += args.overhead
```
hasegeli: I may be better to not call `domain.name()` two times.

src/check_kvm_memory.py
```diff
@@ -0,0 +1,113 @@
...
+        code = 1
+    elif hypervisor_free < args.warning:
+        print('Memory WARNING {} MiB {}% free|{}'.format(
+            hypervisor_free, hypervisor_allocated, hypervisor_free
+        ))
+        code = 2
+    else:
+        print('Memory OK {} MiB {}% free|{}'.format(
+            hypervisor_free, hypervisor_allocated, hypervisor_free
+        ))
```
hasegeli: Those `print()`s all look the same.  Perhaps it is better to `print()` them with STATUS later?

PR #33, opened 2017-01-10
WIP: Warn if pool is not served by all nodes

TuxPowered42:


src/check_lbpools.py
```diff
@@ -71,17 +71,22 @@ def compare_pools(pools, testtool, send):
...
             in_testtool = True
 
+        optimal_nodes = pool_v.get('optimal_nodes')
+
         if pool_v['has_healthchecks']:
             if in_testtool:
                 if num_nodes == 0:
                     exit_code = exit_crit
-                else:
+                elif num_nodes == optimal_nodes:
```
leoleovich: What is "optimal_nodes"? If I have 5 nodes under LB, how much optimal nodes will be?

TuxPowered42: optimal_nodes is the amount of nodes in configuration of a LB Pool. So all possible ones.

TuxPowered42:
It was fixed by other changes.

PR #34, opened 2017-02-03
[NEW] Check if local branch is up to date with remote

elmatze79:


src/check_git_origin.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+####
+# Check the git origin if there are non merged/rebased commits
+####
+
+if [[ ! -d $1/.git ]]
+then
+    echo "Unknown: No valid git repository found in $1"
```
TheFRedFox: Consistent would be: 'UNKNOWN: ', uppercase

TheFRedFox:
I am wondering for which reason this is used. I was looking at igmonplugins as some monitoring scripts for servers. Is it possible that there are local git repos on servers, which have to be checked then?

elmatze79:
We have this one special project where I am not allowed to automatically have changes merged, so I need to know if there are changes to the repositories

src/check_git_origin.sh
```diff
@@ -0,0 +1,23 @@
...
+cd $1
+git fetch > /dev/null
+NUM_COMMITS=`git log --oneline HEAD..origin | wc -l`
+if [[ $NUM_COMMITS -eq 0 ]]
+then
+    echo "OK: $1 is up to date"
+    exit 0
+fi
+
+echo "Warning: $1 is $NUM_COMMITS commits behind origin"
```
TheFRedFox: Sry, but it is otherway around. Uppercase messages everywhere: UNKNOWN, OK, WARNING, CRITICAL

elmatze79: https://nagios-plugins.org/doc/guidelines.html#AEN78

TheFRedFox: Well, I checked all our scripts .. then we should do it in a consistent way soon. I don't mind if it is not uppercase, but I mind consistency. ^^

PR #36, opened 2017-02-09
Rename authentication.py

hasegeli:


TheFRedFox:
Why? Also for the other scripts like atlassian_expiring_licenses etc? Tbh, I am against to prefix everything. If you prefix everything with the same you prefix nothing. It is just harder to read. If we want to introduce something else than checks in the future, I would suggest some subdirectory or the like.

TheFRedFox:
After discussing it in person, I see that there are other scripts than checks and that it is a best practice to do so for nagios scripts. So I am ok with it.

PR #41, opened 2017-04-26
New check of deleted files

leoleovich:


src/check_deleted_files.awk
```diff
@@ -0,0 +1,12 @@
+BEGIN {
+    cmd = "/usr/bin/lsof"
+    while (cmd | getline line) {
+            split(line, line_split)
+            if (line_split[10] == "(deleted)" && line_split[7] > 10485760) {
+                print line_split[7]/1048576 " MB wasted by deleted file " line_split[9]
```
TuxPowered42: Please use binary prefix. Is this supposed to be multiline output for Nagios? Only the 1st file will be show by default. It would be better to show total wasted disk space in the 1st line and then show all files later.

src/check_deleted_files.awk
```diff
@@ -0,0 +1,23 @@
...
+        if (line_split[10] == "(deleted)" && line_split[7] > mebibytes_warning*M) {
+            total_waste += line_split[7]/M
+            message = message "\n" line_split[7]/M " MiB wasted by " line_split[9]
+            big_file = 1
+        }
+    }
+    close(cmd)
+    if (big_file == 1 || total_waste > mebibytes_warning*M) {
+        print total_waste " MiB wasted by deleted files:" message
+        exit 1
```
TuxPowered42: Don't we also need exit 0 with message stating that all is fine?

src/check_deleted_files.awk
```diff
@@ -0,0 +1,23 @@
+BEGIN {
+    mebibytes_warning = ARGV[1]
+    if (mebibytes_warning !~ /^[0-9]+$/) {
+        print "Please specify number of MiB"
```
TuxPowered42: Any idea to write it better? Maybe rather something like "wasted disk space in MiB"? It could also say that this is the warning threshold as other Nagios plugins do.

leoleovich:
```
sentry.csm:~# /usr/bin/awk -f /usr/lib/nagios/igmonplugins/check_deleted_files.awk 10
16.6317 MiB wasted by deleted files
16 MiB wasted by /var/lib/postgresql/9.5/main/pg_xlog/000000010000002A000000A1
sentry.csm:~# echo $?
1
sentry.csm:~# /usr/bin/awk -f /usr/lib/nagios/igmonplugins/check_deleted_files.awk 100
16.6317 MiB wasted by deleted files
sentry.csm:~# echo $?
0
sentry.csm:~# /usr/bin/awk -f /usr/lib/nagios/igmonplugins/check_deleted_files.awk 16
16.6317 MiB wasted by deleted files
sentry.csm:~# echo $?
1
```

PR #42, opened 2017-04-27
Improve check_mysql_processes.py

hasegeli:


src/check_mysql_processes.py
```diff
@@ -0,0 +1,128 @@
...
+    if any(c(processes) for c in args.critical):
+        exit(ExitCodes.critical)
+    if any(c(processes) for c in args.warning):
+        exit(ExitCodes.warning)
+    exit(ExitCodes.ok)
+
+
+def get_processlist():
+    """Return the processes as a list of dicts"""
+    args = parse_args()
```
TheFRedFox: I know, you didn't touch it, but I don't like the `parse_args()` inside the `get_processlist()` function. Would suggest to add parameters to the function.

src/check_mysql_processes.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class ExitCodes:
+    ok = 0
+    warning = 1
+    critical = 2
+
+
+class Check:
+    pattern = recompile('\A(([0-9]+) *for *)?([0-9]+)\Z')
```
TheFRedFox: Shouldn't it be `\s*` instead of ` *`?

src/check_mysql_processes.py
```diff
@@ -0,0 +1,128 @@
...
+    ))
+    parser.add_argument(
+        '--warning',
+        nargs='*',
+        type=Check,
+        default=[Check('1 for 30')],
+        help='Warning threshold in count and seconds (default: %(default)s)'
+    )
+    parser.add_argument(
+        '--critical',
```
TheFRedFox: Saw it through the commits: Why no short arguments?

src/check_mysql_processes.py
```diff
@@ -0,0 +1,128 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from argparse import ArgumentParser, ArgumentTypeError
+from operator import itemgetter
+from sys import exit
+from re import compile as recompile
```
TheFRedFox: To be honest `recompile` could be a bit misleading.

hasegeli: Yes, very much.  I haven't noticed.  Renamed.

src/check_mysql_processes.py
hasegeli: Even though, the short ones are the POSIX standard, I don't see a value in supporting them on the monitoring scripts.  We only write them once on the monitoring configuration files.  Only providing a single more clear way is better in my opinion.

src/check_mysql_processes.py
hasegeli: Good catch.  Changed.

src/check_mysql_processes.py
hasegeli: ` *` matches with the space character only.  `\s*` also matches with tabs and similar.  I think it is better.  Changed.

src/check_mysql_processes.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class ExitCodes:
+    ok = 0
+    warning = 1
+    critical = 2
+
+
+class Check:
+    pattern = regexp_compile('\A(([0-9]+)\s*for\s*)?([0-9]+)\Z')
```
TheFRedFox: Just saw it now. Normally one uses `^` and `$` for begin and end of line in RegExp and I am not sure about the position of the first capturing group. But at least we can change it to a non capturing group as it is not needed:

`^(?:([0-9]+)\s*for\s*)?([0-9]+)$`

Then also change the group indexes below.

src/check_mysql_processes.py
TheFRedFox: I see. Can agree with that. 😉

src/check_mysql_processes.py
hasegeli: `^` and `$` matches with the begin and end of every line.  `\A` and `\Z` matches with the begin and end the string, so it is safer.  I didn't knew about non-capturing groups.  Will change.

src/check_mysql_processes.py
```diff
@@ -98,23 +97,63 @@ class ExitCodes:
     critical = 2
 
 
+class Interval:
+    units = [
+        ('s', 1),
+        ('min', 60),            # "m" would mean "metre"
```
TheFRedFox: Tbh I guess we can ignore that as you could find some other definition for every character here. Like `d` is diameter, `h` could be height. It is known for this script that the parameters are for time ranges, and for that `m` for minutes is good enough.

src/check_mysql_processes.py
```diff
@@ -98,23 +97,63 @@ class ExitCodes:
...
+        ('h', 60 * 60),
+        ('d', 12 * 60 * 60),
+    ]
+
+    def __init__(self, multiplier, unit):
+        for key, value in self.units:
+            if key == unit:
+                break
+        else:
+            raise Exception('Unit "{}" couldn\'t found'.format(unit))
```
TheFRedFox: Double quotes can be used here.

src/check_mysql_processes.py
```diff
@@ -98,23 +97,63 @@ class ExitCodes:
     critical = 2
 
 
+class Interval:
+    units = [
+        ('s', 1),
+        ('min', 60),            # "m" would mean "metre"
+        ('h', 60 * 60),
+        ('d', 12 * 60 * 60),
```
TheFRedFox: All the calculation might be constants anyway after "compiling", right?

src/check_mysql_processes.py
```diff
@@ -98,23 +97,63 @@ class ExitCodes:
...
+                break
+        return '{}{}'.format(self.seconds / value, key)
+
+
 class Check:
-    pattern = regexp_compile('\A\s*([0-9]+)(?:\s*for\s*([0-9]+))?\s*\Z')
+    pattern = regexp_compile(
+        '\A\s*'         # Input start
+        '([0-9]+)'      # | Count part
+        '(?:'           # | Time part start
```
TheFRedFox: I guess we don't even need that group here. I also think that we do not need the non-greedy flags in here. Or do you know some edge cases where it would break?

`\A\s*([0-9]+)\s*for\s*([0-9]+)(s|m|h|d)\s*\Z` should be good enough:

``` python
regexp_compile(
    '\A\s*'
    '([0-9]+)'
    '\s*for\s*'
    '([0-9]+)'
    '({})'
    '\s*\Z'
)
```

src/check_mysql_processes.py
```diff
@@ -134,48 +164,65 @@ def __str__(self):
...
 
-        if count >= self.count:
-            return self.get_problem(count)
+        if count >= self.get_count_limit(max_connections):
+            return self.format_problem(count)
         return None
 
-    def get_problem(self, count):
+    def get_count_limit(self, max_connections=None):
+        if not self.count_unit:
```
TheFRedFox: Maybe rather using its own `relative` function?

src/check_mysql_processes.py
```diff
@@ -134,48 +164,65 @@ def __str__(self):
...
-        '(?:'           # | Time part start
-        '\s*for\s*'     # | | Time separator
-        '([0-9]+)'      # | | Time multiplier
-        '({})?'         # | | Time unit
-        ')?'            # | Time part end
-        '\s*\Z'         # Input end
+        '\A\s*'         # Input
+        '(?:'           # | Optional count clause
+        '([0-9]+)'      # | | Number
+        '(%)?'          # | | Unit
```
TheFRedFox: `"Optional Unit"`

src/check_mysql_processes.py
hasegeli: Changed.

src/check_mysql_processes.py
hasegeli: Changed.

hasegeli:
I have split this commit into two.

Issue #38, opened 2017-04-11
check_cron_runtime fails sometimes with TypeError


juergenthomann:
Stacktrace:
```Traceback (most recent call last):
  File "/usr/lib/nagios/igmonplugins/check_cron_runtime.py", line 183, in <module>
    main(**parse_args())
  File "/usr/lib/nagios/igmonplugins/check_cron_runtime.py", line 114, in main
    if int(cron_child_time) > 3600:
TypeError: int() argument must be a string or a number, not 'list'
```

I haven't found a reason yet, why `execute("ps -o 'etimes=' " + pid).strip('    ').strip('\n')` returns a list and it is quite hard to trigger it, but it can be reproduced with running `while true; do /usr/lib/nagios/igmonplugins/check_cron_runtime.py || break; done` for about a minute or two.

TheFRedFox:
`cron_child_time` is first assigned with `[]` (line 95) and if the execution in line 108 throws an `AttributeError` it will be still a list. As the `TypeError` says, it can not convert a `list` to an `int`.

I would suggest to assign `cron_child_time` to `None` first and then check for `not None` after the execution in line 108.

TheFRedFox:
Should be fixed now. However the script seems to be replaced soon by `check_process_list.py`.

TheFRedFox:
Because of the `elif` there had to be another check. The None check is now done before and if it is None the loop will directly be continued.

So should be fixed now.

TheFRedFox:
Looks fixed now. Will close the issue then.

PR #44, opened 2017-05-02
Fix issue #38 for check_cron_runtime.py

TheFRedFox:
Fixing possible `TypeError` with:
`int() argument must be a string or a number, not 'list'`
by assigning `cron_child_time` with `None` and checking for value
before trying to cast it to int.


hasegeli:
I will replace this script with check_process_list.py.

TheFRedFox:
Ah I see, didn't know `check_process_list` will be a replacement for that.

PR #46, opened 2017-06-02
check_mysql_prcesses.py: Add transaction support

hasegeli:


src/check_mysql_processes.py
```diff
@@ -1,6 +1,48 @@
 #!/usr/bin/env python
 '''InnoGames Monitoring Plugins - check_mysql_process_list.py
```
TheFRedFox: My IDE is saying, that docstring is wrapped with triple double-quotes and not triple singe-quotes.

src/check_mysql_processes.py
```diff
@@ -65,69 +105,135 @@ def parse_args():
...
+
+    def get_innodb_status(self):
+        if self.innodb_status is None:
+            result = self.execute('SHOW ENGINE INNODB STATUS')
+            assert len(result) == 1
+            self.innodb_status = defaultdict(list)
+            lines = result[0]['status'].splitlines()
+            header_id = -2
+            header = None
+            for line_id, line in enumerate(lines):
```
TheFRedFox: Hard to understand for me, what is happening here, as I don't know how the string to parse looks like.

If you don't mind, you can describe an example in the doc block.

Also for the other parsing methods. Would be more easily to understand what is happening there and also to review if it is correct what you want to approach with it.

src/check_mysql_processes.py
```diff
@@ -163,66 +275,127 @@ def __str__(self):
...
-        ')?'            # | '
-        '(?:'           # | Optional time clause
-        '\s*for\s*'     # | | Separator
-        '([0-9]+)'      # | | Number
-        '({})?'         # | | Optional unit
-        ')?'            # | '
-        '\s*\Z'         # '
-        .format('|'.join(k for k, v in Interval.units))
-    )
+    pattern = regexp_compile('\s*'.join([   # Allow spaces between everything
```
TheFRedFox: As the regexp is growing more, it might make sense to separate it into several string variables.

src/check_mysql_processes.py
```diff
@@ -65,69 +105,135 @@ def parse_args():
...
+    def get_max_connections(self):
+        if self.max_connections is None:
+            result = self.execute("SHOW VARIABLES LIKE 'max_connections'")
+            assert len(result) == 1
+            self.max_connections = int(result[0]['value'])
+        return self.max_connections
+
+    def get_problems(self, checks):
+        if any(c.relative() for c in checks) and self.max_connections is None:
+            self.fetch_max_connections()
```
TheFRedFox: `fetch_max_connections` does not exist. I guess you mean `get_max_connections`, right?

src/check_mysql_processes.py
```diff
@@ -65,69 +105,135 @@ def parse_args():
...
+                    continue
+                txn_info = self.parse_transaction_header(lines[header_id])
+                if not txn_info:
+                    continue
+                txn_id = int(line[len('MySQL thread id '):].split(',', 1)[0])
+                self.txns[txn_id] = txn_info
+
+        return self.txns.get(process_id)
+
+    def parse_transaction_header(self, line):
```
TheFRedFox: as `self` isn't used somewhere, the method could be a static one. But not mandatory, just saying. ^^

src/check_mysql_processes.py
```diff
@@ -65,69 +105,135 @@ def parse_args():
...
+            self.processes.sort(key=itemgetter('time'), reverse=True)
+        return self.processes
+
+    def get_innodb_status(self):
+        if self.innodb_status is None:
+            result = self.execute('SHOW ENGINE INNODB STATUS')
+            assert len(result) == 1
+            self.innodb_status = defaultdict(list)
+            lines = result[0]['status'].splitlines()
+            header_id = -2
```
TheFRedFox: Read below first:

Especially here, I do not get the `-2` initialization for `header_id`. Some documentation would really be good to understand, what's happening below.

TheFRedFox:
Also have a look at line 270 to 274: 
``` python
def __str__(self):
    for unit, multiplier in reversed(self.units):
        if self.seconds > multiplier:
            break
    return '{}{}'.format(self.seconds / multiplier, unit)
```
It was not touched by you and I guess it is not possible in this script, but `multiplier` and `unit` might not be assigned if `self.units` is empty.

src/check_mysql_processes.py
hasegeli: It might be better.  I had put effort to make this readable in its current form.  I don't feel like reworking it again.  I promise I will not add anything else.

src/check_mysql_processes.py
hasegeli: I forgot this block there.  Now removed.

src/check_mysql_processes.py
hasegeli: Yes, but it is continuation of the previous method.

src/check_mysql_processes.py
hasegeli: Your editor is right.

src/check_mysql_processes.py
hasegeli: Hopefully now I made it clear.

src/check_mysql_processes.py
hasegeli: I added more comments.

PR #47, opened 2017-06-20
Add check of switch links and cpu load

TuxPowered42:


src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+
+from argparse import ArgumentParser
+from pysnmp import proto, error
+from pysnmp.entity.rfc3413.oneliner import cmdgen
+from pysnmp.entity.rfc3413.oneliner.cmdgen import (
+        CommunityData,
+        UsmUserData,
+        usmHMACSHAAuthProtocol,
+        usmAesCfb128Protocol,
+        usmDESPrivProtocol,
```
hasegeli: Too much indented

src/switch.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python3
+#
+# igmonplugins - Switch checks:
+#                - cpu usage
+#                - port state
+#
+# Copyright (c) 2017, InnoGames GmbH
+#
```
hasegeli: Could you please add a license?  We don't have a global one on this project?

src/switch.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python3
```
hasegeli: Wouldn't it work on Python 2?  If so, just `python` is a more portable choice.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+
+    print(exit_msg)
+    sys.exit(exit_code)
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument(
+            '-H', dest='switch', required=True,
+            type=str, help='Hostname of a switch'
```
hasegeli: Too much indented

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+        exit_msg = 'Unknown check type {} requested'.format(args.check)
+
+    print(exit_msg)
+    sys.exit(exit_code)
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument(
+            '-H', dest='switch', required=True,
```
hasegeli: I don't like `required` optional arguments.  Wouldn't `localhost` be a reasonable default.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+            type=str, help='Hostname of a switch'
+    )
+    parser.add_argument(
+        '--check', type=str, help='Check to perform',
+        choices=['cpu', 'links'],
+        default='links',
+    )
+    parser.add_argument(
+        '-v', dest='verbose', action='store_true',
+        help='Verbose output - show OK ports',
```
hasegeli: Some arguments are long some are short.  I prefer to provide long ones only on such scripts intended for non-interactive use, even though they are less standard.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+    """ Get a single value from SNMP """
+
+    errorIndication, errorStatus, errorIndex, varBinds = cmd_gen.getCmd(
+        snmp['auth_data'],
+        snmp['transport_target'],
+        OID,
+    )
+    if errorIndication:
+        raise SwitchException('Unable to get SNMP value: {}'.
+                              format(errorIndication))
```
hasegeli: Dot before the method please

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+    return ret
+
+
+def convert_snmp_type(varBinds):
+    """ Convert SNMP data types to something more convenient: int or str """
+    val = varBinds[0][1]
+    if type(val) in [
+        proto.rfc1902.Integer,
+        proto.rfc1902.Counter32,
+        proto.rfc1902.Counter64,
```
hasegeli: Aren't they subclass of each other?  Maybe what you need is `isinstance()`.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+
+    if 'PowerConnect' in model:
+        return 'powerconnect'
+    elif 'ProCurve' in model:
+        return 'procurve'
+    elif 'ExtremeXOS' in model:
+        return 'extreme'
+
+    print('Unknown switch model', file=sys.stderr)
+    return None
```
hasegeli: Why not raising an `Exception`?

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+            port_names[port_index].startswith('Vl') or
+            port_names[port_index].startswith('CPU') or
+            port_names[port_index].startswith('Po')
+        )):
+            continue
+
+        if (model == 'procurve' and (
+            port_names[port_index].startswith('VLAN') or
+            port_names[port_index].startswith('Trk')
+        )):
```
hasegeli: One level of `if`/`elif`s for `model`s and `if`s inside it would be more readable.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+            continue
+
+        local_exit = 1
+        msg = (
+            'WARNING: Unhandled but bad status (admin: {}, oper:{})!'.
+            format(port_admin_states[port_index], port_oper_states[port_index])
```
hasegeli: Dot before the method please

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+        )
+
+        # The only OK cases are:
+        # Port is named, enabled and up.
+        if (
+            port_aliases[port_index].strip() != '' and
```
hasegeli: It is Pythonic to rely on empty string being `False`.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+        ):
+            local_exit = 2
+            msg = 'CRITICAL: Unnamed port is up!'
+
+        if (
+            port_aliases[port_index] != '' and
+            port_oper_states[port_index] == 2
+        ):
+            local_exit = 2
+            msg = 'CRITICAL: Named port is down!'
```
hasegeli: I think you can make this also more readable with multiple levels of `if`s.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+
+        if local_exit > 0 or args.verbose:
+            outmsg += (
+                '{} "{}": {}\n'.
+                format(port_names[port_index], port_aliases[port_index], msg)
```
hasegeli: Dot before the method please

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+    elif 'ProCurve' in model:
+        return 'procurve'
+    elif 'ExtremeXOS' in model:
+        return 'extreme'
+
+    print('Unknown switch model', file=sys.stderr)
+    return None
+
+
+def get_laggs(snmp, model):
```
hasegeli: This is not used.

src/switch.py
```diff
@@ -0,0 +1,374 @@
...
+        return {
+            x: y for x, y
+            in get_snmp_table(snmp, LAGG_OIDS[model]).items()
+            if y != 0
+        }
+
+    return None
+
+
+def standarize_portname(port_name):
```
hasegeli: This neither

src/switch.py
TuxPowered42: I have no idea if it works with Python2 and I refuse to check it.

src/switch.py
TuxPowered42: Added.

src/switch.py
TuxPowered42: Fixed.

src/switch.py
TuxPowered42: Fixed.

src/switch.py
TuxPowered42: Fixed.

src/switch.py
TuxPowered42: Right, it came from the previous script. I will remove it.

src/switch.py
TuxPowered42: This too.

src/switch.py
TuxPowered42: Fixed.

src/switch.py
TuxPowered42: That would be less readable IMO.

src/switch.py
TuxPowered42: Fixed.

src/switch.py
TuxPowered42: True, fixed.

src/switch.py
TuxPowered42: Maybe pythonic, but quite unreadable code. I want to know what the string contains by first look on it, not by deep knowledge of language internals.

src/switch.py
TuxPowered42: I dunno. I've seen `-v` in way too many programs as verbose, to consider the long version. The other short parameters are de facto Nagios plugin standard. I really don't feel like forcing long or short on all of them. That would not really make the world a better place and if anybody is confused, there is `-h` for them.

src/switch.py
TuxPowered42: They don't seem to be.

src/switch.py
TuxPowered42: Fixed.

src/switch.py
hasegeli: They are all subclass of `pyasn1.type.univ.Integer` whatever that is.  It is fine to leave it like this.

PR #48, opened 2017-06-21
Checks: Add check_steal_time.py

lacunoc:
This script checks the steal time of all vCPUs
on the regarding domain and raises a warning or
critical state if a reasonable threshold is reached.
Values for the warning and critical thresholds can
be specified using parameters.

### DIFF
--- a/check_steal_time.py
+++ b/check_steal_time.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+#
+# InnoGames Monitoring Plugins - check_steal_time.py
+#
+# This script checks the steal time of all vCPUs on the regarding domain and
+# raises a warning or critical state if a reasonable threshold is reached.
+# Values for the warning and critical thresholds can be specified using
+# parameters.
+#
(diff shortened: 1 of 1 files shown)


check_steal_time.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return parser
+
+
+def get_steal_time():
+    """Get the actual steal time of the host"""
+
+    output = subprocess.check_output("iostat -c", shell=True)
+    output = output.split()
+    steal = re.sub(',', '.', output[18])
```
hasegeli: You don't need `re`.  You can just use `str.replace()`.

check_steal_time.py
```diff
@@ -0,0 +1,88 @@
...
+        default=75
+    )
+
+    return parser
+
+
+def get_steal_time():
+    """Get the actual steal time of the host"""
+
+    output = subprocess.check_output("iostat -c", shell=True)
```
hasegeli: You used single quotes for strings before.

PR #52, opened 2017-07-03
redis cluster: Add nagios check

lacunoc:
This script checks the status of redis cluster
which consists of at least 6 instances (3 master + 3 slaves).
The status can be fully functional but degraded. It raises a
warning state if the cluster is degraded but still working.
A critical state represents a broken cluster.

### DIFF
--- a/src/check_redis_cluster.py
+++ b/src/check_redis_cluster.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python
+#
+# InnoGames Monitoring Plugins - check_redis_cluster.py
+#
+# Check the status of redis cluster
+#
+# This script checks the status of redis cluster which consists of at least
+# 6 instances (3 master + 3 slaves).  The status can be fully functional
+# but degraded.
+#
+# => administrative intervention required
+#
+# It raises a warning state if the cluster is degraded but still working.
+# A critical state represents a broken cluster.
+#
+# Copyright (c) 2017, InnoGames GmbH
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import subprocess
+
+from argparse import ArgumentParser
+from sys import exit
+
+
+def main():
+    """Main entrypoint for script"""
+
+    args = get_parser().parse_args()
+
+    master_addr, master_state, cluster_state_master = _get_cluster_status(
+        args.master_port, args.password)
+    slave_addr, slave_state, cluster_state_slave = _get_cluster_status(
+        args.slave_port, args.password)
+
+    if master_state != 'unknown' and slave_state != 'unknown':
+        if cluster_state_master != 'ok' and cluster_state_slave != 'ok':
+            print('CRITICAL - cluster is broken')
+            code = 2
+        elif master_state != 'master' or slave_state != 'slave':
+            print('WARNING - cluster status is degraded')
+            if master_state != 'master':
+                print('{} got demoted to slave'.format(master_addr))
+                code = 1
+            if slave_state != 'slave':
+                print('{} got promoted to master'.format(slave_addr))
+                code = 1
+        else:
+            print('OK - cluster status is OK')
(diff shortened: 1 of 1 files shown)


src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+def main():
+    """Main entrypoint for script"""
+
+    args = get_parser().parse_args()
+
+    output_master, state_master = _get_cluster_status(
+        args.port_master, args.auth)
+    output_slave, state_slave = _get_cluster_status(args.port_slave, args.auth)
+
+    code = 0
```
hasegeli: This is unnecessary.

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+            if output_master[1] != 'master':
+                print('{} got demoted to slave'.format(output_master[0]))
+                code = 1
+            if output_slave[1] != 'slave':
+                print('{} got promoted to master'.format(output_slave[0]))
+                code = 1
+        else:
+            print('OK - cluster status is OK')
+            print('{} - {}'.format(output_master[0], output_master[1]))
+            print('{} - {}'.format(output_slave[0], output_slave[1]))
```
hasegeli: It may be better not to output multiple lines.

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+def get_parser():
+    """Get argument parser -> ArgumentParser
+
+    We need ports and auth to connect
+    """
+
+    parser = ArgumentParser()
+
+    parser.add_argument(
+        '-p_master',
```
hasegeli: Long style arguments should start with 2 minuses.  Also `_` is ugly.  How about `--master-port`?

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+        action='store',
+        dest='port_master',
+        default='7000',
```
hasegeli: This can be `type=int` and `default=7000`.

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+    parser.add_argument(
+        '-p_slave',
+        action='store',
+        dest='port_slave',
+        default='7001',
+        help='redis port of the slave instance',
+    )
+    parser.add_argument(
+        '-auth',
```
hasegeli: Is this the password?  `--password` maybe?

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+
+    The status of the local instances will be checked
+    """
+
+    try:
+        command = subprocess.check_output(
+            'redis-cli -p {0} -a {1} cluster nodes | grep myself'.format(
+                port, auth), shell=True).split()
+        command[2] = command[2].replace('myself,', '')
+        command = command[slice(1, 3)]
```
hasegeli: You could write it as `command = command[1:3]`, but I would instead get those 2 separate things to different variables like `val1, val2 = command[1:3]`.  I am not sure how to name them, but it cannot get worse then `command`.

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+        state = subprocess.check_output(
+            'redis-cli -p {0} -a {1} cluster info | grep state'.format(
+                port, auth), shell=True).split(':')[1].strip()
+    except subprocess.CalledProcessError as err:
+        command = ['unknown']
+        state = 'unknown'
```
hasegeli: Just not excepting errors, would make this easier to debug.  What is the value of returning unknown in here?

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _get_cluster_status(port, auth):
+    """Get the Cluster Information
+
+    The status of the local instances will be checked
+    """
+
+    try:
+        command = subprocess.check_output(
+            'redis-cli -p {0} -a {1} cluster nodes | grep myself'.format(
```
hasegeli: Please do the `grep` on Python.  It is better not mixing Python with shell code.

src/check_redis_cluster.py
```diff
@@ -0,0 +1,137 @@
...
+                port, auth), shell=True).split()
+        command[2] = command[2].replace('myself,', '')
```
hasegeli: If this must be in the beginning of the command you can check `command[2].startswith('myself,')` and trim it with `x = x[len('myself,'):]`.

src/check_redis_cluster.py
lacunoc: good point, I'll change it

src/check_redis_cluster.py
lacunoc: alright

src/check_redis_cluster.py
lacunoc: also possible

src/check_redis_cluster.py
lacunoc: I'll give it a try

src/check_redis_cluster.py
lacunoc: from my point of view it is, because the overview shows the first line, and in detail (mouseover) you can see the output of the instances (master/slave)

src/check_redis_cluster.py
lacunoc: to set the cluster status check in an unknown state due to connection problems to at least one node.

src/check_redis_cluster.py
```diff
@@ -0,0 +1,145 @@
...
+    args = get_parser().parse_args()
+
+    master_addr, master_state, cluster_state_master = _get_cluster_status(
+        args.master_port, args.password)
+    slave_addr, slave_state, cluster_state_slave = _get_cluster_status(
+        args.slave_port, args.password)
+
+    if master_state != 'unknown' and slave_state != 'unknown':
+        if cluster_state_master != 'ok' and cluster_state_slave != 'ok':
+            print('CRITICAL - cluster status is broken')
```
hasegeli: "cluster status is broken" doesn't sound right to me.  Maybe "cluster is broken"?

src/check_redis_cluster.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    parser = ArgumentParser()
+
+    parser.add_argument(
+        '--master-port',
+        action='store',
+        dest='master_port',
+        type=int,
+        default='7000',
```
hasegeli: `argparse` do not cast the default values.  This should also be `7000` without the quotes.

PR #53, opened 2017-07-04
Provide new check for memory overallocation in FreeBSD

TuxPowered42:


src/check_freebsd_memory.py
```diff
@@ -0,0 +1,145 @@
...
+    'wired': 'v_wire',
+}
+
+# Speci
+NAGIOS_CODES = {
+    'OK': 0,
+    'WARNING': 1,
+    'CRITICAL': 2,
+    'UNKNOWN': 3,
+}
```
hasegeli: I would (and do all the time) make this a class:
```python
class ExitCodes:
    OK = 0,
    WARNING = 1,
...
```

src/check_freebsd_memory.py
```diff
@@ -0,0 +1,145 @@
...
+
+    print(exit_multiline)
+    return exit_code
+
+
+def parse_args():
+    parser = ArgumentParser()
+    for memory_type in MEMORY_TYPES.keys():
+        parser.add_argument(
+            '--{}-warning'.format(memory_type), default='50',
```
hasegeli: `argparse` do not cast the default values.  It shouldn't be in quotes.

src/check_freebsd_memory.py
```diff
@@ -0,0 +1,145 @@
...
+import sys
+
+
+# Translate human-readable names to sysctl
+MEMORY_TYPES = {
+    'active': 'v_active',
+    'wired': 'v_wire',
+}
+
+# Speci
```
hasegeli: ?

src/check_freebsd_memory.py
```diff
@@ -0,0 +1,145 @@
...
+    print(exit_multiline)
+    return exit_code
+
+
+def parse_args():
+    parser = ArgumentParser()
+    for memory_type in MEMORY_TYPES.keys():
+        parser.add_argument(
+            '--{}-warning'.format(memory_type), default='50',
+            required=True, metavar='%', type=int,
```
hasegeli: Required is unnecessary with a default, I believe.

src/check_freebsd_memory.py
TuxPowered42: Fixed.

src/check_freebsd_memory.py
TuxPowered42: Fixed.

PR #61, opened 2017-09-08
Add check_mysql_conf.py to replace the AWK script

hasegeli:


src/check_mysql_conf.py
```diff
@@ -0,0 +1,93 @@
...
+    exit_code = max(p.wait() for p in check_procs)
+    if exit_code == 0:
+        print('OK: pt-config-diff found no difference')
+    else:
+        print('WARNING: pt-config-diff failed:')
+        for proc in check_procs:
+            stdout = proc.stdout.read()
+            if stdout:
+                print(stdout.decode())
+    exit(exit_code)
```
TheFRedFox: Would suggest a blank line before the `exit` to separate it a bit from the rest of the code.

Didn't we also said to use rather `return` or `raise SystemExit`?

PR #63, opened 2017-09-15
Migrations for upcoming Gitlab update

TheFRedFox:
Script has a new argument called `api-version` with which the user can specify the api version.  If not specified the script will try to determine the api version by trying to reach certain endpoints.

All other endpoints which are used in the script should still work.

So the PR can directly be merged, as it will support both api versions: `v3` and the upcoming `v4`.

### DIFF
--- a/src/gitlab_unprotected_branches.py
+++ b/src/gitlab_unprotected_branches.py
@@ -152,8 +152,7 @@ def main(args):
         if branch and
         (
             not branch['protected'] or
-            branch['developers_can_merge'] or
-            branch['developers_can_push']
+            branch['developers_can_merge'] or branch['developers_can_push']
         )
     ]
 
--- a/src/gitlab_unprotected_branches.py
+++ b/src/gitlab_unprotected_branches.py
@@ -115,6 +115,7 @@ def main(args):
     else:
         projects = fetch_projects(base_url, api, args.project, auth=auth)
 
+    projects = list(filter(None, projects))
     if not projects:
         print('No projects found')
         exit(3)
--- a/src/gitlab_unprotected_branches.py
+++ b/src/gitlab_unprotected_branches.py
@@ -106,14 +106,14 @@ def main(args):
         exit(3)
 
     if args.group:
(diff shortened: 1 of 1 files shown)


src/gitlab_unprotected_branches.py
```diff
@@ -191,6 +207,21 @@ def parse_auth_argument(args):
...
 
+def fetch_api_version(base_url, auth=None):
+    versions = ('v3', 'v4')
+    version = None
+    for version in reversed(versions):
+        endpoint = '/api/{api}/version'.format(api=version)
+        response = do_request('head', base_url, endpoint, auth=auth)
+
+        if response.ok:
+            break
```
hasegeli: Why don't you `return version` in here?

src/gitlab_unprotected_branches.py
```diff
@@ -191,6 +207,21 @@ def parse_auth_argument(args):
     return auth
 
 
+def fetch_api_version(base_url, auth=None):
+    versions = ('v3', 'v4')
+    version = None
```
hasegeli: This is unnecessary.  `for` is going to define it.

src/gitlab_unprotected_branches.py
```diff
@@ -83,6 +84,12 @@ def parse_args():
     parser.add_argument('--branch',
                         help='the branch which should be protected, '
                              'default is the default branch of the project')
+    parser.add_argument(
+        '--api',
```
hasegeli: I would prefer `--api-version`.

src/gitlab_unprotected_branches.py
```diff
@@ -83,6 +84,12 @@ def parse_args():
...
+        help=("The api version in form of 'v3'.  If not specified the script "
+              "tries to fetch current api version through the gitlab version "
+              "endpoint.  Right now it will try from api version 4 down to 3.")
```
hasegeli: Why do you use double quotes in here?  I think double quotes are more appropriate in English language.  You would be better off using them inside the string.

PR #69, opened 2017-11-20
Use systemd if found instead of testing for os version

juergenthomann:


src/check_smartmon.py
```diff
@@ -106,43 +154,45 @@
...
+    CSV_PATH = conf_data['CSV_PATH']
     HDD_NAMES = conf_data['HDD_NAMES']
-    ALERTS    = conf_data['ALERTS']
+    ALERTS = conf_data['ALERTS']
 except IOError as e:
     pass
 
-os_major = platform.dist()[1].split('.')[0]
-if os_major == "8":
+if os.path.exists('/bin/systemctl'):
```
TheFRedFox: Is this really the only location where `systemctl` could be? If there are more possibilities, maybe another solution could be to try to run the command and check the exit status?

src/check_smartmon.py
```diff
@@ -160,12 +210,18 @@
         smart_m = int(smart_date.split(":")[1])
         smart_s = int(smart_date.split(":")[2])
 
-        future = datetime.datetime.now() - datetime.datetime(smart_year, smart_mon, smart_day, smart_h, smart_m, smart_s)
-        smart_date_in_seconds = int(future.days * 24 * 60 * 60 + future.seconds)
+        future = datetime.datetime.now() - datetime.datetime(smart_year,
+                                                             smart_mon,
+                                                             smart_day,
+                                                             smart_h, smart_m,
+                                                             smart_s)
```
TheFRedFox: That's ugly, rather:
``` python
from datetime import datetime

future = datetime.now() - datetime(
    smart_year, smart_mon, smart_day, smart_h, smart_m, smart_s
)
```

Or maybe also store the second datetime in an own variable.

src/check_smartmon.py
```diff
@@ -160,12 +210,18 @@
...
+        smart_date_in_seconds = int(
+            future.days * 24 * 60 * 60 + future.seconds)
 
         # Throw error if csv data is more than 6 hours old!
         if smart_date_in_seconds > 6 * 60 * 60:
-            senddata.append("SMART-CSV-Data is > %s hours old! (%s%s)\n" % (str(smart_date_in_seconds/3600), CSV_PATH, csvfile))
+            senddata.append("SMART-CSV-Data is > %s hours old! (%s%s)\n" % (
```
TheFRedFox: Please use `format` method.

src/check_smartmon.py
```diff
@@ -175,45 +231,63 @@
...
                 else:
                     smart_value = int(element[2])
             else:
                 smart_value = int(element[1])
 
             if not int(smart_id) in ALERTS[manufacturer]:
-                senddata.append("Error: No id %s in %s-config found. Please configure this value.\n" % (smart_id, manufacturer))
+                senddata.append(
+                    "Error: No id %s in %s-config found. "
+                    "Please configure this value.\n" % (
```
TheFRedFox: Please use `format` method.

src/check_smartmon.py
```diff
@@ -175,45 +231,63 @@
...
+            if not ('or' in ALERTS[manufacturer][smart_id] and smart_value ==
+                    ALERTS[manufacturer][smart_id]['or']):
                 if smart_value < ALERTS[manufacturer][smart_id]['min']:
-                    senddata.append( "%s: [id=%s] '%s' is '%s' which is lower than '%s'\n" % ( 
-                        manufacturer, smart_id, ALERTS[manufacturer][smart_id]['description'], smart_value, ALERTS[manufacturer][smart_id]['min']))
-                    if exit_status < ALERTS[manufacturer][smart_id]['exit_status']:
-                        exit_status = ALERTS[manufacturer][smart_id]['exit_status']
+                    senddata.append(
+                        "%s: [id=%s] '%s' is '%s' which is lower than "
+                        "'%s'\n" % (
```
TheFRedFox: Please use `format` method.

src/check_smartmon.py
```diff
@@ -175,45 +231,63 @@
...
+                            manufacturer, smart_id,
+                            ALERTS[manufacturer][smart_id]['description'],
+                            smart_value,
+                            ALERTS[manufacturer][smart_id]['min']))
+                    if exit_status < \
```
TheFRedFox: Please use braces to cover multiline python statements.

src/check_smartmon.py
```diff
@@ -175,45 +231,63 @@
...
+                            'exit_status']
 
                 if smart_value > ALERTS[manufacturer][smart_id]['max']:
-                    senddata.append( "%s: [id=%s] '%s' is '%s' which is higher than '%s\n" % (
-                        manufacturer, smart_id, ALERTS[manufacturer][smart_id]['description'], smart_value, ALERTS[manufacturer][smart_id]['max']))
-                    if exit_status < ALERTS[manufacturer][smart_id]['exit_status']:
-                        exit_status = ALERTS[manufacturer][smart_id]['exit_status']
+                    senddata.append(
+                        "%s: [id=%s] '%s' is '%s' which is higher than "
+                        "'%s\n" % (
```
TheFRedFox: Please use `format` method.

src/check_smartmon.py
```diff
@@ -175,45 +231,63 @@
...
+                            manufacturer, smart_id,
+                            ALERTS[manufacturer][smart_id]['description'],
+                            smart_value,
+                            ALERTS[manufacturer][smart_id]['max']))
+                    if exit_status < \
```
TheFRedFox: Please use braces to cover multiline python statements.

src/check_smartmon.py
```diff
@@ -175,45 +231,63 @@
...
+if 0 < kingston_gb_written < 99:
     for i in range(len(senddata)):
         if 'KINGSTON: [id=13]' in senddata[i]:
             del senddata[i]
     if len(senddata) == 0:
         exit_status = 0
 
 if exit_status == 1:
-    print "WARNING: " + " ".join(senddata)
+    print("WARNING: " + " ".join(senddata))
```
TheFRedFox: Please use `format` method.

src/check_smartmon.py
```diff
@@ -175,45 +231,63 @@
...
 elif exit_status == 2:
-    print "CRITICAL: " + " ".join(senddata)
+    print("CRITICAL: " + " ".join(senddata))
```
TheFRedFox: Please use `format` method.

src/check_smartmon.py
```diff
@@ -160,12 +210,18 @@
...
 
-        future = datetime.datetime.now() - datetime.datetime(smart_year, smart_mon, smart_day, smart_h, smart_m, smart_s)
-        smart_date_in_seconds = int(future.days * 24 * 60 * 60 + future.seconds)
+        future = datetime.datetime.now() - datetime.datetime(smart_year,
+                                                             smart_mon,
+                                                             smart_day,
+                                                             smart_h, smart_m,
+                                                             smart_s)
+        smart_date_in_seconds = int(
+            future.days * 24 * 60 * 60 + future.seconds)
```
TheFRedFox: `24 * 60 * 60` could be stored in an own variable also as constant.

src/check_smartmon.py
```diff
@@ -35,69 +37,115 @@
 #   /etc/nagios-plugins/check_smartmon.cfg
 # on each server (distributed by puppet)!
 
-### DEFAULT CONFIG VALUES ### (if /etc/nagios-plugins/check_smartmon.cfg doesn't exist!)
+### DEFAULT CONFIG VALUES ###
+# if /etc/nagios-plugins/check_smartmon.cfg doesn't exist!
 #
 # CSV_PATH: This value should never change
-CSV_PATH="/var/lib/smartmontools/"
+CSV_PATH = "/var/lib/smartmontools/"
```
TheFRedFox: Single-quote.

src/check_smartmon.py
juergenthomann: At least on Debian like systems it is the only place. To support other systems we have to check for a different unit name as well.

TheFRedFox: Ok, guess when there are more dependencies than just the path, then it is fair enough to keep it like that.

TuxPowered42:
It would be way better to not duplicate functionality of other checks. On systems with systemd we have check_systemd_units which will verify if *all* daemons are running. On older systems it is enough to check csv files are not fresh which should be good enough. And we hopefully don't have such systems anymore.

PR #70, opened 2017-11-29
check_dns_sshfp.py: Re-enable not exists warning

hasegeli:
This was disabled because of a bug on Debian Wheezy servers.

### DIFF
--- a/src/check_dns_sshfp.py
+++ b/src/check_dns_sshfp.py
@@ -55,8 +55,7 @@ def main():
         record_id = (record.algorithm, record.fp_type)
         if record_id not in host_keys:
             print('WARNING host key for fingerprint does not exist')
-            # exit(1) TODO: Enable this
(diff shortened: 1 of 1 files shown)


src/check_dns_sshfp.py
```diff
@@ -55,8 +55,7 @@ def main():
         record_id = (record.algorithm, record.fp_type)
         if record_id not in host_keys:
             print('WARNING host key for fingerprint does not exist')
-            # exit(1) TODO: Enable this
-            exit(0)
```
TheFRedFox: Why was it like that in first place? ^^

hasegeli: `ssh-keygen -r` was not working on some older servers.

TheFRedFox: I see. Ok. ^^

PR #73, opened 2018-01-10
Check if 'maintenanceExpiryDate' key exists

TheFRedFox:
Script broke as there was a plugin which seems to use licensing however
has no information about the maintenance expiry date.  Therefore the
script will now check for the key before it continues the process.

### DIFF
--- a/src/atlassian_expiring_licenses.py
+++ b/src/atlassian_expiring_licenses.py
@@ -109,8 +109,10 @@ def main(args):
     expires = [
         (plugin, response.json())
         for plugin, response in responses
-        if response and datetime.utcfromtimestamp(
-            response.json()['maintenanceExpiryDate'] / 1000) < deadline
+        if response and 'maintenanceExpiryDate' in response.json()
+        and datetime.utcfromtimestamp(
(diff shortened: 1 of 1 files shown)


src/atlassian_expiring_licenses.py
```diff
@@ -98,19 +98,21 @@ def main(args):
...
 
     expires = [
         (plugin, response.json())
         for plugin, response in responses
-        if response and datetime.utcfromtimestamp(
-            response.json()['maintenanceExpiryDate'] / 1000) < deadline
+        if response and hasattr(response.json(), 'maintenanceExpiryDate')
+        and datetime.utcfromtimestamp(
+            response.json()['maintenanceExpiryDate'] / 1000
+        ) < deadline
```
hasegeli: I think it is time to turn this incomprehensible list comprehension into a comprehensible straightforward for loop.  Casting zip into list, and calling `.json()` two times is not cool.

hasegeli: I think it is time to turn this incomprehensible list comprehension into a comprehensible straightforward for loop. Casting zip into list, and calling .json() two times is not cool.

PR #76, opened 2018-01-17
check_process_list.py: Improve CPU time scaling

hasegeli:
The `timedelta` object on Python2 doesn't support division.  Because of that I had to cast the both sides to `int` which harms precision and causes `DivisionByZero` when the script is called again quickly.  We can improve this and made the code cleaner once we don't support Python 2 anymore.

### DIFF
--- a/src/check_process_list.py
+++ b/src/check_process_list.py
@@ -315,11 +315,8 @@ class Process(dict):
             return None
 
         diff = value - self.prev_values[var]
-        if isinstance(diff, timedelta):
-            # TODO: Don't use .total_seconds()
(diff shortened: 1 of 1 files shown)


TheFRedFox:
You want to support both right now and remove Python2 code after we don't support it anymore?

hasegeli:
Yes

hasegeli:
Hey, can you merge this?  It's been a while.

kofrezo:
Just rebased this with master (force-push)

kofrezo (approved):

PR #78, opened 2018-02-28
Add file size checker

seqizz:


src/check_logfilesize.py
```diff
@@ -0,0 +1,55 @@
...
+    elif warning_count > 0:
+        print("Warning")
+        print(files)
+        sys.exit(1)
+    else:
+        print("OK")
+        sys.exit(0)
+
+
+def file_size_mb(filePath): return os.path.getsize(filePath) / (1024 * 1024)
```
hasegeli: PEP8 says "don't do it if you are not trying to hide anything".

hasegeli: You may also write `1024**2`.

src/check_logfilesize.py
```diff
@@ -0,0 +1,55 @@
...
+args = parser.parse_args()
+
+def main():
+    critical_count = 0
+    warning_count = 0
+    files = ""
+
+    for file in os.listdir(args.path):
+        if file.endswith(".log"):
+            current_file_size = file_size_mb(args.path+"/"+file)
```
hasegeli: Spaces around opeators

src/check_logfilesize.py
```diff
@@ -0,0 +1,55 @@
...
+import argparse
+
+parser = argparse.ArgumentParser(
+    description="A small nagios script to check "
+                "for size of logfiles in given directory")
+
+parser.add_argument("-p", dest="path", help="Define the directory in which the check should run")
+parser.add_argument("-w", dest="warning", help="Define the warning threshold in MB")
+parser.add_argument("-c", dest="critical", help="Define the critical threshold in MB")
+args = parser.parse_args()
```
hasegeli: It would be nice to make this `parse_args()` function and call from `main()`.

src/check_logfilesize.py
```diff
@@ -0,0 +1,55 @@
...
+
+def main():
+    critical_count = 0
+    warning_count = 0
+    files = ""
+
+    for file in os.listdir(args.path):
```
hasegeli: `file` might be reserved.  `filename` might be better.

seqizz:
Should be better now

PR #83, opened 2018-04-05
Some pep8 and grep sanity for check_cron_runtime

seqizz:
My project's name includes "grep" word, so `grep -v grep` is not good :smile:

### DIFF
--- a/src/check_cron_runtime.py
+++ b/src/check_cron_runtime.py
@@ -2,7 +2,7 @@
 #
 # InnoGames Monitoring Plugins - check_cron_runtime.py
 #
-# Copyright (c) 2016, InnoGames GmbH
+# Copyright (c) 2018, InnoGames GmbH
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the 'Software'), to deal
@@ -32,6 +32,7 @@ exclude_list = [
     'am_stockpile_distribution',
 ]
 
+
 def parse_args():
     """The argument parser"""
 
@@ -47,11 +48,11 @@ def parse_args():
     )
     parser.add_argument(
         '--verbose',
-        dest='verbose',
         action='store_true',
         help='Increase verbosity',
     )
     return vars(parser.parse_args())
+
 
 def main(verbose=False):
     """The main program"""
@@ -59,20 +60,18 @@ def main(verbose=False):
     # Dictionary of pids containing th command that is running
     ps_aux = get_ps_aux()
 
-    #pids = psutil.pids()
-
-    #Nagios Return codes
+    # Nagios Return codes
     ok_code = 0
     warn_code = 1
     crit_code = 2
     un_code = 3
 
-    #pre check ps version
+    # pre check ps version
     if '3.2.8' in execute('ps --version').strip('\n').split(' ')[2]:
         print('PS version 3.2.9 or above needed to run this script')
         exit(int(un_code))
 
-    #Check for main Cron, Return CRITICAL if not running
+    # Check for main Cron, Return CRITICAL if not running
     try:
         parent_cron_pid = execute('pgrep -o cron').strip('\n')
     except AttributeError:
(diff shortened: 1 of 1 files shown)


src/check_cron_runtime.py
```diff
@@ -53,26 +54,27 @@ def parse_args():
...
 
+
 def main(verbose=False):
     """The main program"""
 
     # Dictionary of pids containing th command that is running
     ps_aux = get_ps_aux()
 
-    #pids = psutil.pids()
+    # pids = psutil.pids()
```
TheFRedFox: Looks like old obsolete code. Can't we just remove the line?

src/check_cron_runtime.py
```diff
@@ -101,18 +103,20 @@ def main(verbose=False):
...
     cron_pid_more_than1 = []
-    ok_flag = warn_flag = un_flag = 0
+    ok_flag = warn_flag = 0
     for pid in cron_child_list:
         if pid:
             try:
-                cron_child_time = execute("ps -o 'etimes=' " + pid).strip('    ').strip('\n')
+                cron_child_time = execute(
+                    "ps -o 'etimes=' " + pid
+                ).strip('    ').strip('\n')
```
TheFRedFox: Personal suggestion as the `execute` does some process stuff and I recommend to do this stuff separated, so you can easier check if there is an error during runtime:
``` python
cron_child_time = execute(...)
cron_child_time = cron_child_time.strip(...).strip(...)
```

src/check_cron_runtime.py
```diff
@@ -152,13 +157,17 @@ def get_ps_aux():
...
     return pids
 
+
 def get_child(ppid):
     childs = execute('pgrep -P ' + ppid)
     if childs:
         for child in childs.strip('\n').split():
-       	    print '|' + (execute('ps -ef | grep ' + child + ' | grep -v grep'))
+            print(
+                '|' + (execute('ps -ef | grep ' + child + ' | grep -wv grep'))
```
TheFRedFox: To make it more readable, please store the `execute` result first and also use more `str.format`.

TheFRedFox:
And at line 141 there is a python 2 `print`. So add some braces to make it a method call for python 3 support. 😉

seqizz:
@TheFRedFox Please check again, I force pushed with requested changes. But seems like Github UI does not care :man_facepalming:

src/check_cron_runtime.py
```diff
@@ -152,13 +156,20 @@ def get_ps_aux():
...
 def get_child(ppid):
     childs = execute('pgrep -P ' + ppid)
     if childs:
         for child in childs.strip('\n').split():
-       	    print '|' + (execute('ps -ef | grep ' + child + ' | grep -v grep'))
+            exec_out = execute(
+                    "ps -ef | grep {} | grep -wv grep".format(child)
+                )
+            print(
+                '|' + exec_out
```
TheFRedFox: Can be in one line then.

src/check_cron_runtime.py
```diff
@@ -101,18 +101,21 @@ def main(verbose=False):
...
 
     cron_pid_more_than1 = []
-    ok_flag = warn_flag = un_flag = 0
+    ok_flag = warn_flag = 0
     for pid in cron_child_list:
         if pid:
             try:
-                cron_child_time = execute("ps -o 'etimes=' " + pid).strip('    ').strip('\n')
+                cron_child_time = execute(
+                    "ps -o 'etimes=' " + pid
```
TheFRedFox: Can be in one line then.

src/check_cron_runtime.py
```diff
@@ -131,18 +134,19 @@ def main(verbose=False):
                 print('UNKNOWN PID execution TIME')
                 exit(int(un_code))
 
-    #print cron_pid_more_than1
+    # print cron_pid_more_than1
```
TheFRedFox: Just saw that, seems also obsolete code.

PR #84, opened 2018-04-09
Add a json check that can validate and compare arbitrary JSON

Woellchen:


src/check_json.py
```diff
@@ -0,0 +1,240 @@
...
+
+    @classmethod
+    def get_code(cls, status):
+        """Return the exit code for a written nagios status"""
+
+        status = status.upper()
+        if status not in cls._code_map:
+            raise ValueError('Status "{}" does not exist'.format(status))
+
+        return cls._code_map[status]
```
hasegeli: There are really nice enums on Python 3: https://docs.python.org/3/library/enum.html

src/check_json.py
```diff
@@ -0,0 +1,240 @@
...
+            return None
+
+        return loads(content)
+
+    def run(self):
+        """Run the checks"""
+        try:
+            data = self.load()
+        except JSONDecodeError:
+            return ExitCodes.CRITICAL, 'Invalid JSON'
```
hasegeli: May the error be more descriptive than ours?

src/check_json.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def _execute_checks(self, severity, data):
+        """Execute all checks of given severity"""
+        reasons = []
+
+        for check in self.checks[severity]:
+            if check(data):
+                reasons.append(str(check))
+
+        return ', '.join(reasons) if len(reasons) else ''
```
hasegeli: `if len(reasons)` is not Pythonic, `if reasons` is.  You don't need it in here anyway.

src/check_json.py
Woellchen: like? what else could i say to broken json?

hasegeli: Like:
```
>>> loads('{"a", [}')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/pkg/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/opt/pkg/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/opt/pkg/lib/python2.7/json/decoder.py", line 380, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Expecting : delimiter: line 1 column 5 (char 4)
```

Maybe it is not better though.

Woellchen: ok see your point but this is a nagios check and not a json validator. if its invalid it should turn critical and do not spam nagios with stacktraces imo

TheFRedFox: But should it be really Critical or rather Unknown?

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+        return loads(content)
+
+    def run(self):
+        """Run the checks"""
+        try:
+            data = self.load()
+        except JSONDecodeError:
+            return ExitCodes.CRITICAL, 'Invalid JSON'
+        except (RequestException, IOError):
+            return ExitCodes.CRITICAL, 'Could not load data'
```
TheFRedFox: Also here: Critical or rather Unknown?

src/check_json.py
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/env python3
+
+"""InnoGames Monitoring Plugins - check_json.py
+
+This Python Nagios check can validate and compare arbitrary JSON.
+
+The check can either validate JSON string arguments, files and URLs.
+To check certain values, JSONPath (http://jsonpath.com/) is used.
+
+Operators: ~=, ==, !=, <=, >=, <, >
```
TheFRedFox: Would also like to see an example with `~=`.

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
+from enum import Enum
+from json import loads, JSONDecodeError
+from jsonpath_rw import parse
+from os.path import isfile
+from re import compile
```
TheFRedFox: I would rather import it like `import re.compile` as `compile` isolated could be misleading.

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __str__(self):
+        return '{} {} {}'.format(self.key, self.operator, self.value)
+
+    def __call__(self, data):
+        """Execute the check itself"""
+        path = parse(self.key)
+        matches = path.find(data)
+
+        if not len(matches):
```
TheFRedFox: `if not matches:` is enough here

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+            return True
+
+        return self.executor(matches[0].value)
+
+    @classmethod
+    def from_string(cls, pair):
+        """Parse DSL from given arguments"""
+        for symbol in sorted(cls.executors.keys(), key=len, reverse=True):
```
TheFRedFox: Do not really get why you are sorting the executors before looping over them?

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+    OK = 0
+    WARNING = 1
+    CRITICAL = 2
+    UNKNOWN = 3
+
+
+class Check:
+    """Check consists of the variable name, operator, and a value"""
+    executors = {
+        '~=': lambda b: compile(b).match,
```
TheFRedFox: Maybe you can compile the regex directly in `from_string` to directly check if the regex is correct.

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+            content = self._load_file(self.source)
+        elif url(self.source):
+            content = self._load_url(self.source)
+        else:
+            content = self.source
+
+        if content is None:
+            return None
+
+        return loads(content)
```
TheFRedFox: I personally also find `loads` from `json.loads` ambiguous and like to rather go with `import json` or the like.

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+            return ExitCodes.CRITICAL, 'Invalid JSON'
+        except (RequestException, IOError):
+            return ExitCodes.CRITICAL, 'Could not load data'
+
+        # if we have criticals, we need to prioritize them, obviously
+        for severity in [ExitCodes.CRITICAL, ExitCodes.WARNING]:
+            reasons = self._execute_checks(severity, data)
+
+            if reasons:
+                return severity, reasons
```
TheFRedFox: So not all checks will be done, but directly will be returned by the first finding? Dunno, how we need it, but maybe it can be optional to also check all checks?

src/check_json.py
```diff
@@ -0,0 +1,222 @@
...
+    @staticmethod
+    def _load_file(source):
+        """Load contents from file"""
+        with open(source) as f:
+            return f.read()
+
+    @staticmethod
+    def _load_url(source):
+        """Load contents from URL"""
+        response = request('GET', source, timeout=3000)
```
TheFRedFox: 3 seconds are enough? Could also be made configurable via script parameter.

src/check_json.py
Woellchen: if there is a critical condition met, it should be returned before even evaluating warnings (highest severity wins)
there is no reason and no possibility to report both a critical and a warning state

src/check_json.py
Woellchen: pep8 recommends absolute imports

src/check_json.py
Woellchen: why? compile is ambigious so there is nothing that could possibly mislead. also i find consistency in the notation much more valuable

src/check_json.py
Woellchen: done

src/check_json.py
TheFRedFox: That's right, but pep8 also recommends to avoid ambiguous names. Both can't be served here and I would go for the better readable and better maintainable version, which IMHO is to use `json.loads`.

TheFRedFox: But it is just my personal suggestion, so feel free to keep it. 😉

src/check_json.py
TheFRedFox: What about showing check as Critical but showing all errors, critical and warning ones?

PR #87, opened 2018-04-23
logrotation: Add nagios check

lacunoc:
We need a nagios check, which is searching for logs
which are not logrotated by our configuration but
lying around in the regarding directories.

### DIFF
--- a/src/check_logrotate_missing.py
+++ b/src/check_logrotate_missing.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python
+#
+# InnoGames Monitoring Plugins - check_logrotate_missing.py
+#
+# Search for logs which are not logrotated
+#
+# This script checks that only the configured logfiles for logrotation
+# are lying in the regarding log directories.
+# Following logrotate configuration filetypes are supported:
+# * logrotate
+# * logback
+# * log4j (xml only)
+# It raises a warning if there are unconfigured logfiles found.
+# It raises a unknown if the passed configuration file does not exist.
+#
+# Copyright (c) 2018, InnoGames GmbH
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import glob
+import os
+import mimetypes
+import xml.etree.ElementTree as ET
+from argparse import ArgumentParser, RawTextHelpFormatter
+
+
+class RotationConfigDoesNotExist(Exception):
+    pass
+
+
+def main():
+    """Main entrypoint for script"""
+
+    args = parse_args()
+
+    try:
+        logrotate_missing = check_logrotation_config(args.config, args.exclude)
+    except RotationConfigDoesNotExist as e:
+        code = 3
+        print(e)
+        exit(code)
+
+    code = 0
+
+    if logrotate_missing:
+        code = 1
+        for logfile in logrotate_missing:
+            print(logfile)
+    else:
+        print('OK')
+
+    exit(code)
+
+
+def parse_args():
+    """Get argument parser -> ArgumentParser
+
+    We need the logrotation configuration files
+    and potentially logfiles to exclude
+    """
+
+    parser = ArgumentParser(formatter_class=RawTextHelpFormatter)
+
+    parser.add_argument(
+        '--config',
+        help='''
+        logrotation config to parse,
+        multiple possible
+        EXAMPLE: --config /etc/logrotate.d/foe-logs /etc/logrotate.d/foe-logs2
+        ''',
+        nargs='*',
+        type=str,
+    )
+
+    parser.add_argument(
+        '--exclude',
+        help='''
+        logfiles to exclude from being checked,
+        multiple possible
+        EXAMPLE: --exclude /var/log/lastlog /var/log/faillog
+        ''',
+        nargs='*',
+        type=str,
+        default=''
+    )
+
+    return parser.parse_args()
+
+
+def check_logrotation_config(config, exclude):
+    """Check the logrotation configuration
+
+    Check configured logrotation directories if there are unconfigured logs
+    lying around.
+
+    :param: config: logrotation config file like:
+                    /etc/logrotate.d/foe-logs
+                    /www/gk/configs/gk-master-admin/logback.xml
+                    /www/gk/gk-master/configs/log4j2.xml
+
+    :param: exclude: logfile to exclude from being checked,
+                     e.g. /var/log/lastlog
+
+    :return: set
+    """
+
+    configured_logs = set()
+    unconfigured_logs = set()
+    log4j_appenders = [
+        './Appenders/File',
+        './Appenders/RollingFile',
+        './Appenders/ReopenFile',
+    ]
+
+    for conf in config:
+        conf_type = mimetypes.guess_type(conf)[0]
+
+        if conf_type and 'xml' in conf_type:
+            try:
+                tree = ET.parse(conf)
+            except IOError:
+                raise RotationConfigDoesNotExist('{} does not exist'.format(
+                    conf))
+            # logback
+            for lblog in tree.getroot().findall('appender'):
+                    lblog_file = lblog.find('file')
+                    if lblog_file is not None:
+                        configured_logs.add(lblog_file.text)
+            # log4j
+            for appender in log4j_appenders:
+                for jlog in tree.getroot().findall(appender):
+                    jlog_file = jlog.get('fileName')
+                    if jlog_file is not None:
+                        configured_logs.add(jlog_file)
+        # logrotate
+        else:
+            try:
+                with open(conf) as in_file:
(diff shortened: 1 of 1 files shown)


src/check_logrotate_missing.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env python
```
meikomeik: Should be python3 I think

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,166 @@
...
+def get_parser():
+    """Get argument parser -> ArgumentParser
+
+    We need the logrotation configuration files
+    """
+
+    parser = ArgumentParser()
+
+    parser.add_argument(
+        '--config', '-C', help='logrotation config to parse',
```
ragged: Maybe add a better description of how the parsed logfiles shall look like, as it is not clear that you can add several logrotation configs

src/check_logrotate_missing.py
lacunoc: Why should it? it's fully compatible to python2

src/check_logrotate_missing.py
lacunoc: I will change this by using RawTextHelpFormatter as formatter_class of ArgumentParser. This makes it possible to print multiline help messages.

src/check_logrotate_missing.py
meikomeik: You are right, ignore my comment

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+import glob
+import os
+import re
+import mimetypes
+import xml.etree.ElementTree as ET
+from argparse import ArgumentParser, RawTextHelpFormatter
+
+
+class RotationConfigDoesNotExist(Exception):
+    pass
```
hasegeli: I don't see the advantage of it over `IOError`.

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+            try:
+                tree = ET.parse(conf)
+            except IOError:
+                raise RotationConfigDoesNotExist('{} does not exist'.format(
+                    conf))
+            # logback
+            for lblog in tree.getroot().findall('appender'):
+                try:
+                    configured_logs.append(lblog.find('file').text)
+                except AttributeError:
```
hasegeli: What causes the `AttributeError` in here?

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+                    pass
+            # log4j
+            for appender in log4j_appenders:
+                try:
+                    for jlog in tree.getroot().findall(appender):
+                        configured_logs.append(jlog.get('fileName'))
+                except AttributeError:
```
hasegeli: And in here?

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+                    pass
+        # logrotate
+        else:
+            try:
+                with open(conf, 'r') as in_file:
+                    for line in in_file:
+                        if re.match('^\/[a-z]+\/.*log', line):
```
hasegeli: Shouldn't you escape `.` too?

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+            for appender in log4j_appenders:
+                try:
+                    for jlog in tree.getroot().findall(appender):
+                        configured_logs.append(jlog.get('fileName'))
+                except AttributeError:
+                    pass
+        # logrotate
+        else:
+            try:
+                with open(conf, 'r') as in_file:
```
hasegeli: `'r'` is the default which we omit everywhere.

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+                    for line in in_file:
+                        if re.match('^\/[a-z]+\/.*log', line):
+                            configured_logs.append(
+                                line.replace('{', '').strip())
```
hasegeli: Something like `line.rstrip('{\n\t ')` would be nicer.

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+            except IOError:
+                raise RotationConfigDoesNotExist('{} does not exist'.format(
+                     conf))
+
+        log_dirs = set()
+        for c_log in configured_logs:
+            log_dirs.add(os.path.dirname(c_log))
```
hasegeli:
```python
log_dirs = {os.path.dirname(c) for c in configured_logs}
```

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+
+        for log_dir in log_dirs:
+            logfiles = glob.glob(log_dir + '/*log')
+            for logfile in logfiles:
+                if any(logfile in l for l in configured_logs):
```
hasegeli: Shouldn't the condition be `logfile == l` instead of `logfile in l`?

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+                    pass
+                else:
+                    unconfigured_logs.add(logfile)
```
hasegeli: `if: pass else:` is weird.  You can just write `if not:`

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,171 @@
...
+
+        for c_log in configured_logs:
+            if c_log in unconfigured_logs:
+                unconfigured_logs.remove(c_log)
```
hasegeli:
```python
unconfigured_logs -= configured_logs
```

src/check_logrotate_missing.py
hasegeli: Or even `set(logfiles) - configured_logs`

src/check_logrotate_missing.py
lacunoc: there might be no appender configured in the logback configuration

src/check_logrotate_missing.py
lacunoc: same for log4j

src/check_logrotate_missing.py
lacunoc: This wouldn't match a defined logfile like `/this/is/my/log_dir/logfile.log` then

src/check_logrotate_missing.py
hasegeli: Then you wouldn't loop over anything.  Am I missing something?

src/check_logrotate_missing.py
hasegeli: `.` means match with any character on regexp.  You have to type `\.` when you want to match with actual `.` character.

lacunoc: I do not want to match `.` character, there might be logs with `lognamelog` instead of `logname.log`

src/check_logrotate_missing.py
lacunoc: will be changed

src/check_logrotate_missing.py
lacunoc: will be changed

src/check_logrotate_missing.py
lacunoc: even nicer, thanks. will be changed

src/check_logrotate_missing.py
lacunoc: ofc, changed

src/check_logrotate_missing.py
lacunoc: `TypeError: unsupported operand type(s) for -=: 'set' and 'list'`

src/check_logrotate_missing.py
hasegeli: Ah yes, you are right.

hasegeli: It is a very loose regexp though.

src/check_logrotate_missing.py
lacunoc: ah sorry, there could be appenders found, which do not have a `file` component configured.

hasegeli: I guess then `.find()` would return `None`.  Can you save it and `if x: configured_logs.append(x.text)`?  This is safer and nicer coding.  The reader sees what can happen.

src/check_logrotate_missing.py
hasegeli: You can make `configured_logs` a `set` too.

src/check_logrotate_missing.py
lacunoc: good catch! changed

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,168 @@
...
+        else:
+            try:
+                with open(conf) as in_file:
+                    for line in in_file:
+                        if re.match('^\/[a-z]+\/.*log', line):
+                            configured_logs.append(
+                                line.rstrip('{\n\t '))
+            except IOError:
+                raise RotationConfigDoesNotExist('{} does not exist'.format(
+                     conf))
```
hasegeli: I would split the above part of the function to another one like `read_config`.

src/check_logrotate_missing.py
lacunoc: indeed -> changed

src/check_logrotate_missing.py
lacunoc: any better fitting regexp in mind?

src/check_logrotate_missing.py
lacunoc: I will think about this

src/check_logrotate_missing.py
lacunoc: I would like to keep the custom exception, if you don't mind.

src/check_logrotate_missing.py
hasegeli: Maybe just `line.startswith('/')`?

src/check_logrotate_missing.py
lacunoc: I will give it a try

lacunoc: actually it's only working as expected when using `if x is not None:`

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,177 @@
...
+        logrotation config to parse,
+        multiple possible
+        EXAMPLE: --config /etc/logrotate.d/foe-logs /etc/logrotate.d/foe-logs2
+        ''',
+        nargs='*',
+        type=str,
+    )
+
+    parser.add_argument(
+        '--exclude', '-X',
```
hasegeli: Why support short arguments?  Even though they are more standard, I am in favor of providing a single better way on our scripts.  As the Zen of Python says: `There should be one-- and preferably only one --obvious way to do it.`

src/check_logrotate_missing.py
```diff
@@ -0,0 +1,177 @@
...
+        code = 1
+        for logfile in logrotate_missing:
+            print(logfile)
+    else:
+        print('OK')
+
+    exit(code)
+
+
+def get_parser():
```
hasegeli: Most of our scripts have `parse_args()` instead which returns the arguments.  I find it better.

lacunoc: changed

src/check_logrotate_missing.py
lacunoc: changed

PR #93, opened 2018-05-29
mysql: Add nagios check for buffer_pool_size

lacunoc:
We need a nagios check, which is comparing the
configured innodb_buffer_pool_size with the data_size
of the database at runtime.

### DIFF
--- a/src/check_mysql_buffer_pool.py
+++ b/src/check_mysql_buffer_pool.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python
+#
+# InnoGames Monitoring Plugins - check_mysql_buffer_pool.py
+#
+# Check the size of the configured mysql innodb_buffer_pool_size
+#
+# This script gets the configured innodb_buffer_pool_size at runtime and
+# compares this value with the data_size of the database.  It's also possible
+# to multiply the buffer_pool_size with a factor, to keep track on growing
+# databases, which are too big to fit into the available memory.
+# It raises a warning if the data_size of the database (this includes indexes)
+# is greater than the configured buffer_pool.
+# It raises a unknown if the connect/query could not be executed.
+#
+# Copyright (c) 2018, InnoGames GmbH
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


src/check_mysql_buffer_pool.py
```diff
@@ -0,0 +1,164 @@
...
+    from mysql.connector import (
+        connect,
+        DatabaseError,
+    )
+except ImportError:
+    from MySQLdb import (
+        connect,
+        OperationalError as DatabaseError,
+    )
+
```
elmatze79: I would remove the mysqldb support. mysql.connector is available for python2 and python3, so that I don't think we need anything else

src/check_mysql_buffer_pool.py
```diff
@@ -0,0 +1,164 @@
...
+    parser.add_argument('--host', default='localhost')
+    parser.add_argument('--password', default='password')
+    parser.add_argument(
+        '--unix-socket',
+        default='/var/run/mysqld/mysqld.sock',
+    )
+    parser.add_argument(
+        '--buffer_pool_factor',
+        default=1,
+        type=int
```
elmatze79: should be float when used as percentage calculation

src/check_mysql_buffer_pool.py
lacunoc: good catch, I will remove try/except then

src/check_mysql_buffer_pool.py
```diff
@@ -0,0 +1,164 @@
...
+        'SELECT (@@innodb_buffer_pool_size / POWER(1024,2))',
+    )
+
+    data_size = query_database(
+        user,
+        password,
+        host,
+        unix_socket,
+        'SELECT FLOOR(SUM(DATA_LENGTH+INDEX_LENGTH)/POWER(1024,2)) '
+        'FROM information_schema.TABLES;'
```
elmatze79: Only as a thought, and most probably won't apply to us:
You are counting all sizes regardless of engine type. This is OK as long InnoDB only databases are used. But when (external) people use this check with different engines, the check would return wrong values.
So adding a `WHERE ENGINE='InnoDB'` wouldn't hurt.

src/check_mysql_buffer_pool.py
lacunoc: yep, possible. changed

src/check_mysql_buffer_pool.py
lacunoc: since we are comparing this to innodb only, there where clause should be mandatory. thank you and changed.

src/check_mysql_buffer_pool.py
```diff
@@ -0,0 +1,158 @@
...
+    try:
+        buffer_pool_size, data_size = get_buffer_and_data_size(
+            args.user, args.password, args.host, args.unix_socket)
+    except DatabaseError as error:
+        print('UNKNOWN: ' + str(error))
+        exit(3)
+
+    if (buffer_pool_size * args.buffer_pool_factor) < data_size:
+        print('WARNING: buffer_pool_size ({}MB) < data_size ({}MB)'.format(
+            buffer_pool_size, data_size))
```
elmatze79: When using a factor != 1 the output might be misleading. I'm not sure how to phrase this right.

PR #95, opened 2018-06-01
swap: Introduce proper nagios check

lacunoc:
We need a proper check for swapping, which got
reasonable thresholds and is complaining, if the
swap is nearly full.


lacunoc:
The previous check was not reaching hard state in any case. Furthermore can you please explain how the `check_disk` check is also checking the swap partition?

lacunoc:
We will go with default `check_swap`, which is already available with the same thresholds (80%/90%) and will not touch the `check_swapping` check now.

PR #97, opened 2018-06-07
Add check_zookeeper.py

seqizz:
Copied from old plugins pack and ported to python3.

### DIFF
--- a/src/check_zookeeper.py
+++ b/src/check_zookeeper.py
@@ -1,4 +1,4 @@
-<F22>#! /usr/bin/env python
+#! /usr/bin/env python3
 #  Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
@@ -16,10 +16,12 @@
 # limitations under the License.
 """ Check Zookeeper Cluster
 
-Generic monitoring script that could be used with multiple platforms (Ganglia, Nagios, Cacti).
+Generic monitoring script that could be used with
+multiple platforms (Ganglia, Nagios, Cacti).
 
-It requires ZooKeeper 3.4.0 or greater. The script needs the 'mntr' 4letter word
-command (patch ZOOKEEPER-744) that was now commited to the trunk.
+It requires ZooKeeper 3.4.0 or greater.
+The script needs the 'mntr' 4letter word command (patch ZOOKEEPER-744)
+that was now commited to the trunk.
 The script also works with ZooKeeper 3.3.x but in a limited way.
 """
 
@@ -29,13 +31,14 @@ import logging
 import re
 import subprocess
 
-from StringIO import StringIO
+from io import StringIO
 from optparse import OptionParser, OptionGroup
 
 __version__ = (0, 1, 0)
 
-log = logging.getLogger()
 logging.basicConfig(level=logging.ERROR)
+log = logging.getLogger()
+
 
 class NagiosHandler(object):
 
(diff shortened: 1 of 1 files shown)


src/check_zookeeper.py
```diff
@@ -0,0 +1,407 @@
...
+        )
+
+        parser.add_option_group(group)
+
+    def call(self, *args, **kwargs):
+        subprocess.call(*args, **kwargs)
+
+    def analyze(self, opts, cluster_stats):
+        if len(cluster_stats) != 1:
+            print >>sys.stderr, 'Only allowed to monitor a single node.'
```
HerrSpace: I'll just stop marking these now.

src/check_zookeeper.py
```diff
@@ -0,0 +1,407 @@
...
+        s.close()
+
+        return data
+
+    def _parse(self, data):
+        """ Parse the output from the 'mntr' 4letter word command """
+        h = StringIO(data.decode())
+
+        result = {}
+        for line in h.readlines():
```
HerrSpace: Not sure about StringIO but can't you just as well use data.decode() directly here?

src/check_zookeeper.py
```diff
@@ -0,0 +1,407 @@
...
+                key, value = self._parse_line(line)
+                result[key] = value
+            except ValueError:
+                pass  # ignore broken lines
+
+        return result
+
+    def _parse_stat(self, data):
+        """ Parse the output from the 'stat' 4letter word command """
+        h = StringIO(data)
```
HerrSpace: Do you not need to decode it here?

src/check_zookeeper.py
```diff
@@ -0,0 +1,407 @@
...
+        if version:
+            result['zk_version'] = version[version.index(':')+1:].strip()
+
+        # skip all lines until we find the empty one
+        while h.readline().strip():
+            pass
+
+        for line in h.readlines():
+            m = re.match('Latency min/avg/max: (\d+)/(\d+)/(\d+)', line)
+            if m is not None:
```
HerrSpace: `if m:` would work just as well, I think

src/check_zookeeper.py
```diff
@@ -0,0 +1,407 @@
...
+        log.error('undefined handler: %s' % opts.output)
+        sys.exit(1)
+
+    return handler.analyze(opts, cluster_stats)
+
+
+def create_handler(name):
+    """ Return an instance of a platform specific analyzer """
+    try:
+        return globals()['%sHandler' % name.capitalize()]()
```
HerrSpace: This is just weird.

src/check_zookeeper.py
```diff
@@ -0,0 +1,407 @@
...
+    if opts.servers is None:
+        parser.error('The list of servers is mandatory')
+
+    opts.servers = [s.split(':') for s in opts.servers.split(',')]
+
+    return (opts, args)
+
+
+if __name__ == '__main__':
+    sys.exit(main())
```
HerrSpace: Just call main() here and sys.exit in there, not here.

src/check_zookeeper.py
```diff
@@ -0,0 +1,407 @@
...
+        return key, value
+
+
+def main():
+    opts, args = parse_cli()
+
+    cluster_stats = get_cluster_stats(opts.servers)
+    if opts.output is None:
+        dump_stats(cluster_stats)
+        return 0
```
HerrSpace: sys.exit(0) here

HerrSpace:
Are you sure all of this is working in python3?

PR #98, opened 2018-06-12
Redo check_expire.py

hasegeli:


src/check_ssl_cert_expire.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def get_extension_value(cert, short_name):
+    for ext_num in range(cert.get_extension_count()):
+        if cert.get_extension(ext_num).get_short_name() == short_name:
+            return str(cert.get_extension(ext_num))
+
+
+def decode_san(san_string):
+    if san_string:
```
TheFRedFox: I am not sure about yielding methods. But can you also do an early return check here?
``` python
if not san_string:
    return None  # or just return
```

src/check_ssl_cert_expire.py
```diff
@@ -0,0 +1,142 @@
...
+from datetime import datetime, timedelta
+from dateutil.tz import tzutc
+from dateutil.parser import parse
+from sys import exit
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument(
+        'pemfile',
```
TheFRedFox: Do you already know about `argparse.FileType`? https://docs.python.org/2/library/argparse.html#filetype-objects

src/check_ssl_cert_expire.py
```diff
@@ -0,0 +1,142 @@
...
+        print('OK: ' + output)
+    elif exit_code == 1:
+        print('WARNING: ' + output)
+    elif exit_code == 2:
+        print('CRITICAL: ' + output)
+    exit(exit_code)
+
+
+def interval(arg):
+    index = next(
```
TheFRedFox: Didn't really read it completely, but it somehow looks overcomplicated. I am wondering if there is no other way to get the different values.

src/check_ssl_cert_expire.py
hasegeli: Done

src/check_ssl_cert_expire.py
hasegeli: Done.  Good to learn.

src/check_ssl_cert_expire.py
hasegeli: Maybe it is good a have a little tricky code to not bore the reader.

TheFRedFox: 😆 Sometimes a good thing, yeah. But good then to have a TLDR comment above then. \^\^

src/check_ssl_cert_expire.py
TheFRedFox: 👍

PR #99, opened 2018-06-13
 check_dns_sshfp.py: Expect empty answer

hasegeli:


src/check_dns_sshfp.py
```diff
@@ -36,9 +37,10 @@
 
 def main():
     fqdn = getfqdn()
-    result = dns_query(fqdn, SSHFP, raise_on_no_answer=False)
 
-    if not result.rrset:
+    try:
+        result = dns_query(fqdn + '.', SSHFP)
```
TheFRedFox: What's about that extra dot?

src/check_dns_sshfp.py
```diff
@@ -36,9 +37,10 @@
 
 def main():
     fqdn = getfqdn()
-    result = dns_query(fqdn, SSHFP, raise_on_no_answer=False)
 
-    if not result.rrset:
```
TheFRedFox: No need to check `result.rrset` additionally?

src/check_dns_sshfp.py
hasegeli: I added a comment about it.

src/check_dns_sshfp.py
hasegeli: No, we don't need because it would raise an exception without `raise_on_no_answer`.

TheFRedFox: Alright. 👍

PR #100, opened 2018-06-14
Tech 19984 systemd timers

Felixoid:
Implemented check for systemd.timer

### DIFF
--- a/src/check_systemd_timers.py
+++ b/src/check_systemd_timers.py
@@ -37,7 +37,7 @@ class CheckTimers(object):
         CRITICAL=2,
         UNKNOWN=3,
     )
-    REV_CODES = {v: k for k, v in CODES.items()}
+    REV_CODES = dict(map(reversed, CODES.items()))
 
     def __init__(self, check_all, crit_threshold, ignored_timers, timer_units,
                  warn_threshold, **kwargs):
@@ -89,18 +89,14 @@ class CheckTimers(object):
         m = 1000000
         properties = self.timers[t_unit].properties
         if properties.LoadState != 'loaded':
-            setattr(
-                properties,
-                'MonitoringState',
-                (self.CODES['CRITICAL'], 'Timer is not loaded')
+            properties.MonitoringState = (
+                self.CODES['CRITICAL'], 'Timer is not loaded'
             )
             return
 
         if properties.ActiveState != 'active':
-            setattr(
-                properties,
-                'MonitoringState',
-                (self.CODES['CRITICAL'], 'Timer is not active')
+            properties.MonitoringState = (
+                self.CODES['CRITICAL'], 'Timer is not active'
             )
             return
 
@@ -117,24 +113,16 @@ class CheckTimers(object):
 
             if (self._warn_threshold <= since_last_execute/min_interval
                     < self._crit_threshold):
-                setattr(
-                    properties,
-                    'MonitoringState',
-                    (
-                        self.CODES['WARNING'],
-                        "Timer wasn't launch since {}".format(last_execute)
-                    )
+                properties.MonitoringState = (
+                    self.CODES['WARNING'],
+                    "Timer wasn't launch since {}".format(last_execute)
                 )
                 return
             elif self._crit_threshold <= since_last_execute/min_interval:
-                setattr(
-                    properties,
-                    'MonitoringState',
-                    (
-                        self.CODES['CRITICAL'],
-                        "Timer wasn't launch since {}, look at {}"
-                        .format(last_execute, properties.Unit)
-                    )
+                properties.MonitoringState = (
+                    self.CODES['CRITICAL'],
+                    "Timer wasn't launch since {}, look at {}"
+                    .format(last_execute, properties.Unit)
                 )
                 return
 
@@ -144,50 +132,43 @@ class CheckTimers(object):
         s_properties = self.services[t_unit].properties
         t_properties = self.timers[t_unit].properties
         if s_properties.LoadState != 'loaded':
-            setattr(
-                t_properties,
-                'MonitoringState',
-                (self.CODES['CRITICAL'], 'Related service is not loaded')
+            t_properties.MonitoringState = (
+                self.CODES['CRITICAL'], 'Related service is not loaded'
             )
             return
 
         if s_properties.ActiveState == 'failed':
-            setattr(
-                t_properties,
-                'MonitoringState',
-                (self.CODES['CRITICAL'], 'Related service is failed')
+            t_properties.MonitoringState = (
+                self.CODES['CRITICAL'], 'Related service is failed'
             )
             return
 
         if (s_properties.ActiveState == 'active'
                 and s_properties.SubState == 'exited'):
-            setattr(
-                t_properties,
-                'MonitoringState',
-                (
-                    self.CODES['CRITICAL'],
-                    "Related service is misconfigured, "
-                    "remove 'RemainAfterExit'"
-                )
+            t_properties.MonitoringState = (
+                self.CODES['CRITICAL'],
+                "Related service is misconfigured, "
+                "remove 'RemainAfterExit'"
             )
             return
 
         setattr(t_properties, 'MonitoringState', (self.CODES['OK'], ''))
 
     def get_nagios(self):
-        statuses = {name: self.timers[name].properties.MonitoringState
-                    for name in self.timers}
+        statuses = {name: obj.properties.MonitoringState
+                    for name, obj in self.timers.items()}
         logger.debug('Statuses in get_nagios: {}'.format(statuses))
-        exit_code = max([statuses[n][0] for n in statuses])
+        exit_code = max([n[0] for n in statuses.values()])
         logger.debug("Exit_code = {}".format(exit_code))
         if exit_code == 0:
             message = 'OK'
             return exit_code, message
         msg_format = '{}: {}'
         messages = [
-            msg_format.format(t, statuses[t][1])
-            for t in statuses if statuses[t][0] > 0
+            msg_format.format(name, status[1])
+            for name, status in statuses.items() if status[0] > 0
         ]
+        # Add "STATUS:" to the first message
         messages[0] = msg_format.format(self.REV_CODES[exit_code], messages[0])
         return exit_code, ";".join(messages)
 
--- a/src/check_systemd_timers.py
+++ b/src/check_systemd_timers.py
@@ -14,10 +14,11 @@
 
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
 from datetime import datetime
-# python2/3 compatibility
-from time import time
+from enum import Enum
 from systemd_dbus.manager import Manager
 from systemd_dbus.exceptions import SystemdError
+# python2/3 compatibility
+from time import time
 import sys
 import logging
 
@@ -27,52 +28,49 @@ logging.basicConfig(
 logger = logging.getLogger(__name__)
 
 
+class Codes(Enum):
+    OK = 0
+    WARNING = 1
+    CRITICAL = 2
+    UNKNOWN = 3
+
+
 class CheckTimers(object):
     """
     Check timers for status, last executed service time, service status
     """
-    CODES = dict(
-        OK=0,
-        WARNING=1,
-        CRITICAL=2,
-        UNKNOWN=3,
-    )
-    REV_CODES = dict(map(reversed, CODES.items()))
 
     def __init__(self, check_all, crit_threshold, ignored_timers, timer_units,
-                 warn_threshold, **kwargs):
+                 warn_threshold, **_):
         self._crit_threshold = crit_threshold
         self._warn_threshold = warn_threshold
-        try:
-            s_manager = Manager()
-            timers = s_manager.list_timers()
-            logger.debug(
-                'Timers in system are: {}'
-                .format([t.properties.Id for t in timers])
-            )
-            if check_all:
-                self.timers = {
-                    str(t.properties.Id): t for t in timers
-                    if str(t.properties.Id) not in ignored_timers
-                }
-            else:
-                self.timers = {
-                    str(t.properties.Id): t for t in timers
-                    if str(t.properties.Id) in timer_units
-                }
-            logger.debug('Timers after filtering are: {}'.format(self.timers))
-            self.services = {
-                str(t.properties.Id): s_manager.get_unit(t.properties.Unit)
-                for t in timers
+        s_manager = Manager()
+        timers = s_manager.list_timers()
+        logger.debug(
+            'Timers in system are: {}'
+            .format([t.properties.Id for t in timers])
+        )
+
+        if check_all:
+            self.timers = {
+                str(t.properties.Id): t for t in timers
+                if str(t.properties.Id) not in ignored_timers
             }
-        except SystemdError:
-            print('UNKNOWN: Error while receive info for systemd from dbus')
-            sys.exit(self.CODES['UNKNOWN'])
-
-        if self.timers == {}:
-            print('UNKNOWN: Timer units not found')
-            sys.exit(self.CODES['UNKNOWN'])
-
+        else:
+            self.timers = {
+                str(t.properties.Id): t for t in timers
+                if str(t.properties.Id) in timer_units
+            }
+        logger.debug('Timers after filtering are: {}'.format(self.timers))
+        self.services = {
+            str(t.properties.Id): s_manager.get_unit(t.properties.Unit)
+            for t in timers
+        }
+        logger.debug(
+            'Linked to timers services are: {}'
+            .format([(s, obj.properties.Id)
+                     for s, obj in self.services.items()])
+        )
         self.now = time()
 
     def check_timers(self):
@@ -90,37 +88,38 @@ class CheckTimers(object):
         properties = self.timers[t_unit].properties
         if properties.LoadState != 'loaded':
             properties.MonitoringState = (
-                self.CODES['CRITICAL'], 'Timer is not loaded'
+                Codes.CRITICAL.value, 'Timer is not loaded'
             )
             return
 
         if properties.ActiveState != 'active':
             properties.MonitoringState = (
-                self.CODES['CRITICAL'], 'Timer is not active'
+                Codes.CRITICAL.value, 'Timer is not active'
             )
             return
 
         intervals = properties.TimersMonotonic
         if intervals:
-            min_interval = min([
-                p[1]/m for p in intervals
+            min_interval = min(
+                p[1] / m for p in intervals
                 if p[0] in checked_intervals
-            ])
+            )
             since_last_execute = self.now - properties.StateChangeTimestamp/m
             last_execute = datetime.fromtimestamp(
-                properties.StateChangeTimestamp/m
+                properties.StateChangeTimestamp / m
             )
 
             if (self._warn_threshold <= since_last_execute/min_interval
                     < self._crit_threshold):
                 properties.MonitoringState = (
-                    self.CODES['WARNING'],
+                    Codes.WARNING.value,
                     "Timer wasn't launch since {}".format(last_execute)
                 )
                 return
-            elif self._crit_threshold <= since_last_execute/min_interval:
+
+            if self._crit_threshold <= since_last_execute/min_interval:
                 properties.MonitoringState = (
-                    self.CODES['CRITICAL'],
+                    Codes.CRITICAL.value,
                     "Timer wasn't launch since {}, look at {}"
                     .format(last_execute, properties.Unit)
                 )
@@ -133,44 +132,44 @@ class CheckTimers(object):
         t_properties = self.timers[t_unit].properties
         if s_properties.LoadState != 'loaded':
             t_properties.MonitoringState = (
-                self.CODES['CRITICAL'], 'Related service is not loaded'
+                Codes.CRITICAL.value, 'Related service is not loaded'
             )
             return
 
         if s_properties.ActiveState == 'failed':
             t_properties.MonitoringState = (
-                self.CODES['CRITICAL'], 'Related service is failed'
+                Codes.CRITICAL.value, 'Related service is failed'
             )
             return
 
         if (s_properties.ActiveState == 'active'
                 and s_properties.SubState == 'exited'):
             t_properties.MonitoringState = (
-                self.CODES['CRITICAL'],
-                "Related service is misconfigured, "
-                "remove 'RemainAfterExit'"
+                Codes.CRITICAL.value,
+                'Related service is misconfigured, '
+                'try to remove "RemainAfterExit"'
             )
             return
 
-        setattr(t_properties, 'MonitoringState', (self.CODES['OK'], ''))
+        t_properties.MonitoringState = (Codes.OK.value, '')
 
-    def get_nagios(self):
-        statuses = {name: obj.properties.MonitoringState
-                    for name, obj in self.timers.items()}
-        logger.debug('Statuses in get_nagios: {}'.format(statuses))
-        exit_code = max([n[0] for n in statuses.values()])
-        logger.debug("Exit_code = {}".format(exit_code))
+    def get_results(self):
+        statuses = [(name, obj.properties.MonitoringState)
+                    for name, obj in self.timers.items()]
+        logger.debug('Statuses in get_results: {}'.format(statuses))
+        exit_code = max(int(n[1][0]) for n in statuses)
+        logger.debug('Exit_code = {}'.format(exit_code))
         if exit_code == 0:
             message = 'OK'
             return exit_code, message
         msg_format = '{}: {}'
         messages = [
             msg_format.format(name, status[1])
-            for name, status in statuses.items() if status[0] > 0
+            for name, status in statuses if status[0] > 0
         ]
         # Add "STATUS:" to the first message
-        messages[0] = msg_format.format(self.REV_CODES[exit_code], messages[0])
-        return exit_code, ";".join(messages)
+        messages[0] = msg_format.format(Codes(exit_code), messages[0])
+        return exit_code, ';'.join(messages)
 
 
 def parse_args():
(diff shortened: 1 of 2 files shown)


src/check_systemd_units.py
```diff
@@ -146,7 +146,7 @@ def get_exitcode(unit_name):
     command = 'systemctl show -pExecMainStatus {}'.format(unit_name)
     try:
         output = subprocess.check_output(command.split())
-    except subprocess.CalledProcessError as error:
+    except subprocess.CalledProcessError:
```
hasegeli: Why do you touch this?

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+
+    check = CheckTimers(**vars(args))
+    check.check_timers()
+    exit_code, message = check.get_nagios()
+    print(message)
+    sys_exit(exit_code)
+
+
+if __name__ == '__main__':
+    main()
```
hasegeli: Please make the code PEP8 compatible before submitting for review.  It is easier to read the code in familiar style.

src/check_systemd_units.py
Felixoid: because error not used in code?

src/check_systemd_timers.py
Felixoid: My flake8 is completely ok with it. What do you mean?

src/check_systemd_units.py
hasegeli: Yes, it is not used, so not needed.

src/check_systemd_timers.py
hasegeli:
```
check_systemd_timers.py:24:80: E501 line too long (83 > 79 characters)
check_systemd_timers.py:49:80: E501 line too long (94 > 79 characters)
check_systemd_timers.py:51:80: E501 line too long (117 > 79 characters)
check_systemd_timers.py:53:80: E501 line too long (110 > 79 characters)
check_systemd_timers.py:55:80: E501 line too long (105 > 79 characters)
check_systemd_timers.py:81:80: E501 line too long (99 > 79 characters)
check_systemd_timers.py:85:80: E501 line too long (99 > 79 characters)
check_systemd_timers.py:90:80: E501 line too long (88 > 79 characters)
check_systemd_timers.py:92:80: E501 line too long (84 > 79 characters)
check_systemd_timers.py:99:80: E501 line too long (96 > 79 characters)
check_systemd_timers.py:122:80: E501 line too long (111 > 79 characters)
check_systemd_timers.py:126:80: E501 line too long (107 > 79 characters)
check_systemd_timers.py:129:80: E501 line too long (86 > 79 characters)
check_systemd_timers.py:133:80: E501 line too long (102 > 79 characters)
check_systemd_timers.py:140:80: E501 line too long (95 > 79 characters)
check_systemd_timers.py:186:80: E501 line too long (80 > 79 characters)
check_systemd_timers.py:194:80: E501 line too long (81 > 79 characters)
```

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+        WARNING=1,
+        CRITICAL=2,
+        UNKNOWN=3,
+    )
+    REV_CODES = {v: k for k, v in CODES.items()}
+
+    def __init__(self, check_all,
+                 crit_threshold, ignored_timers,
+                 timer_units, warn_threshold,
+                 **kwargs):
```
TheFRedFox: `**kwargs` is not used.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+        self._crit_threshold = crit_threshold
+        self._warn_threshold = warn_threshold
+        try:
+            s_manager = Manager()
+            timers = s_manager.list_timers()
+            logger.debug('Timers in system are: {}'.format(
```
TheFRedFox: We mostly go for this solution:
``` python
logger.debug(
    'Timers ...'
    .format(...)
)
```

So having the braces as last resp only thing in the line.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+                [t.properties.Id for t in timers]))
+            if check_all:
+                self.timers = {str(t.properties.Id): t
```
TheFRedFox: Same here with braces.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+                               for t in timers
+                               if str(t.properties.Id) not in ignored_timers}
+            else:
+                self.timers = {str(t.properties.Id): t
```
TheFRedFox: Same here with braces.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+                               for t in timers
+                               if str(t.properties.Id) in timer_units}
+            logger.debug('Timers after filtering are: {}'.format(self.timers))
+            self.services = {
```
TheFRedFox: Same here with braces.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+                    )
+                )
+                return
+            elif self._crit_threshold <= since_last_execute/min_interval:
+                setattr(
+                    properties,
+                    'MonitoringState',
+                    (
+                        self.CODES['CRITICAL'],
+                        "Timer wasn't launch since {}, look at {}".format(
```
TheFRedFox: Go for:
``` python
"Timer wasn't ...".
format(...)
```

It is possible because the string is already in braces.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+            setattr(
+                properties,
+                'MonitoringState',
+                (self.CODES['CRITICAL'], 'Timer is not active')
+            )
+            return
+
+        intervals = properties.TimersMonotonic
+        if intervals:
+            min_interval = min([p[1]/M
```
TheFRedFox: Same here with braces.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+        ]
+        messages[0] = msg_format.format(self.REV_CODES[exit_code], messages[0])
+        return exit_code, ";".join(messages)
+
+
+def parse_args():
+    """Parse the arguments"""
+    parser = ArgumentParser(formatter_class=ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        '-a',
```
TheFRedFox: Please go for this to reduce lines:
``` python
parser.add_argument(
    '-a', action='store_true', dest='check_all', default=False,
    help='check ...'
)
```
for all arguments. You can keep the help line in a separate line to have a direct view to it.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+def main():
+    """The main program"""
+    args = parse_args()
+    logger.setLevel(args.log_level)
+    logger.debug('Initial arguments are: {}'.format(args))
+    if not args.check_all and not args.timer_units:
+        print('UNKNOWN: Either CHECK_ALL or TIMER_UNITS should be defined')
+        sys_exit(CheckTimers.UNKNOWN)
```
TheFRedFox: `CheckTimers.UNKNOWN` does not exists, but `CheckTimers.CODES.get('UNKNOWN')` does.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+#
+# * critical when a timer unit is inactive of service unit is stoped
+# * warning for long term unexecuted
+#
+# Copyright (c) 2018, InnoGames GmbH
+#
+
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
+from datetime import datetime
+# python2/3 compatibility
```
TheFRedFox: Do we still need python2 compatibility?

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+from time import time
+from systemd_dbus.manager import Manager
+from systemd_dbus.exceptions import SystemdError
+from sys import exit as sys_exit
```
TheFRedFox: Why? `sys.exit` or `sys_exit` isn't that a huge difference?

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+    """
+    CODES = dict(
+        OK=0,
+        WARNING=1,
+        CRITICAL=2,
+        UNKNOWN=3,
+    )
+    REV_CODES = {v: k for k, v in CODES.items()}
+
+    def __init__(self, check_all,
```
TheFRedFox: Why are the parameters split like that? Why not:

``` python
def __init__(self, check_all, crit_threshold, ignored_timers, timer_units,
             warn_threshold, **kwargs):
```
?

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+            return
+
+        if (s_properties.ActiveState == 'active'
+                and s_properties.SubState == 'exited'):
+            setattr(
+                t_properties,
+                'MonitoringState',
+                (
+                    self.CODES['CRITICAL'],
+                    ("Related service is misconfigured, "
```
TheFRedFox: I rather think that you mean that:
``` python
(
    self.CODES['CRITICAL'],
    "Related service is misconfigured, remove"
    "'RemainAfterExit'"
)
```

It is possible, because it is already in braces.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+            'OnUnitActiveUSec',
+            'OnActiveUSec',
+            'OnUnitInactiveUSec',
+        ]
+        # Microseconds to seconds
+        M = 1000000
+        properties = self.timers[t_unit].properties
+        if properties.LoadState != 'loaded':
+            print(properties.LoadState)
+            setattr(
```
TheFRedFox: You are using it a lot .. you should consider to separate that code into a separate method. But maybe with next change.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,262 @@
...
+            self._check_timer(t_unit)
+
+    def _check_timer(self, t_unit):
+        checked_intervals = [
+            'OnUnitActiveUSec',
+            'OnActiveUSec',
+            'OnUnitInactiveUSec',
+        ]
+        # Microseconds to seconds
+        M = 1000000
```
TheFRedFox: Variables are normally named lowercase_snake_casing

src/check_systemd_timers.py
Felixoid: Thank you, good catch

src/check_systemd_timers.py
TheFRedFox: @hasegeli Or what is your opinion about that? I think it is a lot lines for almost nothing.

src/check_systemd_timers.py
Felixoid: they are here to ignore of args.log_level

src/check_systemd_timers.py
Felixoid: well, what's the difference?

src/check_systemd_timers.py
Felixoid: thanks

src/check_systemd_timers.py
Felixoid: true

src/check_systemd_timers.py
Felixoid: makes sense

src/check_systemd_timers.py
Felixoid: done

src/check_systemd_timers.py
Felixoid: you are right, I misunderstood that import uses all object (which about `from blah import *`)

src/check_systemd_timers.py
Felixoid: Not we, maybe other users.

src/check_systemd_timers.py
hasegeli: I either put arguments in-line, or altogether to a line, or all to separate lines:
```python
f(a, b, c)

f(
    a, b, c
)

f(
    a,
    b,
    c,
)
```

It's no rule though.

src/check_systemd_timers.py
Felixoid: ok, the difference in lines amount

src/check_systemd_timers.py
```diff
@@ -0,0 +1,243 @@
...
+import sys
+import logging
+
+logging.basicConfig(
+    format='%(levelname)-8s [%(filename)s:%(lineno)d] %(message)s'
+)
+logger = logging.getLogger(__name__)
+
+
+class CheckTimers(object):
```
hasegeli: This shouldn't be a class.

src/check_systemd_timers.py
TheFRedFox: Question is rather, if we want to provide python2 compatibility support. I don't know what the status of this is.
Maybe @hasegeli can say more.

src/check_systemd_timers.py
TheFRedFox: I mean it is actually pythonic way to use `from import`s, however, at least for `sys` module it does not make any memory difference and just to change `sys.exit` to `sys_exit` I find better to just go for `import sys`.

Also, just for you to know, I also would not go the `from import` style if it makes the code misleading, like for `from re import compile` I think `compile` is too generic and misleading without the context of regex. So there I would also rather go for `re.compile` than just `compile`. 😉

src/check_systemd_timers.py
TheFRedFox: I actually like the possibility to not go for separated lines. But I always check what makes sense in regards of readability and maintainability and for the argument parser I guess it is fair enough to go with my style. ^^ Just my 2 cents.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,243 @@
...
+    """
+    Check timers for status, last executed service time, service status
+    """
+    CODES = dict(
+        OK=0,
+        WARNING=1,
+        CRITICAL=2,
+        UNKNOWN=3,
+    )
+    REV_CODES = {v: k for k, v in CODES.items()}
```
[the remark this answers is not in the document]
TheFRedFox: Nice thing, however, I found this quote:
> filter and map should die and be subsumed into list comprehensions, not grow more variants - Guido Van Rossum in PEP 279

https://stackoverflow.com/questions/483666/python-reverse-invert-a-mapping#comment33428395_22047530

src/check_systemd_timers.py
Felixoid: If you want to implement logic from here with using systemd_dbus, then you need install https://github.com/wiliamsouza/python-systemd/pull/10 on hosts as well.

Library already built from my branch and uploaded to innogames-stable-amd64 repo.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            else:
+                self.timers = {
+                    str(t.properties.Id): t for t in timers
+                    if str(t.properties.Id) in timer_units
+                }
+            logger.debug('Timers after filtering are: {}'.format(self.timers))
+            self.services = {
+                str(t.properties.Id): s_manager.get_unit(t.properties.Unit)
+                for t in timers
+            }
```
hasegeli: Too long `try:` block.  It is not clear what can raise the exception.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            if check_all:
+                self.timers = {
+                    str(t.properties.Id): t for t in timers
+                    if str(t.properties.Id) not in ignored_timers
+                }
+            else:
+                self.timers = {
+                    str(t.properties.Id): t for t in timers
+                    if str(t.properties.Id) in timer_units
+                }
```
hasegeli: No need to iterate the same list 2 times.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+    def __init__(self, check_all, crit_threshold, ignored_timers, timer_units,
+                 warn_threshold, **kwargs):
+        self._crit_threshold = crit_threshold
+        self._warn_threshold = warn_threshold
+        try:
+            s_manager = Manager()
+            timers = s_manager.list_timers()
+            logger.debug(
+                'Timers in system are: {}'
+                .format([t.properties.Id for t in timers])
```
hasegeli: No need for `[]`

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+    CODES = dict(
+        OK=0,
+        WARNING=1,
+        CRITICAL=2,
+        UNKNOWN=3,
+    )
+    REV_CODES = dict(map(reversed, CODES.items()))
+
+    def __init__(self, check_all, crit_threshold, ignored_timers, timer_units,
+                 warn_threshold, **kwargs):
```
hasegeli: `**kwargs` is still wrong.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    Check timers for status, last executed service time, service status
+    """
+    CODES = dict(
+        OK=0,
+        WARNING=1,
+        CRITICAL=2,
+        UNKNOWN=3,
+    )
+    REV_CODES = dict(map(reversed, CODES.items()))
```
hasegeli: This should better be an `object` or `Enum`.  `dict` is the wrong data structure.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            s_manager = Manager()
+            timers = s_manager.list_timers()
+            logger.debug(
+                'Timers in system are: {}'
+                .format([t.properties.Id for t in timers])
+            )
+            if check_all:
+                self.timers = {
+                    str(t.properties.Id): t for t in timers
+                    if str(t.properties.Id) not in ignored_timers
```
hasegeli: Calling `str()` for 2 times on the same thing.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            logger.debug('Timers after filtering are: {}'.format(self.timers))
+            self.services = {
+                str(t.properties.Id): s_manager.get_unit(t.properties.Unit)
+                for t in timers
+            }
+        except SystemdError:
+            print('UNKNOWN: Error while receive info for systemd from dbus')
+            sys.exit(self.CODES['UNKNOWN'])
+
+        if self.timers == {}:
```
hasegeli: Not Pythonic way of testing

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            print('UNKNOWN: Timer units not found')
+            sys.exit(self.CODES['UNKNOWN'])
```
hasegeli: We shouldn't `print()` or `exit()` from the random places of the code.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            min_interval = min([
+                p[1]/m for p in intervals
+                if p[0] in checked_intervals
+            ])
+            since_last_execute = self.now - properties.StateChangeTimestamp/m
+            last_execute = datetime.fromtimestamp(
+                properties.StateChangeTimestamp/m
+            )
+
+            if (self._warn_threshold <= since_last_execute/min_interval
```
hasegeli: No space around operator

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            t_properties.MonitoringState = (
+                self.CODES['CRITICAL'],
+                "Related service is misconfigured, "
+                "remove 'RemainAfterExit'"
+            )
+            return
+
+        setattr(t_properties, 'MonitoringState', (self.CODES['OK'], ''))
+
+    def get_nagios(self):
```
hasegeli: You get the Nagios???

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+        statuses = {name: obj.properties.MonitoringState
+                    for name, obj in self.timers.items()}
+        logger.debug('Statuses in get_nagios: {}'.format(statuses))
+        exit_code = max([n[0] for n in statuses.values()])
+        logger.debug("Exit_code = {}".format(exit_code))
```
hasegeli: Please do not mix single and double quotes randomly.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            )
+            return
+
+        setattr(t_properties, 'MonitoringState', (self.CODES['OK'], ''))
+
+    def get_nagios(self):
+        statuses = {name: obj.properties.MonitoringState
+                    for name, obj in self.timers.items()}
+        logger.debug('Statuses in get_nagios: {}'.format(statuses))
+        exit_code = max([n[0] for n in statuses.values()])
```
hasegeli: No need for the `[]`

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+                "Related service is misconfigured, "
+                "remove 'RemainAfterExit'"
+            )
+            return
+
+        setattr(t_properties, 'MonitoringState', (self.CODES['OK'], ''))
+
+    def get_nagios(self):
+        statuses = {name: obj.properties.MonitoringState
+                    for name, obj in self.timers.items()}
```
hasegeli: `dict` is again wrong in here.  `dict` is the data structure you get to use when you need to random access the elements.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+        if (s_properties.ActiveState == 'active'
+                and s_properties.SubState == 'exited'):
+            t_properties.MonitoringState = (
+                self.CODES['CRITICAL'],
+                "Related service is misconfigured, "
+                "remove 'RemainAfterExit'"
+            )
+            return
+
+        setattr(t_properties, 'MonitoringState', (self.CODES['OK'], ''))
```
hasegeli: Don't use `setattr()`.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+                    if str(t.properties.Id) in timer_units
+                }
+            logger.debug('Timers after filtering are: {}'.format(self.timers))
+            self.services = {
+                str(t.properties.Id): s_manager.get_unit(t.properties.Unit)
+                for t in timers
+            }
+        except SystemdError:
+            print('UNKNOWN: Error while receive info for systemd from dbus')
+            sys.exit(self.CODES['UNKNOWN'])
```
hasegeli: We shouldn't `print()` or `exit()` from the random places of the code.

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+        if properties.ActiveState != 'active':
+            properties.MonitoringState = (
+                self.CODES['CRITICAL'], 'Timer is not active'
+            )
+            return
+
+        intervals = properties.TimersMonotonic
+        if intervals:
+            min_interval = min([
+                p[1]/m for p in intervals
```
hasegeli: No space around the operator

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+
+        if properties.ActiveState != 'active':
+            properties.MonitoringState = (
+                self.CODES['CRITICAL'], 'Timer is not active'
+            )
+            return
+
+        intervals = properties.TimersMonotonic
+        if intervals:
+            min_interval = min([
```
hasegeli: Unnecessary `[]`

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+        # Microseconds to seconds
+        m = 1000000
+        properties = self.timers[t_unit].properties
+        if properties.LoadState != 'loaded':
+            properties.MonitoringState = (
+                self.CODES['CRITICAL'], 'Timer is not loaded'
+            )
+            return
+
+        if properties.ActiveState != 'active':
```
hasegeli: Properties are not named `CamelCase` style

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+            )
+
+            if (self._warn_threshold <= since_last_execute/min_interval
+                    < self._crit_threshold):
+                properties.MonitoringState = (
+                    self.CODES['WARNING'],
+                    "Timer wasn't launch since {}".format(last_execute)
+                )
+                return
+            elif self._crit_threshold <= since_last_execute/min_interval:
```
hasegeli: Unnecessary `elif`

src/check_systemd_timers.py
```diff
@@ -0,0 +1,224 @@
...
+        logger.debug('Statuses in get_nagios: {}'.format(statuses))
+        exit_code = max([n[0] for n in statuses.values()])
+        logger.debug("Exit_code = {}".format(exit_code))
+        if exit_code == 0:
+            message = 'OK'
+            return exit_code, message
+        msg_format = '{}: {}'
+        messages = [
+            msg_format.format(name, status[1])
+            for name, status in statuses.items() if status[0] > 0
```
hasegeli: Can it be less than `0`?

src/check_systemd_timers.py
Felixoid: replaced, done

src/check_systemd_timers.py
Felixoid: done

src/check_systemd_timers.py
Felixoid: done

src/check_systemd_timers.py
Felixoid: fixed

src/check_systemd_timers.py
Felixoid: Yes, (str, (int, str)) looks, at least, not worse

src/check_systemd_timers.py
Felixoid: fixed

src/check_systemd_timers.py
Felixoid: Check for the rest file, done

hasegeli:
I suggest #101 instead.

PR #113, opened 2018-08-06
Add missing check_ prefixes

hasegeli:


TheFRedFox:
Need to check it with Felix. Will ask him if he can do it before his Sabbatical.

PR #121, opened 2018-08-31
Rewrite check_puppetd to python

axxelG:


axxelG:
Not tested on BSD yet because we don't have python3 there. @mtkh will install it.

igmonplugins/check_puppetd.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python3
```
sponkydive: we should not force python3 and keep the check python2 compatible.
please use only python instead of python3

sponkydive:
i also think we should rename the check to something like check_puppet_interval or puppet_lastrun.
The check does not check the daemon and as we have to change check commands etc anyway, it is a good time to do this.

PR #123, opened 2018-09-04
check_puppetd: Increase the critical threshold

seqizz:


igmonplugins/check_puppetd.sh
```diff
@@ -22,7 +22,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-diff_crit=8000
+diff_crit=15000
```
Felixoid: What about making an argument with default from it?

sponkydive: there is already a python version with optional arguments for thresholds which needs minor changes until pulled into master.
Until that we can go with this i would say

PR #126, opened 2018-10-09
check_lbpools: Disregard down nodes

seqizz:


seqizz:
Re-ping for evaluation.

src/check_lbpools.py
```diff
@@ -127,7 +127,10 @@ def compare_pools(pools, testtool, send):
                 num_nodes = testtool[testtool_name]['nodes_alive']
                 in_testtool = True
 
-            optimal_nodes = len(pool_v['nodes'])
+            optimal_nodes = 0
+            for item in pool_v['nodes']:
```
TuxPowered42: Those are keys, not items. You can itearte over items too and then you don't need to address the dictionary in the next line.

src/check_lbpools.py
```diff
@@ -127,7 +127,10 @@ def compare_pools(pools, testtool, send):
                 num_nodes = testtool[testtool_name]['nodes_alive']
                 in_testtool = True
 
-            optimal_nodes = len(pool_v['nodes'])
+            optimal_nodes = 0
+            for value in pool_v['nodes'].values():
+                if not value['downtime']:
+                    optimal_nodes += 1
```
TheFRedFox: How about this?:
``` python
optimal_nodes = sum(not v['downtime'] for v in pool_v['nodes'].values())
```

seqizz: I forgot about this issue since noone used downtimed ready-to-go nodes in production apparently.
Good suggestion btw. I'll re-consider this MR and change if still needed :+1:

seqizz:
Narrator: _But it was not needed..._

PR #130, opened 2018-10-15
Revert "Rename src/ directory"

TuxPowered42:
This reverts commit 1ecc7be8982a0c8eec829fa2105e48af2a170608.

BSD version of install can only instal files, not dirs.


TheFRedFox:
I removed all duplicated files which were created due to renames in between.

TheFRedFox:
Just a type in the commit message: `instal files` should be `install files`

PR #133, opened 2019-01-04
Reimplement dbus and timers

Felixoid:
This PR is based on both #100 and #101 and implements everything from them in one check.

It shouldn't be merged until company wide dbus tuning `max_match_rules_per_connection=2048` for system bus (`/etc/dbus-1/system.d/max_match_rules_per_connection.conf`)

### DIFF
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -138,14 +138,20 @@ class SystemdUnit:
             # See the man 5 systemd.service for ExecMainStatus and
             # SuccessExitStatus
             # All currently running services have ExecMainStatus=0
-            if (self.type_properties.ExecMainStatus not in
-                    self.type_properties.SuccessExitStatus[0] + [0]):
+            if (
+                # Old versions of systemd don't have ExecMainStatus
+                (hasattr(self.type_properties, 'SuccessExitStatus') and
+                 self.type_properties.ExecMainStatus not in
+                 self.type_properties.SuccessExitStatus[0])
+                or self.type_properties.ExecMainStatus == 0
+            ):
                 return (
                     self._warn_level,
                     'the service exited with {} code'.format(
                         self.type_properties.ExecMainStatus
                     )
                 )
+                print()
             if (
                 self.unit_properties.ActiveState == 'active'
                 and self.unit_properties.SubState == 'exited'
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -208,21 +208,21 @@ class SystemdUnit:
             # We could check only monotonic triggers for regular execution
             min_interval = min(p[1] for p in intervals
                                if p[0] in checked_intervals) / m
-            since_last_execute = (
+            inactivity = (
                 now - self.type_properties.LastTriggerUSec / m
             )
             last_execute = datetime.fromtimestamp(
                 self.type_properties.LastTriggerUSec / m
             )
             logger.info(
-                '{}: min_interval={}, since_last_execute={}, last_execute={}, '
+                '{}: min_interval={}, inactivity={}, last_execute={}, '
                 'since_last_execute / min_interval={}'
                 .format(
-                    str(self), min_interval, since_last_execute, last_execute,
-                    since_last_execute / min_interval
+                    str(self), min_interval, inactivity, last_execute,
+                    inactivity / min_interval
                 )
             )
-            if (timer_warn <= since_last_execute / min_interval < timer_crit):
+            if (timer_warn <= inactivity / min_interval < timer_crit):
                 return (
                     self._warn_level,
                     'the timer hasn\'t been launched since {}, look at {}'
@@ -230,7 +230,7 @@ class SystemdUnit:
                         last_execute, self.type_properties.Unit
                     )
                 )
-            if (timer_crit <= since_last_execute / min_interval):
+            if (timer_crit <= inactivity / min_interval):
                 return (
                     self._crit_level,
                     'the timer hasn\'t been launched since {}, look at {}'
@@ -287,7 +287,7 @@ def parse_args():
         dest='timer_warn',
         default=3,
         type=float,
-        help='warning threshold of timer (inactivity/max_monotonic_interval)',
+        help='warning threshold of timer (inactivity/min_monotonic_interval)',
     )
     parser.add_argument(
         '-c',
@@ -295,7 +295,7 @@ def parse_args():
         dest='timer_crit',
         default=7,
         type=float,
-        help='critical threshold of timer (inactivity/max_monotonic_interval)',
+        help='critical threshold of timer (inactivity/min_monotonic_interval)',
     )
     parser.add_argument(
         '-l',
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -16,7 +16,7 @@
 #
 
 from argparse import ArgumentParser
-from enum import Enum
+from enum import IntEnum
 from sys import exit
 
 from datetime import datetime
@@ -38,7 +38,7 @@ now = time()
 systemd_manager = Manager()
 
 
-class Codes(int, Enum):
+class Codes(IntEnum):
     OK = 0
     WARNING = 1
     CRITICAL = 2
@@ -228,14 +228,16 @@ class SystemdUnit:
             if (timer_warn <= since_last_execute / min_interval < timer_crit):
                 return (
                     self._warn_level,
-                    'the timer wasn\'t launch since {}, look at {}'.format(
+                    'the timer hasn\'t been launched since {}, look at {}'
+                    .format(
                         last_execute, self.type_properties.Unit
                     )
                 )
             if (timer_crit <= since_last_execute / min_interval):
                 return (
                     self._crit_level,
-                    'the timer wasn\'t launch since {}, look at {}'.format(
+                    'the timer hasn\'t been launched since {}, look at {}'
+                    .format(
                         last_execute, self.type_properties.Unit
                     )
                 )
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -131,7 +131,8 @@ class SystemdUnit:
                 self.unit_properties.ActiveState,
             )
         )
-        if critical:
+        self.__critical = critical
+        if self.__critical:
             self._crit_level = Codes.CRITICAL
             self._warn_level = Codes.WARNING
         else:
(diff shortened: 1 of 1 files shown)


Felixoid:
@HerrSpace and @TheFRedFox, could you review it, please?

TheFRedFox:
I'm sry, but I won't have much time left the next weeks. But a tip, separate the PR into more smaller commits to make it easier to review the changes. If it is easier for us to review you will get your review sooner. 😉

Felixoid:
> I'm sry, but I won't have much time left the next weeks. But a tip, separate the PR into more smaller commits to make it easier to review the changes. If it is easier for us to review you will get your review sooner. wink

Thanks, Freddy. Unfortunately, it's nothing to split here. Logic was totally changed.

src/check_systemd_units.py
```diff
@@ -1,53 +1,250 @@
...
+                )
+        return (Codes.OK, '')
+
+    def _check_timer(self, timer_warn, timer_crit):
+        '''
+        Detects problems for a timer unit
+        '''
+        checked_intervals = [
+            'OnUnitActiveUSec',
+            'OnActiveUSec',
```
HerrSpace: OnUnitActiveUSec = Last time the unit was started
OnActiveUSec = Last time the timer was triggered

Is that correct?

src/check_systemd_units.py
```diff
@@ -1,53 +1,250 @@
...
+            min_interval = min(
+                p[1] / m for p in intervals
+                if p[0] in checked_intervals
+            )
+            since_last_execute = (
+                now - self.type_properties.LastTriggerUSec / m
+            )
+            last_execute = datetime.fromtimestamp(
+                self.type_properties.LastTriggerUSec / m
+            )
```
HerrSpace: You could simplify this to
```python3
            last_execute = datetime.fromtimestamp(
                self.type_properties.LastTriggerUSec / m
            )
            since_last_execute = now - last_execute
```

src/check_systemd_units.py
```diff
@@ -1,53 +1,250 @@
...
+                'since_last_execute / min_interval={}'
+                .format(
+                    str(self), min_interval, since_last_execute, last_execute,
+                    since_last_execute / min_interval
+                )
+            )
+            if (timer_warn <= since_last_execute / min_interval < timer_crit):
+                return (
+                    self._warn_level,
+                    'the timer wasn\'t launch since {}, look at {}'.format(
```
HerrSpace: * hasn't been launched
I think

src/check_systemd_units.py
```diff
@@ -1,53 +1,250 @@
...
+            )
+            logger.info(
+                '{}: min_interval={}, since_last_execute={}, last_execute={}, '
+                'since_last_execute / min_interval={}'
+                .format(
+                    str(self), min_interval, since_last_execute, last_execute,
+                    since_last_execute / min_interval
+                )
+            )
+            if (timer_warn <= since_last_execute / min_interval < timer_crit):
```
HerrSpace: I don't quite get the conditional, shouldn't it be `since_last_execute - min_interval`? Also if you move the critical check above the warning check, couldn't you remove the ` < timer_crit` part?

src/check_systemd_units.py
```diff
@@ -77,117 +282,115 @@ def parse_args():
...
-    return pattern == unit
-
+            remaining_units.append(unit)
 
-def check_unit(unit_name, serv_load, serv_active, serv_sub):
-    """Detect problems of a unit"""
-    if serv_load != 'loaded':
-        if serv_active != 'inactive':
-            return Problem.not_loaded_but_not_inactive
+    units[:] = remaining_units
```
HerrSpace: Interesting way to keep the reference, haven't seen this before.

src/check_systemd_units.py
```diff
@@ -1,53 +1,247 @@
...
-    """Enum for problems that can apply to the units"""
 
-    # From more important to less
-    failed = 0
-    activating_auto_restart = 1
-    not_loaded_but_not_inactive = 2
-    not_loaded_but_not_dead = 3
-    dead = 4
-    not_loaded = 5
+class Codes(int, Enum):
```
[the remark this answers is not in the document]
Felixoid: In fact, it's an alias and do exactly the same

src/check_systemd_units.py
```diff
@@ -1,53 +1,247 @@
 #!/usr/bin/env python
```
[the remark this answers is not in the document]
Felixoid:
```
┌[19:36:17] github/innogames/igmonplugins/ (systemd_dbus_timers|✔) 
└>±> python3 ./src/check_systemd_units.py -a
WARNING: the unit is failed: bluetooth.service; the timer wasn't launch since 1970-01-01 01:00:00, look at systemd-tmpfiles-clean.service: systemd-tmpfiles-clean.timer
┌[19:38:40] github/innogames/igmonplugins/ (systemd_dbus_timers|✔) 
└>±> python2 ./src/check_systemd_units.py -a
WARNING: the unit is failed: bluetooth.service; the timer wasn't launch since 1970-01-01 01:00:00, look at systemd-tmpfiles-clean.service: systemd-tmpfiles-clean.timer
```
Works quite well in both versions

src/check_systemd_units.py
```diff
@@ -1,53 +1,247 @@
...
+        '''
+        if hasattr(self, '__type_properties'):
+            return self.__type_properties
+        properties = self.__properties_interface.GetAll(
+            self.__type_interface
+        )
+        self.__type_properties = Property()
+        for k, v in properties.items():
+            setattr(self.__type_properties, k, v)
+        return self.__type_properties
```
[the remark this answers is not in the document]
Felixoid: It just doesn't make a sense. I'll show you tomorrow how terribly designed python-systemd-dbus and why we could use neither Unit nor Manager and any specific unit-type class

src/check_systemd_units.py
```diff
@@ -1,53 +1,250 @@
...
+        if self.unit_properties.LoadState != 'loaded':
+            if self.unit_properties.ActiveState != 'inactive':
+                return (
+                    Codes.CRITICAL,
+                    'the unit is not loaded but not inactive'
+                )
+            return (self._warn_level, 'the unit is not loaded')
+
+        if self.unit_properties.ActiveState == 'failed':
+            return (self._crit_level, 'the unit is failed')
```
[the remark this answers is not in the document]
Felixoid: Restart loop cause `ActiveState==failed` quite fast. We could make an experiment

src/check_systemd_units.py
Felixoid: No. It's different ways to define the intervals for related service launch. See [here](https://www.freedesktop.org/software/systemd/man/systemd.timer.html#OnActiveSec=)

src/check_systemd_units.py
```diff
@@ -1,53 +1,250 @@
...
+        '''
+        Detects problems for a timer unit
+        '''
+        checked_intervals = [
+            'OnUnitActiveUSec',
+            'OnActiveUSec',
+            'OnUnitInactiveUSec',
+        ]
+        # Microseconds to seconds
+        m = 1000000
```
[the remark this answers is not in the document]
Felixoid: just `million`, to making a code shortener

src/check_systemd_units.py
Felixoid: Thank you, I will try

src/check_systemd_units.py
Felixoid: About swapping cases - good point.
About `since_last_execute - min_interval` - no, warn and crit are relative values.

src/check_systemd_units.py
```diff
@@ -1,53 +1,250 @@
...
+                    'the timer wasn\'t launch since {}, look at {}'.format(
+                        last_execute, self.type_properties.Unit
+                    )
+                )
+        service_unit = SystemdUnit(
+            systemd_manager.get_unit(self.type_properties.Unit)
+        )
+        return service_unit.check(
+            timer_warn, timer_crit, self.__critical, timer=True
+        )
```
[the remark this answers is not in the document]
Felixoid: No, `self.type_properties.Unit` contains the name for the timer-related service unit. And we get it via `get_unit` by the name

src/check_systemd_units.py
Felixoid: Emre has made it :-j

src/check_systemd_units.py
Felixoid: We check, in fact, if the minimum of possible [intervals](https://www.freedesktop.org/software/systemd/man/systemd.timer.html#OnActiveSec=) has triggered the related service less, then `since_last_execute / warn_or_crit_threshold`

E.g., OnUnitActiveSec=60 and OnUnitInactiveSec=30 seconds. The minimum is 30. The warning will be triggered on since_last_execute=90 (3 * 30) seconds and crtitical on since_last_execute=210 (7 * 30)

src/check_systemd_units.py
Felixoid: It breaks the python2 compatibility:
```
last_execute = datetime.fromtimestamp(             
    self.type_properties.LastTriggerUSec / m       
)                                                  
since_last_execute = now - last_execute.timestamp()

AttributeError: 'datetime.datetime' object has no attribute 'timestamp'
```
`datetime.datetime` has method `timestamp` only since 3.4

src/check_systemd_units.py
Felixoid: Thank you

src/check_systemd_units.py
Felixoid: Could you, maybe, suggest the name for the variable `since_last_execute / min_interval`?

src/check_systemd_units.py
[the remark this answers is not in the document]
HerrSpace: 🤔 Did you install an enum module for python2? It doesn't seem to be there of our control servers and I though it was only introduced in 3.4 or something.

src/check_systemd_units.py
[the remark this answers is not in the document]
HerrSpace: Ah cool, didn't know this custom typed Enum stuff before :)

src/check_systemd_units.py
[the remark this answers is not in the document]
HerrSpace: So if self is a timer unit, self.type_properties.Unit is the related service unit? Strange.

src/check_systemd_units.py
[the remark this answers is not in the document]
HerrSpace: Does it get ActiveState failed even if the service exists with 0? I agree, we should make an experiment with a unit that has restart=always enabled.

src/check_systemd_units.py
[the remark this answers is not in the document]
HerrSpace: I know but `p[1] / m for p in intervals` is not very obvious.

src/check_systemd_units.py
HerrSpace: That seems to confirm what I wrote, no?

src/check_systemd_units.py
HerrSpace: Ah indeed. Darn python2.

src/check_systemd_units.py
HerrSpace: Ah I think those names are actually fine, it's more the timer_warn that confused me. I think adding your explanation in the code should make it clear enough 👍

HerrSpace:
I'm hesitantly approving this. I don't really like the result, due to missing history and heavy non essential changes it's hard to understand what's really going on here. I think the feature this check is IMO really about, checking timers execution, is important though and I'm not willing to spend more time on it.

PR #137, opened 2019-02-05
Reduce computation for lbpool states check

mtkh:


src/states_lbpool.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def get_lbpools():
+    try:
+        with open("/etc/iglb/lbpools.json") as jsonfile:
+            lbpools_obj = json.load(jsonfile)
+    except IOError:
+        with open('/etc/iglb/iglb.json') as jsonfile:
+            lbpools_obj = json.load(jsonfile)['lbpools']
+    return lbpools_obj
```
TuxPowered42: Exceptions are for handling errors, this is not an error but rather a configuration option. Just iterate over both files and check if each exist and use open this one.

src/states_lbpool.py
```diff
@@ -0,0 +1,159 @@
...
+
+import json, subprocess, re
+
+WARNING = 0.70
+CRITICAL = 0.85
+nagios_service = 'states_lbpool'
+
+
+def main():
+    debug = False
```
TuxPowered42: Please make this a commandline argument. Or see comment about Nagios servers.

src/states_lbpool.py
```diff
@@ -0,0 +1,159 @@
...
+                            default_state_limit,
+                            nagios_service,
+                            separator,
+                            )
+
+    if debug:
+        print("default_state_limit is {}\n".format(default_state_limit))
+        print(send_msg)
+    else:
+        for monitor in ('af-monitor.admin.ig.local', 'aw-monitor.ig.local'):
```
TuxPowered42: Please make nagios servers configurable from commandline parameters. Even better, lack of Nagios servers could mean displaying output to stdout, that is the debug mode.

src/states_lbpool.py
```diff
@@ -0,0 +1,159 @@
...
+    msg = ''
+    for lbname, lb_params in lbpools.items():
+        statuses = 'States: '
+        exit_code = local_exit_code = 0
+        if lb_params['state_limit']:
+            state_limit = int(lb_params['state_limit'])
+        else:
+            state_limit = default_state_limit
+        lbpool = lb_params['pf_name']
+        if lb_params['default_snat'] != True and lb_params['protocol_port'] and \
```
TuxPowered42: No need to check booleans for exact True.

src/states_lbpool.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # The default limit set in pf.conf, fetch it from what is loaded in pf
+    default_limits, err1 = subprocess.Popen(['sudo', 'pfctl', '-sm'],
+                                            stdout=subprocess.PIPE,
+                                            stderr=subprocess.PIPE).communicate()
+
+    states_dict = pfctl_parser(pfctl_output)
+
+    # Hard state limit is always printed in first line
+    default_state_limit_line = default_limits.decode().splitlines()[0]
```
TuxPowered42: Don't trust which line it is. It might change at any point.

src/states_lbpool.py
```diff
@@ -0,0 +1,159 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import json, subprocess, re
+
+WARNING = 0.70
+CRITICAL = 0.85
```
TuxPowered42: This should be configurable from command line arguments.

src/states_lbpool.py
mtkh: done

src/states_lbpool.py
mtkh: done

src/lbpool_states.py
```diff
@@ -0,0 +1,196 @@
...
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument(
+        '-H', dest='nsca_srv', nargs='+',
+        help='Nagios servers to report the results to'
+    )
+
+    parser.add_argument(
+        '-W', dest='warning', type=int, default=70,
```
TuxPowered42: Those are lower case parametrs per standard Nagios convenction and full word parameters should be there too, see https://nagios-plugins.org/doc/guidelines.html#AEN302

src/lbpool_states.py
```diff
@@ -0,0 +1,196 @@
...
+            try:
+                with open(config) as jsonfile:
+                    dict = json.load(jsonfile)
+                    if 'lbpools' in dict.keys():
+                        lbpools_obj = dict['lbpools']
+                    else:
+                        lbpools_obj = dict
+            except IOError:
+                print("Config file not found")
+                exit(1)
```
TuxPowered42: Exiting from inside of program is highly discouraged. You should return through normal path to main function if there is any cleanup to be done or just don't catch this exception at all. Your hadler just prints a message and quits anyway.

src/lbpool_states.py
```diff
@@ -0,0 +1,196 @@
...
+        exit_code = local_exit_code = 0
+        if lb_params['state_limit']:
+            state_limit = int(lb_params['state_limit'])
+        else:
+            state_limit = default_state_limit
+        lbpool = lb_params['pf_name']
+        critical = state_limit * (crit * 0.01)
+        warning = state_limit * (warn * 0.01)
+
+        if not lb_params['default_snat'] and \
```
TuxPowered42: You can put all conditions in `()`, then `\` are not necessary.

src/lbpool_states.py
```diff
@@ -0,0 +1,196 @@
...
+                lb_params['protocol_port'] and \
+                lb_params['nodes'] and \
+                lbpool in states_dict:  # To exclude those lbpools which have just
+                                        # been made but not deployed
+
+            cur_states = int(states_dict[lbpool]['cur_states'])
+
+            if cur_states >= critical:
+                local_exit_code = 2
+                statuses += 'Used states are above {}% of states limit | Total states limit: {}, Current states: {}'.format(
```
TuxPowered42: This line is over 80 characters long. Our code must be PEP-8 compliant.

src/lbpool_states.py
mtkh: added

src/lbpool_states.py
mtkh: ok, removed

src/lbpool_states.py
mtkh: I have changed it to some other place. its removed now.

src/lbpool_states.py
mtkh: Corrected, thanks.

PR #138, opened 2019-02-06
QEMU Version Match Check

lamaral:


src/check_qemu_version.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python
+"""InnoGames Monitoring Plugins - QEMU Version Check
+
+Copyright (c) 2018 InnoGames GmbH
```
TuxPowered42: It's 2019 already.

src/check_qemu_version.py
```diff
@@ -0,0 +1,128 @@
...
+def get_args():
+    parser = ArgumentParser(
+        description=(
+            'Check if libvirt domain is running with the installed qemu version'
+        )
+    )
+    parser.add_argument(
+        'domain',
+        type=str,
+        help="Libvirt domain to have version checked"
```
TuxPowered42: Will this check be assigned to VM object in Nagios so that given VM's sysadmin will be notified?

src/check_qemu_version.py
```diff
@@ -0,0 +1,128 @@
...
+    process = Popen(
+        cmd,
+        shell=True,
+        stdout=PIPE,
+        stderr=STDOUT,
+    )
+    for line in iter(process.stdout.readline, ''):
+        content += line
+    returncode = process.wait()
+    process.stdout.close()
```
TuxPowered42: There are better ways to run commands and read their output in Python 3. We don't need this code to be Python2-compatible if it runs on HVs.

src/check_qemu_version.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if returncode > 0:
+        return False
+
+    return content
+
+def check(args):
+    domain_qemu_pid = execute("pgrep -u libvirt-qemu -f '\-name {0}'".format(args.domain))
```
TuxPowered42: Wouldn't using psutil library be easier?

src/check_qemu_version.py
```diff
@@ -0,0 +1,128 @@
...
+    
+    # Hack to check for domain names w/o .ig.local suffix
+    if not domain_qemu_pid:
+        domain_qemu_pid = execute("pgrep -u libvirt-qemu -f '\-name {0}.ig.local'".format(args.domain))
+        if not domain_qemu_pid:
+            return ExitCodes.critical, "No libirt domain with the name '{0}' could be found on hypervisor '{1}'.".format(args.domain, platform.node())
+
+    domain_qemu_version = execute("sudo /proc/{0}/exe -version".format(domain_qemu_pid.rstrip()))
```
TuxPowered42: Nagios user has no permissions to use sudo. Our NRPE setup allows whole checks to be run as root, though.

src/check_qemu_version.py
lamaral: This is one thing that we have to discuss what our approach will be, as the script needs to be ran on the HV, but the result needs to go to the VM. 
We will probably have to take the same approach as the libvirt_hosts check.

src/check_qemu_version.py
lamaral: Yeah, I'll change that. I had done a pure 1:1 of what was in the ticket.

src/check_qemu_version.py
lamaral: I might actually be able to get this info using the libvirt package for Python.

src/check_qemu_version.py
TuxPowered42: Then maybe it would be better to send this information to Nagios together with the infamous libvirt_hosts check.

lamaral: Actually, because this doesn't need to aggregate info from all HVs, we can send back the results of the check straight on this script. I'll check with Serenus if the send_nsca is a good approach for this.

src/check_qemu_version.py
```diff
@@ -0,0 +1,181 @@
...
+def print_nagios_message(code, reason):
+    if code == ExitCodes.ok:
+        state_text = 'OK'
+    elif code == ExitCodes.warning:
+        state_text = 'WARNING'
+    elif code == ExitCodes.critical:
+        state_text = 'CRITICAL'
+    else:
+        state_text = 'UNKNOWN'
+    print("{0} - {1}".format(state_text, reason))
```
TuxPowered42: I think the standard is to use `:`.

src/check_qemu_version.py
```diff
@@ -0,0 +1,181 @@
...
+        return ExitCodes.warning, "QEMU version mismatch.\nVirtual machines {0} do not match HV QEMU version."+", ".join(filter(lambda x: x["status"] == CheckResult.mismatch, domains))
+    # If any machine couldn't have version retrieved
+    elif any(domain["status"] == CheckResult.unknown for domain in domains):
+        return ExitCodes.unknown, "Error obtaining QEMU versoin.\nCouldn't obtain QEMU versio for virtual machines {0}."+", ".join(filter(lambda x: x["status"] == CheckResult.unknown, domains))
+    else:
+        return ExitCodes.ok, "All versions match."
+
+def send_nsca(output):
+    if SEND:
+        for monitor in ('af-monitor.admin', 'aw-monitor'):
```
TuxPowered42: Please make those command line arguments, maybe with `-H`, as @mtkh already used those in his check.

src/check_qemu_version.py
```diff
@@ -0,0 +1,181 @@
...
+        state_text = 'CRITICAL'
+    else:
+        state_text = 'UNKNOWN'
+    print("{0} - {1}".format(state_text, reason))
+
+
+def execute(cmd):
+    process = Popen(
+        cmd,
+        shell=True,
```
TuxPowered42: Please never run commands in shell, this allows for code injections.

src/check_qemu_version.py
lamaral: The majority of them on Nagios is using ` - `.

src/check_qemu_version.py
lamaral: Removed.

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+                'hvname': hvname
+                }
+            domains.append(domain)
+    return domains
+
+
+def check_versions(domains):
+    # Get installed QEMU version
+    installed_qemu_version = execute(
+        "/usr/bin/qemu-system-x86_64 -version".split(" ")
```
sponkydive: please use single quotes everywhere

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+    if not installed_qemu_version:
```
sponkydive: if not execute(['/usr/bin/qemu-system-x86_64', '-version'])
should be enough. Make a comment to say it is installed versions.

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+        return CheckResult.unknown, "Could not retrieve QEMU version for "
+        "hypervisor '{0}'.".format(platform.node())
+
+    for domain in domains:
+        domain["version"] = execute(
+            "/proc/{0}/exe -version".format(domain['pid']).split(" ")
+            )
```
sponkydive: `
execute(['/proc/{}/exe'.format(domain['pid']), '-version'])
`
you can also define the comamnd in the for loop like:
`
cmd = ['/proc/{}/exe'.format(domain['pid']), '-version']
execute(cmd)
`
No need to define the index for format anymore since python 2.5 or something.

lamaral: The second way would make it troublesome to get the version value back into domain['version'].

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+            domain["status"] = CheckResult.match
+        else:
+            domain["status"] = CheckResult.mismatch
+
+    return domains
+
+
+def build_nsca_output(domains):
+    output = ""
+    for domain in domains:
```
sponkydive: i would set the status here instead iterating over it again later
`mystatus = None`

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+            "QEMU domain version DOES NOT match HV {2} version. Domain:{3} "
+            "Hypervisor:{4}\x17".format(
+                domain["vmname"], ExitCodes.warning, domain["hvname"],
+                domain["version"], installed_qemu_version)
+        elif domain["status"] == CheckResult.unknown:
+            output += "{0}\tqemu_version\t{1}\tUNKNOWN - "
+            "QEMU domain version could not be determined on HV {2}.".format(
+                domain["vmname"], ExitCodes.unknown, domain["hvname"])
+
+    return output
```
sponkydive: return output, status

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+    args = get_args()
+
+    # Get domain list
+    domains = get_domain_list()
+
+    # Get back domains with version result
+    domains = check_versions(domains)
+
+    # Generate output for NSCA
+    nsca_output = build_nsca_output(domains)
```
sponkydive: nsca_output, mystatus = build_nsca_output(domains)

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+
+    # Push NSCA results
+    nsca = send_nsca(args.hosts, nsca_output)
+    if not nsca:
+        print_nagios_message(ExitCodes.critical,
+                             "Error when pushing VM results with NSCA")
+
+    # Generate plugin results for HV
+    code, reason = build_plugin_output(domains)
```
sponkydive: code, reason = build_plugin_output(domains, mystatus)

src/check_qemu_version.py
```diff
@@ -0,0 +1,226 @@
...
+def print_nagios_message(code, reason):
+    if code == ExitCodes.ok:
+        state_text = 'OK'
+    elif code == ExitCodes.warning:
+        state_text = 'WARNING'
+    elif code == ExitCodes.critical:
+        state_text = 'CRITICAL'
+    else:
+        state_text = 'UNKNOWN'
+    print("{0} - {1}".format(state_text, reason))
```
sponkydive: please remove the index

src/check_qemu_version.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            domain['status'] = CheckResult.mismatch
+
+    return domains
+
+
+def build_nsca_output(domains):
+    nsca_output = ''
+    mismatch = False
+    unknown = False
```
sponkydive: both booleans are kind of duplicates.
Just remove them and check later if the lkists are empty
`if mismatch_doms:`
this is equal to mismatch = True

src/check_qemu_version.py
```diff
@@ -0,0 +1,249 @@
...
+def print_nagios_message(code, reason):
+    if code == ExitCodes.ok:
+        state_text = 'OK'
+    elif code == ExitCodes.warning:
+        state_text = 'WARNING'
+    elif code == ExitCodes.critical:
+        state_text = 'CRITICAL'
+    else:
+        state_text = 'UNKNOWN'
+    print('{0} - {1}'.format(state_text, reason))
```
sponkydive: pls remove the index

PR #139, opened 2019-02-11
Return all libvirt domains in check_libvirt_hosts.

lamaral:
This is part of TECH-9271 requirements.

I added the list of domains in the HV without touching the way the plugin initially worked.

### DIFF
--- a/src/check_libvirt_hosts.py
+++ b/src/check_libvirt_hosts.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """InnoGames Monitoring Plugins - Libvirt Hosts Check
 
-Copyright (c) 2017 InnoGames GmbH
+Copyright (c) 2019 InnoGames GmbH
 """
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the 'Software'), to deal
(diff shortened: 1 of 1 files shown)


src/check_libvirt_hosts.py
```diff
@@ -33,15 +102,19 @@ def main():
...
     if inactive_domains:
         print('WARNING: ' + ', '.join(
             '{} is defined but not running'.format(d.name())
             for d in inactive_domains
         ))
+        print_domains(domains)
         exit(1)
 
     print('OK: all defined domains are running')
+    print_domains(domains)
```
TuxPowered42: Printing domains and exiting is a common operation, not depending on inactive domains.

src/check_libvirt_hosts.py
```diff
@@ -33,15 +102,19 @@ def main():
         print('WARNING: could not connect to libvirt: ' + str(error))
         exit(1)
 
-    inactive_domains = [d for d in conn.listAllDomains() if not d.isActive()]
+    domains = conn.listAllDomains()
+
+    inactive_domains = [d for d in domains if not d.isActive()]
     if inactive_domains:
         print('WARNING: ' + ', '.join(
             '{} is defined but not running'.format(d.name())
```
TuxPowered42: I think this will print the whole sentence for each domain. Nobody's gonna read that.

lamaral: Please notice that the change here was simply moving `conn.listAllDomains()` out of the comprehensive list, so I could reuse it later for the print_domains().

src/check_libvirt_hosts.py
lamaral: I did it that way to keep the original flow of the check. It will still throw a warning in case any domains are not running, as the check for duplicate domains will be done separately through passive check.

lamaral:
I have made it somewhat better. Couldn't get rid of one of the exits, tho.

src/check_libvirt_hosts.py
TuxPowered42: OK

PR #140, opened 2019-02-14
WIP: Grafsy helper for igmonplugins

mtkh:


src/lbpool_states.py
```diff
@@ -52,13 +53,22 @@ def main():
...
+
+
     states_dict = pfctl_parser(pfctl_output)
 
     lbpools = get_lbpools()
 
     lbpools = {
         lbname: {
-            'state_limit': lb_params['state_limit'],
+            'state_limit': int(lb_params['state_limit']) if lb_params['state_limit'] else default_state_limit,
```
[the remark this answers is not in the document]
HerrSpace: I recommend you install falke8 and mypy linters in your editor.

src/send_grafsy.py
```diff
@@ -0,0 +1,60 @@
...
+
+    for k1, v1 in data.items():
+        template = k1.replace('.', '_') + '.'
+        msg += fetch(template, v1)
+
+
+    with tempfile.NamedTemporaryFile('w', delete=False) as tmpfile:
+        grafsy_file = "/tmp/grafsy/" + tmpfile.name.split('tmp/')[1]
+        tmpfile.write(msg)
+        tmpfile.flush()
```
HerrSpace: No need to flush, python will flush the files automagically on closing this file. I will close the file when you leave this with block.

src/send_grafsy.py
```diff
@@ -0,0 +1,60 @@
...
+        tmpname = tmpfile.name
+        chmod(tmpname, 0o644)
+        rename(tmpname, grafsy_file)
```
HerrSpace: Instead of renaming it here you can give `dir='/tmp/grafsy'` to the NamedTemporaryFile constructor. Renaming it here could make the with block unhappy anyway.

src/send_grafsy.py
```diff
@@ -0,0 +1,60 @@
+"""InnoGames Monitoring Plugins - Helper for sending metrics to grafana and
```
HerrSpace: Graphtite, not Grafana.

src/send_grafsy.py
```diff
@@ -0,0 +1,60 @@
...
+    if isinstance(v1, dict):
+        for k2, v2 in v1.items():
+            prefix = template
+            prefix += k2.replace('.', '_') + '.'
+            ret = fetch(prefix, v2)
+            msg2 += ret
+    else:
+        msg2 = template.rstrip('.') + ' ' + str(v1) + ' ' + (str(int(time()))) + '\n'
+
+    return msg2
```
HerrSpace: I can't even follow what exactly this function is supposed to format, how and why. Please add doc strings and proper variable names.

HerrSpace:
Maybe @Woellchen has some good ideas here. Kajetan has already hinted at wanting to merge igcollect and igmonplugins but I'm not yet sure what he has in mind exactly. Maybe we should split our scripts into 1) data collection 2) monitoring style reporting and 3) metric sending where 2 and 3 type scripts can include from type 1 or something.

src/lbpool_states.py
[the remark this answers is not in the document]
Woellchen: also pycharm comes with an auto-format feature. just set the line limit to 79 (the 80th character is reserved for newline) and hit autoformat. it does most stuff already for you automatically

Woellchen:
i also think grafsy doesn't belong here at all. the right place is definitely igcollect and not igmonplugins.  i like the idea of helpers in general, because this repository is just a loose collection of random scripts imo, but a monitoring check has nothing to do with metrics logging does it?
also whats the deal with grafsy anyhow? are we using any special features of it? because grafsy is otherwise just a proxy to graphite and accepts metrics in the same way. thats why i'd consistently call if graphite and not grafsy or even grafana everywhere
we can just make a simple igcollect script out of it, let it print the metrics to stdout like all the other scripts and then the rest is done in puppet like we do it normally
then again we can just have an aggregator that monitors graphite metrics and we don't even need this monitoring check in this repo at all right?

HerrSpace:
Writing to /tmp/grafsy/ is actually grafsy specific, otherwise I concur. As this is a passive check already, generating it from graphite would not even pose any downsides from what we currently have.

src/lbpool_states.py
```diff
@@ -52,13 +53,22 @@ def main():
                                             stdout=subprocess.PIPE,
                                             stderr=subprocess.PIPE).communicate()
 
+    # Hard state limit is always printed in first line
+    default_limit_lines = default_limits.decode().splitlines()
+
+    for line in default_limit_lines:
+        if "states" in line:
+            default_state_limit = int(
+                ''.join(default_limit_lines).split(' ')[-1])
```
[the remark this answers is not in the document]
mtkh: true

mtkh:
@Woellchen We need two things here and mostly, one is to send check results to nagios and the other is to send metrics/performance data to graphite and we do not want to query the source (pf in this case) twice which I think will happen if we create separate scripts in this case, one for nagios and other for igcollect. With this technique, we avoid query pf twice as that is expensive and just send the data which is already in memory to grafana. Also, we can use it at many other places for our hwlbs and routers.

Woellchen:
@mtkh i totally get your point but the solution is not to fake an igcollect plugin as a monitoring check. what i tried to suggest to you is that you can push those metrics to graphite and then setup a monitoring check that monitors graphite metrics. you don't need a monitoring check in this repository at all

src/send_grafsy.py
mtkh: please read this article, the last part explains about why flushing is important. I just followed this article. guide if you think this is not true https://blog.gocept.com/2013/07/15/reliable-file-updates-with-python/

src/send_grafsy.py
mtkh: We want atomicity for writing this file to grafsy

mtkh:
I would not rely on graphite for monitoring all of our 2500+ lbpools. Nagios is primary monitoring and therefore I want to send the check results directly to Nagios first. We could send it from Nagios to Graphite but that option doesn't work currently

src/send_grafsy.py
mtkh: I have added a doc string

src/lbpool_states.py
```diff
@@ -111,6 +113,19 @@ def main():
             )
             nsca.communicate(send_msg.encode())
 
+    lbpools_igcollect = {
+        'network' :
+            {
+                'lbpools' : {
+                    lbname : lb_params
+                    for lbname, lb_params in lbpools.items()
+                    if lb_params.pop('carp_master')
```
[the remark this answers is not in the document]
mtkh: yes sure

TuxPowered42:
The problem with collecting data from pf is that it is a rather costly operation and I've seen similar opeartions overloading loadbalancers especially if they overlapped. Collecting data for Nagios and Graphite *must* stay within a single script for pf.

Another issue is that we are writing the same code twice: for igcollect and igmonplugins. This is true for multiple other collectors. We just never had any motivation to unify those two projects and with this particular pf collector we finally have. Let's put it to good use and start merging. I really don't care where the code goes, even if that means starting a new project, but it has to eventually be done.

Woellchen:
i am citing myself again because i have the feeling it is simply ignored:
> i totally get your point but the solution is not to fake an igcollect plugin as a monitoring check. what i tried to suggest to you is that you can push those metrics to graphite and then setup a monitoring check that monitors graphite metrics. you don't need a new monitoring check in this repository at all

TuxPowered42:
Do we have already existing methods of writing Nagios checks against Graphite?

src/lbpool_states.py
```diff
@@ -52,13 +52,24 @@ def main():
                                             stdout=subprocess.PIPE,
                                             stderr=subprocess.PIPE).communicate()
 
+    # Hard state limit is always printed in first line
+    default_limit_lines = default_limits.decode().splitlines()
+
+    for line in default_limit_lines:
+        if "states" in line:
+            default_state_limit = int((line).split(' ')[-1])
```
HerrSpace: The extra () around line is not required.

mtkh: yep, changed. thx

Woellchen:
yes we do :) we can configure everything in puppet

src/send_grafsy.py
HerrSpace: Pretty cool article on defensive file system operations, thanks! I think you misunderstood that flushing part though. Here's the bit in question:
> With the **truncate-write pattern**, we have to issue an fsync after finishing write operations but before closing the file. Note that there is usually another level of write caching involved. The glibc buffer holds back writes inside the process even before they are passed to the kernel. To get the glibc buffer empty as well, we have to flush() it before fsync’ing:

```python
with open(filename, 'w') as f:
   model.write(f)
   f.flush()
   os.fdatasync(f)
```

To rephrase: The `fdatasync` is required for truncate-write to not suck. The `flush` is required for `fdatasync` to do anything usefull as otherwise `fdatasync` would not sync the glibc-cashed bits. You are not using the trancate-write pattern, but something close to the spooldir pattern though. That's why you don't need the `fdatasync` and why the automatic `close+flush` python gives you when leaving the context manager is enough.

src/send_grafsy.py
```diff
@@ -0,0 +1,65 @@
...
+
+    with tempfile.NamedTemporaryFile('w', delete=False) as tmpfile:
+        grafsy_file = "/tmp/grafsy/" + tmpfile.name.split('tmp/')[1]
+        tmpfile.write(msg)
+        tmpfile.flush()
+        tmpname = tmpfile.name
+        chmod(tmpname, 0o644)
+
+    # We want Atomicity for writing files to grafsy
+    rename(tmpname, grafsy_file)
```
HerrSpace: I think we don't really care for atomicity as the whole grafsy chain is best-effort delivery anyway, but it doesn't hurt either. If you want to do it I suggest you move the `"/tmp/grafsy/" + tmpfile.name.split('tmp/')[1]` line here too.

src/send_grafsy.py
HerrSpace: These variable names still have no meaning.

src/send_grafsy.py
mtkh: moved

src/send_grafsy.py
mtkh: I have now renamed the variables

src/send_grafsy.py
```diff
@@ -29,37 +29,36 @@
...
 
     with tempfile.NamedTemporaryFile('w', delete=False) as tmpfile:
-        grafsy_file = "/tmp/grafsy/" + tmpfile.name.split('tmp/')[1]
-        tmpfile.write(msg)
-        tmpfile.flush()
+        tmpfile.write(output)
         tmpname = tmpfile.name
         chmod(tmpname, 0o644)
 
+    grafsy_file = "/tmp/grafsy/" + tmpfile.name.split('tmp/')[1]
```
HerrSpace: tmpfile is not defined here. You need to use the tmpname you noted done within the context manager.

src/lbpool_states.py
```diff
@@ -292,6 +293,42 @@ def pfctl_parser(pfctl_output):
...
+    data = ''
+    if isinstance(v1, dict):
+        for k2, v2 in v1.items():
+            prefix = template
+            prefix += k2.replace('.', '_') + '.'
+            ret = carbonize(prefix, v2)
+            data += ret
+    else:
+        data = template.rstrip('.') + ' ' + str(v1) + ' ' + (
+            str(int(time()))) + '\n'
```
HerrSpace: You could change this to use datetime. This way you'd always get UTC independent of system time and it's easier to read then `str(int(time())))`
```python
from datetime import datetime
datetime.utcnow().strftime("%s")
```

src/lbpool_states.py
```diff
@@ -281,6 +293,42 @@ def pfctl_parser(pfctl_output):
 
     return states_dict
 
+def send_grafsy(data):
```
HerrSpace: Two blank lines after function definition please.

src/lbpool_states.py
```diff
@@ -281,6 +293,42 @@ def pfctl_parser(pfctl_output):
...
+        tmpname = tmpfile.name
+        chmod(tmpname, 0o644)
+
+    grafsy_file = "/tmp/grafsy/" + tmpfile.name.split('tmp/')[1]
+    # We want Atomicity for writing files to grafsy
+    rename(tmpname, grafsy_file)
+
+    return output
+
+def carbonize(template, v1):
```
HerrSpace: Two blank lines after function definition please.

src/send_grafsy.py
mtkh: I have changed it although it is working fine with tmpfile, i dont know how..

HerrSpace:
Good enough. Next step: Move this to igcollect and generate check from graphite.

src/lbpool_states.py
mtkh: changed

PR #141, opened 2019-02-20
Update check_lbpools to work with new config pulling

TuxPowered42:


src/check_lbpools.py
```diff
@@ -34,181 +34,156 @@
...
+
+        # Skip LB Pools for which this HWLB is not master carp
+        first_lbnode = list(pool_v['nodes'].values())[0]
+        if not (
+            first_lbnode['route_network'] in carps_states
+            and
+            carps_states[first_lbnode['route_network']]['carp_master']
+        ):
+            print(
+                'Skipping LB Pool {} because carp is not master {}'.format(pool_k, pool_v['route_network'])
```
mtkh: It crosses 80 char limits here

PR #143, opened 2019-02-22
Remove stdout and stderr from check_qemu_version

lamaral:
The send NSCA was missing the redirection of stdout and stderr to DEVNULL.


lamaral:
@TuxPowered42 This is a minor fix to suppress output from send_nsca.

PR #144, opened 2019-03-01
Enhance QEMU version check output on mismatch

lamaral:


src/check_qemu_version.py
```diff
@@ -117,32 +118,36 @@ def get_domain_list():
...
-    if not hypervisor_qemu_version:
-        return CheckResult.unknown, 'Could not retrieve QEMU version for '
-        'hypervisor {}.'.format(platform.node())
-
+    match = version_re.match(result)
+    if match is None:
+        print_nagios_message(
+            CheckResult.unknown, 'Could not retrieve QEMU '
+            'version for hypervisor {}.'.format(platform.node()))
+    hypervisor_qemu_version = match.group(1)
```
Woellchen: this will fail if `match` is `None`

src/check_qemu_version.py
```diff
@@ -81,6 +81,7 @@ def print_nagios_message(code, reason):
     else:
         state_text = 'UNKNOWN'
     print('{} - {}'.format(state_text, reason))
+    exit(code)
```
Woellchen: i wouldn't expect a print function to exit the whole program. it was good where it was before imo

src/check_qemu_version.py
```diff
@@ -117,32 +118,36 @@ def get_domain_list():
...
     # Get QEMU on the hypervisor
-    hypervisor_qemu_version = execute(
+    result = execute(
         ['/usr/bin/qemu-system-x86_64', '-version']
         )
-    if not hypervisor_qemu_version:
-        return CheckResult.unknown, 'Could not retrieve QEMU version for '
-        'hypervisor {}.'.format(platform.node())
-
+    match = version_re.match(result)
```
Woellchen: let's extract this version retrieval code into a separate function. also try and unify it with the code that retrievs the domain version

src/check_qemu_version.py
```diff
@@ -158,7 +163,7 @@ def build_nsca_output(domains):
             mismatch_doms.append(domain['vmname'])
             nsca_output += ('{}\tqemu_version\t{}\tWARNING - '
```
Woellchen: let's also try and eliminate some unnecessary lines from the code. for example `            nsca_output += ('{}\tqemu_version\t{}\tWARNING - '` is almost the same on the other two elifs.

lamaral: I can't really find a way to get away from those lines. Although they are similar, they still differ somewhat greatly.

src/check_qemu_version.py
```diff
@@ -98,6 +96,16 @@ def execute(cmd):
     return content
 
 
+def parse_qemu_version(version):
+    r = re.compile(r'.*version\s([\d\.]+\([\w\d\.\s:\+-]+\)).*')
+    match = r.match(version)
+    # Return None if couldn't match string. Deal with None as couldn't 
+    # retrieve version.
+    if match is None:
+        return None
```
Woellchen: we should throw the exception right here i think

src/check_qemu_version.py
Woellchen: let's keep it then for now

PR #147, opened 2019-03-26
WIP: Add the API reachability check

lamaral:


src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+# Returns
+#
+# * Critical when at least one URL is not reachable
+#
+# Copyright (c) 2019 InnoGames GmbH
+#
+
+import sys
+from argparse import ArgumentParser
+from requests import (
```
kofrezo: I don't like installing an extra module just for a nagios check and adding it to my repositories. Using native urllib does the job too.

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+# Accepted HTTP codes: 200, 201, 301, 302, 307, 403
+#
+# Returns
+#
+# * Critical when at least one URL is not reachable
+#
+# Copyright (c) 2019 InnoGames GmbH
+#
+
+import sys
```
kofrezo: This import is not needed. exit is always available.

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+from argparse import ArgumentParser
+from requests import (
+    get,
+    RequestException,
+)
+
+
+class ExitCodes:
+    ok = 0
+    warning = 1
```
kofrezo: You don't use warning

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+    critical = 2
+    unknown = 3
```
kofrezo: You don't use unknown.

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+    for url in args.url_list:
+        check, domain = check_url(url)
+        if not check:
+            unreachable_urls.append(domain)
+            failure = True
+            failure_count += 1
+
+    if not failure:
+        output = 'OK: All URLs are reachable.'
+        exit_code = ExitCodes.ok
```
kofrezo: You can just return right here, no need to force the user to continue reading and keep this if branch in mind.

lamaral: I have got complaints before for exiting in the middle of an if, hence I'm doing it only in the "end".

Woellchen: actually it would be much cleaner to move this whole logic out of `main` and put it into a separate function. main is only responsible for parsing args, getting some result from somewhere and then print/exit accordingly

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+        output = 'CRITICAL: {} URLs not reachable. {}'.format(
+                failure_count, '; '.join(output_list)
+                )
+        exit_code = ExitCodes.critical
+
+    print(output)
+    exit(exit_code)
+
+
+def check_url(url, accepted_codes={200, 201, 301, 302, 307, 403}):
```
Woellchen: these default values don't really make sense. usually you would accept 1/2/3xx codes and return an error for 4/5xx.
let's use sane default and give the user the option to override

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+    try:
+        request = urlopen(url)
+    except HTTPError:
+        return False, {'url': url, 'error': 'exception'}
```
Woellchen: why would you return the url here? the url is already known because it was given as an argument

Woellchen: also you can just return the error if there is one, otherwise return None. no need for an additional boolean

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+    if not failure:
+        output = 'OK: All URLs are reachable.'
+        exit_code = ExitCodes.ok
+    else:
+        output_list = ([domain['url'] + ' - ' + str(domain['error'])
+                        for domain in unreachable_urls])
+        output = 'CRITICAL: {} URLs not reachable. {}'.format(
+                failure_count, '; '.join(output_list)
+                )
+        exit_code = ExitCodes.critical
```
Woellchen: no warning anywhere? :)

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+        if not check:
+            unreachable_urls.append(domain)
+            failure = True
+            failure_count += 1
+
+    if not failure:
+        output = 'OK: All URLs are reachable.'
+        exit_code = ExitCodes.ok
+    else:
+        output_list = ([domain['url'] + ' - ' + str(domain['error'])
```
Woellchen: we use format: `'{} - {}'.format(arg1, arg2)`

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+#
+# Returns
+#
+# * Critical when at least one URL is not reachable
+#
+# Copyright (c) 2019 InnoGames GmbH
+#
+
+from argparse import ArgumentParser
+# from requests import (
```
Woellchen: i am generally against commented code. it doesn't bring any value

src/check_api_reachability.py
```diff
@@ -0,0 +1,83 @@
...
+
+class ExitCodes:
+    ok = 0
+    critical = 2
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument(
+        '--url', dest='url_list',
```
Woellchen: this seems like it should be a positional argument and not a parameter. no need for `--url`. but maybe for `--status_codes`

src/check_api_reachability.py
lamaral: The script takes multiple URLs. It's there so we can identify which one is the troublesome.

Woellchen: this function takes only one url.

src/check_api_reachability.py
lamaral: Timo didn't ask for a warning 😅

src/check_api_reachability.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+#
+# Nagios API reachability check
+#
+# This is a Nagios check to check our external services,
+# that have to be reachable by our game servers.
+# Checks by default for HTTP and HTTPS connections.
+# Accepted HTTP codes: 200, 201, 301, 302, 307, 403
```
Woellchen: change comment :)

src/check_api_reachability.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument(
+        'url_list', metavar='URL', type=str, nargs='+'
+        )
+    parser.add_argument(
+        '--code', dest='codes_list', type=int, action='append',
+        default=None  # [200, 201, 301, 302, 307]
```
Woellchen: change comment :)

src/check_api_reachability.py
```diff
@@ -0,0 +1,108 @@
...
+                })
+        except (URLError, HTTPError) as e:
+            unreachable_urls.append({'url': url, 'error': str(e)})
+
+    return unreachable_urls
+
+
+def main():
+    args = parse_args()
+
```
Woellchen: lets remove those unnecessary newlines. we could group these few calls in 3 groups. get args, evaluate checks, return/print results. or i wouldn't even mind if get args and evaluate checks would be grouped together

Woellchen:
as always i would opt for proper indentations and trailing commas, because it makes code a whole lot more readable. so stuff like this: `
        output = '{} URLs with problem. {}'.format(
                len(output_list), '; '.join(output_list)
                )
`
would become: `
        output = '{} URLs with problem. {}'.format(
                len(output_list), '; '.join(output_list),
        )
`

PR #149, opened 2019-05-02
Warn on NSCA send failure

lamaral:
The qemu_version check had a behavior where it would go critical when failing to reach NSCA server to send passive check results. 
This PR changes that behavior to a warning.

### DIFF
--- a/src/check_qemu_version.py
+++ b/src/check_qemu_version.py
@@ -263,10 +263,10 @@ def main():
         exit(ExitCodes.unknown)
     except Exception:
         print_nagios_message(
-            ExitCodes.critical,
-            'An unknown error ocurried retrieving HV QEMU version'
+            ExitCodes.warning,
+            'An exception occurred when retrieving HV QEMU version'
         )
-        exit(ExitCodes.critical)
+        exit(ExitCodes.warning)
 
     # Build output
     nsca_output, mismatch_doms, unknown_doms \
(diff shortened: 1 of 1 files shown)


src/check_qemu_version.py
```diff
@@ -256,21 +261,22 @@ def main():
             ExitCodes.unknown, 'Could not retrieve hypervisor QEMU version'
         )
         exit(ExitCodes.unknown)
+    except Exception:
+        print_nagios_message(
+            ExitCodes.critical,
+            'An unknown error ocurried retrieving HV QEMU version'
+        )
+        exit(ExitCodes.critical)
```
mtkh: Why critical for an exception? I think a warning would be better for such cases

src/check_qemu_version.py
```diff
@@ -256,21 +261,22 @@ def main():
             ExitCodes.unknown, 'Could not retrieve hypervisor QEMU version'
         )
         exit(ExitCodes.unknown)
+    except Exception:
+        print_nagios_message(
+            ExitCodes.critical,
+            'An unknown error ocurried retrieving HV QEMU version'
```
mtkh: may be "An exception occurred..." instead would be better. Also is it `ocurried` or `occurred`?

src/check_qemu_version.py
lamaral: Valid point. I changed it.

src/check_qemu_version.py
lamaral: Oops. Fixed that.

PR #152, opened 2019-05-08
TECH-33529: Fix logic, force python3

elmatze79:


src/check_mysql_master_flapping.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
HerrSpace: FYI: This doesn't force anything. Puppet generates the nrpe commands config with explicit interpreters.

PR #155, opened 2019-05-21
Add jenkins plugin check scripts

konstantinbest:


src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+    args = parse_cmd_args()
+    username = args.username
+    password = args.password
+
+    url = 'https://{}/scriptText'.format(args.host)
+    head = {}
+
+    with open('/usr/lib/nagios/igmonplugins/plugin_vulnerability.groovy', 'r') \
```
TheFRedFox: PEP8 says not to backslash line-continuation. For line-continuation always use parantheses, however, in this case, just put the path into another variable.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+        if post_response[0] == '0':
+            print('OK - No Issues found')
+            exit(0)
+        else:
+            print('Critical - Issues found for: {}'.format(post_response))
+            exit(1)
+
+    except HTTPError:
+        req = requests.get(
+            'https://{}/crumbIssuer/api/xml?xpath='
```
TheFRedFox: What the heck is this request doing and why is it needed when the other one fails. Explain this please in some comments, why the except and new request is needed and what the new request is doing.

TheFRedFox: After reading the method doc, I assume this is somehow getting a token for authentication? Think about, if you want to make it as script arguments. And not trying both.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def post_script_data(url, data, head):
+    """Send script data to a given endpoint
+
+    :param: url: url of the target
+    :param; data: data wich will be send within the post
+    :param: head: needed headers for the post
+
+    :return: str
```
TheFRedFox: That's not the correct use for return type. And additionally you can also specify the type of a param:
``` python
"""
:param url: url of the target
:type url: str
...

:return: semantic description of return value
:rtype: str
"""
```

And actually in your case it is not `str` but `Response` for the return type.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            print('Critical - Issues found for: {}'.format(post_response))
+            exit(1)
+
+
+def post_script_data(url, data, head):
+    """Send script data to a given endpoint
+
+    :param: url: url of the target
+    :param; data: data wich will be send within the post
```
TheFRedFox: Semicolon, instead of colon. However, the trailing colon is not needed.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+            exit(0)
+        else:
+            print('Critical - Issues found for: {}'.format(post_response))
+            exit(1)
+
+
+def post_script_data(url, data, head):
+    """Send script data to a given endpoint
+
+    :param: url: url of the target
```
TheFRedFox: Trailing colon is not needed.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+    :param; data: data wich will be send within the post
+    :param: head: needed headers for the post
```
TheFRedFox: Trailing colon is not needed.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+
+    :return: str
+    """
+
+    args = parse_cmd_args()
+    username = args.username
+    password = args.password
+
+    response = requests.post(url, auth=(username, password),
```
TheFRedFox: For multiline function calls, we go for:
``` python
response = requests.post(
    url, auth=(username, password),
    data={'script': data}, headers=head
)
```

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+    """Send script data to a given endpoint
+
+    :param: url: url of the target
+    :param; data: data wich will be send within the post
+    :param: head: needed headers for the post
+
+    :return: str
+    """
+
+    args = parse_cmd_args()
```
TheFRedFox: Uh, this is bad. Don't do it. Just the main function should either get the arguments it needs passed in or calls the arg_parsing function by itself. But do not call it in other functions. Pass the arguments to the function instead.
In that case, there could be a param `auth` which actually holds already the value for the request, so it is kinda just forwarding it.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+    password = args.password
+
+    response = requests.post(url, auth=(username, password),
+                             data={'script': data}, headers=head)
+
+    return response
+
+
+def parse_cmd_args():
+    """parse arguments
```
TheFRedFox: Better no documentation than redundant and/or wrong documentation.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+
+    :return: str
```
TheFRedFox: It's definitely not `str`.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+    parser = argparse.ArgumentParser(
+        description='A small Nagios script to check'
+                    'for security vulnerabilities in'
+                    'Jenkins Plugins'
+    )
+    parser.add_argument('username', type=str)
```
TheFRedFox: The type `str` is default, so no need to explicitly set it here. But there are other options which are missing here. The host e.g. is mandatory, so you should set it to `required=True`.

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+plugin_vulnerabilities = new jenkins.security.UpdateSiteWarningsMonitor().getActivePluginWarningsByPlugin()
+def crit_plu = []
+
+plugin_vulnerabilities.each{
+  key,value->
```
TheFRedFox: `key, value ->` and I prefer it on the same line as the curly brace.

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+plugin_vulnerabilities = new jenkins.security.UpdateSiteWarningsMonitor().getActivePluginWarningsByPlugin()
+def crit_plu = []
+
+plugin_vulnerabilities.each{
```
TheFRedFox: Either `each({` or `each {`

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+plugin_vulnerabilities = new jenkins.security.UpdateSiteWarningsMonitor().getActivePluginWarningsByPlugin()
```
TheFRedFox: Please separate instantiation and method calling here for better Error handling.

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+def crit_plu = []
+
+plugin_vulnerabilities.each{
+  key,value-> 
+  plugin = key.getLongName()
+  crit_plu << plugin
+}
```
TheFRedFox: If `plugin_vulnerabilities` is a Map, then you could write this also with that:
``` groovy
crit_plu = plugin_vulnerabilities.keySet()*.longName
```

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+plugin_vulnerabilities = new jenkins.security.UpdateSiteWarningsMonitor().getActivePluginWarningsByPlugin()
+def crit_plu = []
```
TheFRedFox: `crit_plu` .. Just name it `critical_plugins` and it will be clear for everyone.

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+
+plugin_vulnerabilities.each{
+  key,value-> 
+  plugin = key.getLongName()
+  crit_plu << plugin
+}
+if (crit_plu.size() > 0) {
```
TheFRedFox: `if (!crit_plu.isEmpty())`

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+  def plu = crit_plu.join(", ")
+  print(plu)
+}else{
```
TheFRedFox: `} else {`

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+def crit_plu = []
+
+plugin_vulnerabilities.each{
+  key,value-> 
+  plugin = key.getLongName()
+  crit_plu << plugin
+}
+if (crit_plu.size() > 0) {
+  def plu = crit_plu.join(", ")
+  print(plu)
```
TheFRedFox: `print` vs `println` below. What's correct here?

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+}else{
+  println(0)
```
TheFRedFox: Is a `0` really mandatory? Or is an empty string also Ok? If, you could replace the `if-else` with just:
`println(crit_plu.join(", "))`

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,36 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
```
TheFRedFox: In general, we are here in a groovy script, so java based environment. Normally, you don't have snake_casing but camelCasing here.

TheFRedFox:
One more thing. I would not put the groovy script in the root src folder, but rather in some resource folder. It is not a script which should be called by nagios or in a standalone way here, but rather be loaded and sent somewhere to be executed. So, I would suggest to create some resource folder and put it there then.

src/plugin_vulnerability.groovy
konstantinbest: Done

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+
+    url = 'https://{}/scriptText'.format(args.host)
+    headers = {}
+
+    with open(script_path) as script:
+        data = script.read()
+
+    try:
+        response = post_script_data(url, data, headers, auth)
+        response.raise_for_status()
```
TheFRedFox: Unfortunately, this will also raise when there is something else than a 403 or 401. Even a Timeout would here be treated as an Authentication Failure. It is therefore needed to rather check for the actual http status and consider following actions then.
So, for 200 (or what else is sent back for a successful script run on Jenkins), the result should be printed directly.
For the http status which will be sent back for a missing CSRF token that other request should be sent to fetch the CSRF token and repeat the actual request.
For authentication failure http status, you could raise-exit the script with a reasonable message as username or password does not match.
For other failures like Timeout or the like just raise-exit the script with the error.
Also possible to exit safely with UNKNOWN and a reasonable error message.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+            print('Critical - Issues found for: {}'.format(post_response))
+            exit(1)
+
+
+def post_script_data(url, data, headers, auth):
+    """Send script data to a given endpoint
+
+    :param url: url of the target
+    :param data: data which will be send within the post
+    :param head: needed headers for the post
```
TheFRedFox: Replace `head` with `headers`

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+    args = parse_cmd_args()
+    username = args.username
+    password = args.password
+    auth = (username, password)
+    script_path = '/usr/lib/nagios/igmonplugins/plugin_vulnerability.groovy'
+
+    url = 'https://{}/scriptText'.format(args.host)
+    headers = {}
+
+    with open(script_path) as script:
```
TheFRedFox: Rename `script` to `fd` and `data` to `script`.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+
+    response = requests.post(
+        url, auth=auth, data={'script': data}, headers=headers
+    )
+
+    return response
+
+
+def parse_cmd_args():
+
```
TheFRedFox: Remove that redundant line.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+        req = requests.get(
+            'https://{}/crumbIssuer/api/xml?xpath='
+            'concat(//crumbRequestField,":",//crumb)'.
+            format(args.host), auth=auth
+        )
+
+        crumb = req.text.split(':')
+        headers[crumb[0]] = crumb[1]
+        post_response = post_script_data(url, data, headers, auth).text
+        if not post_response:
```
TheFRedFox: Even worse here, if through race condition Jenkins will be down e.g. exactly here, there won't be now content even though it is a Timeout. So, in the end it will print out, that everything is ok, but Jenkins is down.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+def main():
+    """Print the response of a given endpoint in Nagios format.
+
+    Fetch CSRF-Token if needed
+    """
+
+    args = parse_cmd_args()
+    username = args.username
+    password = args.password
+    auth = (username, password)
```
kofrezo: You can use auth = (args.username, args.password) directlly.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,120 @@
...
+    script_path = '/usr/lib/nagios/igmonplugins/plugin_vulnerability.groovy'
```
kofrezo: Maybe you should think about using a relative path here and rely on having it in the same directory or just store the contents of the groovy script directly in a variable to avoid manipulation of the groovy file which could then be abused for arbitary code execution on jenkins.

src/check_plugin_vulnerability.py
kofrezo: I agree here

src/check_plugin_vulnerability.py
kofrezo: +1

src/check_plugin_vulnerability.py
kofrezo: +1

konstantinbest: Some of the Jenkins server has CSRF protection enabled and expect the CSRF-Token in a request. This request is used for getting the token if the first request raises an http error.

src/check_plugin_vulnerability.py
TheFRedFox: Actually, I think thats a good idea to have the code inline in some variable. It is just 4 lines of code, we don't have that separated file, which needs to be deployed too and should rather be in some resources file. So, we absolutely would gain more advantage than we would loose.
👍 from me for that change.

Btw. arbitrary code execution would still be possible when you could change files on the server. As you then could also change the python files and therefore also change any execution there, so also requests to Jenkins.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+        "def plugin_vulnerabilities = warnings_monitor.getActivePluginWarningsByPlugin()\n"  # NOQA E501
+        "def critical_plugins = plugin_vulnerabilities.keySet()*.longName\n"
+        'print(critical_plugins.join(", "))'
+    )
+
+    url = 'https://{}/scriptText'.format(args.host)
+    headers = {}
+
+    response = post_script_data(url, script, headers, auth)
+    response.raise_for_status()
```
TheFRedFox: You are checking the status now, so do not raise for the status. 😛

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+    if response.status_code == 200:
+        if not post_response:
+            print('OK - No Issues found')
+            exit(0)
+        else:
+            print('Critical - Issues found for: {}'.format(post_response))
+            exit(1)
+    elif response.status_code == 403:
+        # CSRF Token needed, fetch token
+        req = requests.get(
```
TheFRedFox: You are getting a response, so use either `res` or `response` instead of `req` here.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+            'https://{}/crumbIssuer/api/xml?xpath='
+            'concat(//crumbRequestField,":",//crumb)'.
+            format(args.host), auth=auth
+        )
+        if req.status_code == 200:
```
TheFRedFox: Early exit-return:
``` python
if req.status_code != 200:
    print('UNKNOWN - Jenkins service is not reachable...')
    exit(2)
```

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,118 @@
...
+            crumb = req.text.split(':')
+            headers[crumb[0]] = crumb[1]
+            post_response = post_script_data(url, script, headers, auth).text
```
TheFRedFox: Also check http status code here, as it is not impossible that between fetching the csrf token and the retry request jenkins dies.

src/plugin_vulnerability.groovy
```diff
@@ -0,0 +1,31 @@
+/**
```
TheFRedFox: Delete this file.

src/check_plugin_vulnerability.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+import argparse
+import requests
+from sys import exit
+from requests.models import Response
+from requests.exceptions import HTTPError
+
+SCRIPT = (
+    "def warnings_monitor = new jenkins.security.UpdateSiteWarningsMonitor()\n"  # NOQA E501 Don't wrap because of groovy code
```
kofrezo: No double quotes needed here.

konstantinbest: Done

PR #156, opened 2019-05-21
Skip for smart_ids that are not in config file

mtkh:


src/check_smartmon.py
```diff
@@ -93,6 +93,10 @@
             element = elements.split(";")
             smart_id = int(element[0])
 
+            # Skip attributes that do not exists in our config
+            if not ALERTS[manufacturer].get(smart_id):
```
TheFRedFox: To check existence of a key, you rather use:
``` python
if smart_id not in ALERTS[manufacturer]:
```

PR #159, opened 2019-06-13
Change for only new style loadbalancers

mtkh:


src/lbpool_states.py
```diff
@@ -151,75 +151,37 @@ def args_parse():
...
+        dict = json.load(jsonfile)
 
-                if 'lbpools' in dict.keys():
-                    lbpools_obj = dict['lbpools']
-                else:
-                    lbpools_obj = dict
+        if 'lbpools' in dict.keys():
+            lbpools_obj = dict['lbpools']
+        else:
+            lbpools_obj = dict
```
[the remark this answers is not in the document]
mtkh: I have updated this section, please review

src/lbpool_states.py
```diff
@@ -151,75 +151,35 @@ def args_parse():
...
 
-                if 'lbpools' in dict.keys():
-                    lbpools_obj = dict['lbpools']
-                else:
-                    lbpools_obj = dict
+    try:
+        with open(config) as jsonfile:
+            lbpools_obj = json.load(jsonfile)
+    except IOError as e:
+        print("Exception occured while opening {} \n{}".format(config, e))
```
TuxPowered42: Double quotes where single quotes should be used.

src/lbpool_states.py
```diff
@@ -151,75 +151,35 @@ def args_parse():
...
-                    {
-                        'carp_master': v['carp_master']
-                    }
-                for k, v in carp_state.items()
-                for kn, vn in network['internal_networks'].items()
-                if k == kn
-                }
+    configs = '/var/run/iglb/carp_state.json'
+
+    with open(configs, 'r') as carp_statejson, \
```
TuxPowered42: The other `open` is in try/except block but those are not. Let's make them behave identically.

mtkh: updated

src/lbpool_states.py
```diff
@@ -151,75 +151,39 @@ def args_parse():
...
                         'carp_master': v['carp_master']
-                    }
+                }
                 for k, v in carp_state.items()
                 for kn, vn in network['internal_networks'].items()
                 if k == kn
-                }
+            }
+    except IOError as e:
+        print('Exception occured while opening {} \n{}'.format(configs, e))
```
TuxPowered42: That will write both filenames no matter which one failed. Are you sure you care for try/except at all? A normal failure would be just as good. Maybe @HerrSpace can help us with this.

mtkh: removed that part

PR #160, opened 2019-06-19
Add file age check

meikomeik:
This check was written by Jon Dawson.

### DIFF
--- a/src/check_file_age.py
+++ b/src/check_file_age.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - File Age Check
+
+This script checks the age of a file and raises a warning or critical state
+depending on the thresholds specified as arguments
+
+Copyright 2019 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from argparse import ArgumentParser
+import os
+import time
+
+def main():
+    """Main entrypoint, performs check of file age"""
+
+    args = get_parser().parse_args()
+
+    if not args.path:
+        print('UNKNOWN - no path supplied')
+        exit(3)
+
+    age = 0
+    try:
+        age = time.time() - os.path.getmtime(args.path)
+    except OSError:
+        print('UNKNOWN - {} is not present or no permissions'.format(
+            args.path))
+        exit(3)
+
+    status = 'OK'
+    code = 0
+
+    if age > args.critical*60:
+        code = 2
+        status = 'CRITICAL'
+    elif age > args.warning*60:
+        code = 1
+        status = 'WARNING'
+
+    print('{} - {} last changed {} minutes ago'.format(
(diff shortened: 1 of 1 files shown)


src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+import os
+import time
+
+def main():
+    """Main entrypoint, performs check of file age"""
+
+    args = get_parser().parse_args()
+
+    if not args.path:
+        print('UNKNOWN - no path supplied')
```
TheFRedFox: Why not making the argument required then?

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+        exit(3)
+
+    age = 0
```
TheFRedFox: No need to neither explicitly initialise nor declare it. `age` will be available outside the `try` scope too.

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+    try:
+        age = time.time() - os.path.getmtime(args.path)
+    except OSError:
+        print('UNKNOWN - {} is not present or no permissions'.format(
```
TheFRedFox: Apply Multiline style code:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+            args.path))
+        exit(3)
+
+    status = 'OK'
+    code = 0
+
+    if age > args.critical*60:
```
TheFRedFox: PEP8: Spaces around operators.

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+        code = 2
+        status = 'CRITICAL'
+    elif age > args.warning*60:
```
TheFRedFox: PEP8: Spaces around operators.

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+        code = 1
+        status = 'WARNING'
+
+    print('{} - {} last changed {} minutes ago'.format(
```
TheFRedFox: Apply Multiline style code:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+    exit(code)
+
+
+def get_parser():
+    """Get argument parser -> ArgumentParser"""
+
+    parser = ArgumentParser()
+
+    parser.add_argument('--path', '-p', help='path of file to check')
+    parser.add_argument(
```
TheFRedFox: Apply Multiline style code:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+        '--warning', '-w',
+        help='warning threshold in minutes for file age',
+        default=30, type=int)
+    parser.add_argument(
```
TheFRedFox: Apply Multiline style code:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+
+def get_parser():
+    """Get argument parser -> ArgumentParser"""
+
+    parser = ArgumentParser()
+
+    parser.add_argument('--path', '-p', help='path of file to check')
+    parser.add_argument(
+        '--warning', '-w',
+        help='warning threshold in minutes for file age',
```
TheFRedFox: Make `help` being the last argument standing alone.

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+        default=30, type=int)
+    parser.add_argument(
+        '--critical', '-c',
+        help='critical threshold in minutes for file age',
```
TheFRedFox: Make `help` being the last argument standing alone.

src/check_file_age.py
```diff
@@ -0,0 +1,82 @@
...
+
+    args = get_parser().parse_args()
+
+    if not args.path:
+        print('UNKNOWN - no path supplied')
+        exit(3)
+
+    age = 0
+    try:
+        age = time.time() - os.path.getmtime(args.path)
```
TheFRedFox: Would that also work on directories and would it be ok if this also works on these?

src/check_file_age.py
TheFRedFox: `is not present` is an uncommon statement for rather `does not exist`.

Morthy: was intentionally chosen due to different meaning

src/check_file_age.py
meikomeik: it also works on directories

src/check_file_age.py
```diff
@@ -0,0 +1,90 @@
...
+    elif age > args.warning * 60:
+        code = 1
+        status = 'WARNING'
+
+    print(
+        '{} - {} last changed {} minutes ago'.format(
+            status, args.path, round(age / 60)
+        )
+    )
+    exit(code)
```
TheFRedFox: Please use `sys.exit`. `quit` and `exit` are rather used for interactive shells. You will also need to import sys for that.

PR #161, opened 2019-06-26
Send results to new per-hwlb-group Nagios check

TuxPowered42:


src/check_lbpools.py
```diff
@@ -173,9 +187,15 @@ def compare_pools(lbpools, carps_states, lbpools_states, nsca_servers):
...
 
-def nagios_output(pool, exit_code, separator, message):
+def nagios_output(pool, exit_code, separator, message, hwlb_group):
     output = ''
-    for nagios_service in ['check_lbpool', ]:
+
+    nagios_services = ['check_lbpool']
+
+    if hwlb_group:
+        nagios_services += ['lbpool_nodes_{}'.format(hwlb_group)]
```
TheFRedFox: This looks like it could be rather just:
``` python
nagios_services.append('lbpool_nodes_' + hwlb_group)
```

lamaral: Indeed. I have updated it in the latest commit.

PR #162, opened 2019-07-04
Fix always returning critical on non OK checks in check_multiple.py

elmatze79:
Behaviour before:
Whenever one check in check_multiple returned something else than "OK", check_multiple would retun "CRITICAL"
Behavior now:
check multiple returns the worst state it finds in an individual check: CRITICAL > WARNING > UNKNOWN > OK

### DIFF
--- a/src/check_multiple.py
+++ b/src/check_multiple.py
@@ -31,6 +31,7 @@ parser.add_argument('iterate_over', nargs='+', help='Queue list separated by spa
 args=parser.parse_args()
 
 exit_code = 0
+unknown = False
 message = ''
 subs = {}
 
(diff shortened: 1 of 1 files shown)


src/check_multiple.py
```diff
@@ -39,9 +40,15 @@
 
 for iterate_param, process in subs.iteritems():
     out, err = process.communicate()
+    # Return the worst error code, because nagios interprets '3' as unknown we have to do some magic
     if process.returncode != 0:
-        exit_code = 2
+        if process.returncode == 3:
```
Woellchen: hm this looks a bit too complicated for me. i think if we remove the new `unknown` variable again and this whole `if/elif` we could end up with a simple `exit_code = max(exit_code, process.returncode)` no?

elmatze79: unfortunately UNKNOWN has the highest exit code, so if you have a CRITICAL and an UNKNOWN at the same time, it would return UNKNOWN

Woellchen: ok got that. my mistake

PR #166, opened 2019-07-16
Fix empty argument for min()

Felixoid:


src/check_systemd_units.py
```diff
@@ -186,12 +186,12 @@ def _check_timer(self, timer_warn, timer_crit):
             return (
                 self._crit_level, 'the timer is not active'
             )
-        intervals = self.type_properties.TimersMonotonic
+        intervals = [p[1] for p in self.type_properties.TimersMonotonic
```
TheFRedFox: Apply Multiline style code. Example:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

PR #167, opened 2019-07-17
Do not check timers unless they are executed after boot

Felixoid:


lamaral:
LGTM, but I would wait for @Woellchen opinion

src/check_systemd_units.py
```diff
@@ -189,7 +189,9 @@ def _check_timer(self, timer_warn, timer_crit):
         intervals = [p[1] for p in self.type_properties.TimersMonotonic
                      if p[0] in checked_intervals]
         logger.debug('Monotonic timers are: {}'.format(intervals))
-        if intervals:
+        # We check intervals only if timer has been triggered after reboot,
+        # otherwise LastTriggerUSec=0 (1970-01-01:00:00:00)
+        if intervals and self.type_properties.LastTriggerUSec != 0:
```
TheFRedFox: This check could be done already before line 189. No need to get the list of intervals if we do not need them.
But with that change the code is getting slowly but surely too complex.

A quick refactoring could be done by just separating the interval code part here into an own function and call it here.

PR #168, opened 2019-08-01
Elasticsearch

elmatze79:


src/check_es_index_exists.py
```diff
@@ -0,0 +1,63 @@
...
+
+def main():
+    args = parse_args()
+    es = Elasticsearch(args.eshost)
+    # suppress print output from es function
+    text_trap = io.StringIO()
+    sys.stderr = text_trap
+    if not es.indices.exists(args.index + '*', allow_no_indices=False):
+        print('Index not found in cluster: {}'.format(args.index))
+        sys.exit(1)
```
TheFRedFox: There is no Critical condition?

src/check_es_index_lifecycle.py
```diff
@@ -0,0 +1,66 @@
...
+    indices = es.indices.get_settings()
+    for index in indices:
+        # Except system indices
+        if index.startswith('.'):
+            continue
+        try:
+            indices[index]['settings']['index']['lifecycle']
+        except KeyError:
+            failed_indices.append(index)
+    if len(failed_indices) > 0:
```
TheFRedFox:
``` python
if failed_indices:
    ...
```
is sufficient.

src/check_es_index_lifecycle.py
```diff
@@ -0,0 +1,66 @@
...
+            'This Nagios check investigates every index (except '
+            'system indices) if an index lifecycle is attached. '
+            'This lifecycle takes care of deleting the index after '
+            'a specified amount of time.'
+        ),
+        formatter_class=RawTextHelpFormatter
+    )
+
+    parser.add_argument(
+        '--eshost', dest='eshost', required=True,
```
TheFRedFox: In an already `es` context, I think `--host` is sufficient and also more commen then.

TheFRedFox: An explicit `dest` is just necessary if it is different than the dashed name, so `--eshost` would already mapped to `eshost`.
So, no explicit `dest` needed here.

src/check_es_index_exists.py
```diff
@@ -0,0 +1,63 @@
...
+        description=(
+            'This Nagios check accepts an index prefix and checks '
+            'if a matching index exists. The parameter `blafasel` '
+            'would match an index named `blafasel-2019.01.01`'
+        ),
+        formatter_class=RawTextHelpFormatter
+    )
+
+    parser.add_argument(
+        '--eshost', dest='eshost', required=True,
```
TheFRedFox: In an already `es` context, I think `--host` is sufficient and also more commen then.

TheFRedFox: An explicit `dest` is just necessary if it is different than the dashed name, so `--eshost` would already mapped to `eshost`.
So, no explicit `dest` needed here.

src/check_es_cluster_state.py
```diff
@@ -0,0 +1,57 @@
...
+def parse_args():
+    parser = ArgumentParser(
+        description=(
+            'This Nagios check investigates the cluster health.'
+        ),
+        formatter_class=RawTextHelpFormatter
+    )
+
+    parser.add_argument(
+        '--eshost', dest='eshost', required=True,
```
TheFRedFox: In an already `es` context, I think `--host` is sufficient and also more commen then.

TheFRedFox: An explicit `dest` is just necessary if it is different than the dashed name, so `--eshost` would already mapped to `eshost`.
So, no explicit `dest` needed here.

TheFRedFox:
👍 for that nice structured PR! 🙂

src/check_es_cluster_state.py
kofrezo: If this parameter is requires a positional parameter without the dashes is preferred as it is required automatically.

src/check_es_index_exists.py
kofrezo: If this parameter is requires a positional parameter without the dashes is preferred as it is required automatically.

src/check_es_index_exists.py
```diff
@@ -0,0 +1,63 @@
...
+        help='Elasticsearch host to run the query against'
+    )
+    parser.add_argument(
+        '-i', '--index', dest='index', required=True,
```
kofrezo: If this parameter is requires a positional parameter without the dashes is preferred as it is required automatically.

src/check_es_index_exists.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    return parser.parse_args()
+
+def main():
+    args = parse_args()
+    es = Elasticsearch(args.eshost)
+    # suppress print output from es function
+    text_trap = io.StringIO()
+    sys.stderr = text_trap
```
kofrezo: Keep in mind that you are manipulating all stderr not only the one of es function. Besides that you can just disabled stderr or pipe it to /dev/null without the overhead of a StringIO file if you don't need it anyway.
```
>>> import sys
>>> sys.stderr = None # or open /dev/null with open.
>>> import os
>>> sys.stderr.write("spam")
>>> sys.stdout.write("spam")
```

src/check_es_index_lifecycle.py
kofrezo: If this parameter is requires a positional parameter without the dashes is preferred as it is required automatically.

src/check_es_index_exists.py
elmatze79: right

PR #169, opened 2019-08-05
Fix totally non-working check BGP peers

TuxPowered42:


lamaral:
Even better with the printf :)

PR #170, opened 2019-09-09
Fix systemd check on Buster

lamaral:
On Buster, the systemd-dbus contains `SuccessExitStatus` but that comes
back empty sometimes.

In order to mitigate the problem, we combine the returned array with a
`[0]` list.

### DIFF
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -140,16 +140,19 @@ class SystemdUnit:
             # See the man 5 systemd.service for ExecMainStatus and
             # SuccessExitStatus
             # All currently running services have ExecMainStatus=0
-            last_run_failed = (
-                (
-                    # SuccessExitStatus is not always present in older versions
-                    # of systemd
-                    hasattr(self.type_properties, 'SuccessExitStatus') and
-                    self.type_properties.ExecMainStatus not in
-                    self.type_properties.SuccessExitStatus[0]
-                )
-                or self.type_properties.ExecMainStatus != 0
-            )
+
+            # SuccessExitStatus is not always present in older versions
+            # of systemd.
+            # SuccessExitStatus[0] is an array that could be empty.
+            if hasattr(self.type_properties, 'SuccessExitStatus') and \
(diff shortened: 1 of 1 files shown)


src/check_systemd_units.py
```diff
@@ -146,7 +146,7 @@ def _check_service(self, timer=False):
                     # of systemd
                     hasattr(self.type_properties, 'SuccessExitStatus') and
                     self.type_properties.ExecMainStatus not in
-                    self.type_properties.SuccessExitStatus[0]
+                    self.type_properties.SuccessExitStatus[0] + [0]
```
Woellchen: this seems weird to me. we cannot just assume that `0` is always a success exit status. even when an empty return list is expected when nothing is explicitly set, then we must still not add `0` always. we should instead assume `0` is a success exit code when the list is empty and only when it is empty.

src/check_systemd_units.py
```diff
@@ -146,7 +146,7 @@ def _check_service(self, timer=False):
...
                 )
                 or self.type_properties.ExecMainStatus != 0
```
Woellchen: you haven't changed this line but we might still want to put it into context now: if the first condition in the parenthesis evaluates to false (meaning no fail) but `0` is actually not in success exit status, then this `or` would still make the check fail, although everything was ok. am i right? is this a bug?

src/check_systemd_units.py
```diff
@@ -140,16 +140,19 @@ def _check_service(self, timer=False):
...
-                    self.type_properties.ExecMainStatus not in
-                    self.type_properties.SuccessExitStatus[0]
-                )
-                or self.type_properties.ExecMainStatus != 0
-            )
+
+            # SuccessExitStatus is not always present in older versions
+            # of systemd.
+            # SuccessExitStatus[0] is an array that could be empty.
+            if hasattr(self.type_properties, 'SuccessExitStatus') and \
```
TuxPowered42: Please use `()` instead of `/`/

PR #171, opened 2019-09-10
add igvm_locked check

Jowe1999:
add igvm_locked check

### DIFF
--- a/src/check_igvm_locked.py
+++ b/src/check_igvm_locked.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """InnoGames Monitoring Plugins - IGVM locked Check
 This script checks if a vm or hypervisor is igvm_locked
 if so the check will go to the state warning never critical.
@@ -36,7 +36,7 @@ def main():
     )
     parser.add_argument("monitoring_master", type=str)
     parser.add_argument("--time-in-minutes", type=int, default=480)
-    parser.add_argument("-v")
+    parser.add_argument("-v", type=bool, default=False)
     args = parser.parse_args()
 
     max_minutes = args.time_in_minutes
--- a/src/check_igvm_locked.py
+++ b/src/check_igvm_locked.py
@@ -34,9 +34,9 @@ def main():
     parser = argparse.ArgumentParser(
         description='nagios check for long running igvm_locked attribute',
     )
-    parser.add_argument("monitoring_master", type=str)
-    parser.add_argument("--time-in-minutes", type=int, default=480)
-    parser.add_argument("-v", type=bool, default=False)
+    parser.add_argument('monitoring_master', type=str)
+    parser.add_argument('--time-in-minutes', type=int, default=480)
+    parser.add_argument('-v', action='store_true')
     args = parser.parse_args()
 
     max_minutes = args.time_in_minutes
@@ -67,8 +67,8 @@ def main():
 def nagios_create(hosts_locked, hosts_not_locked, max_minutes):
     nsca_output = ""
     for host in hosts_locked:
-        nsca_output += ("{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer"
-                        " than {}h {}m\x17"
+        nsca_output += ('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer'
+                        ' than {}h {}m\x17'
                         .format(host['hostname'], 1, int(max_minutes / 60),
                                 int(max_minutes - 60 * (max_minutes / 60))))
 
@@ -100,16 +100,16 @@ def console_out(hosts_locked, max_minutes):
     count_locked_servers = 0
     locked_hosts = ""
     for host in hosts_locked:
-        locked_hosts += "{}:{}\n".format(host['hostname'], host['igvm_locked'])
+        locked_hosts += '{}:{}\n'.format(host['hostname'], host['igvm_locked'])
         count_locked_servers += 1
 
     if locked_hosts == "":
-        print("No igvm_locked Servers !!!")
+        print('No igvm_locked Servers !!!')
     else:
         hours = int(max_minutes / 60)
         minutes = max_minutes - (int(max_minutes / 60) * 60)
 
-        print("({}) server/s are/is locked longer than {}h and {}m : \n{}"
+        print('({}) server/s are/is locked longer than {}h and {}m : \n{}'
               .format(count_locked_servers, hours, minutes, locked_hosts))
 
 
--- a/src/check_igvm_locked.py
+++ b/src/check_igvm_locked.py
@@ -1,0 +1,95 @@
+from adminapi.dataset import Query
+from adminapi.dataset.filters import Not, Any
+from datetime import datetime, timezone, timedelta
+from subprocess import Popen, PIPE, DEVNULL
+
+import argparse
+
+
+
+
+def main():
+
+    parser = argparse.ArgumentParser(description='nagios check for long running igvm_locked attribute')
+    parser.add_argument("monitoring_master", type=str)
+    parser.add_argument("--time_in_minutes", type=int, default=480)
+    args = parser.parse_args()
+
+    max_minutes = args.time_in_minutes
+    master = args.monitoring_master
+
+    hosts = Query({'servertype': Any('hypervisor', 'vm'), 'no_monitoring': False, 'state': Not('retired'), 'hostname': 'luiz.test.ig.local'}, ['igvm_locked', 'hostname'])
+
+    hosts_not_locked = []
+    hosts_locked = []
+    for host in hosts:
+        if not host['igvm_locked']:
+            hosts_not_locked.append(host)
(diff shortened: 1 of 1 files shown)


src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
+from adminapi.dataset import Query
```
Woellchen: copyright and documentation missing

Woellchen:
`flake8` currently reports 24 errors. please fix them

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+from datetime import datetime, timezone, timedelta
+from subprocess import Popen, PIPE, DEVNULL
+
+import argparse
+
+
+
+
+def main():
+
```
Woellchen: please remove useless newline

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+    parser = argparse.ArgumentParser(description='nagios check for long running igvm_locked attribute')
+    parser.add_argument("monitoring_master", type=str)
+    parser.add_argument("--time_in_minutes", type=int, default=480)
```
Woellchen: a nice cli flag would rather look like `--time-in-minutes`

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+    nsca_output = ""
+    for host in hosts_locked:
+        nsca_output += ('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer than {}h {}m\x17'
+            .format(host['hostname'],1,int(max_minutes/60),int(max_minutes - 60 * (max_minutes/60))))
+
+    for host in hosts_not_locked:
+        nsca_output += ('{}\tigvm_locked\t{}\tOK\x17'
+            .format(host['hostname'],0))
+
+    return(nsca_output)
```
Woellchen: remove useless parenthesis please

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+    result = True
+    nsca = Popen(
+        [
+            '/usr/sbin/send_nsca',
+            '-H', host,
+            '-c', '/etc/send_nsca.cfg',
+        ],
+        stdin=PIPE,
+        stdout=DEVNULL,
+        stderr=DEVNULL
```
Woellchen: lets have trailing commas. that way it is consistent and doesn't bloat the diff next time we change something here

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    nsca.communicate(nsca_output.encode())
+    returncode = nsca.wait()
+    if returncode > 0:
+        result = False
+
+    return result
```
Woellchen: how about `return not bool(returncode)`

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def console_out(hosts_locked, max_minutes):
+    count_locked_servers = 0
+    locked_hosts = ""
+    for host in hosts_locked:
+        #locked_hosts = str(locked_hosts) + str(host['hostname'] + ":" + str(host['igvm_locked']) + "\n")
```
Woellchen: please remove dead/commented code if it is not needed

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+        locked_hosts += "{}:{}\n".format(host['hostname'], host['igvm_locked'])
+        count_locked_servers += 1
+
+    if locked_hosts == "":
+       print("No igvm_locked Servers !!!")
+    else:
+       #print("This servers are locked longer than " + str(int(max_minutes/60)) + "h and " + str(max_minutes - (int(max_minutes/60) * 60)) + "m :" + str(locked_hosts))
```
Woellchen: please remove dead/commented code if it is not needed

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+    hosts_locked = []
+    for host in hosts:
+        if not host['igvm_locked']:
+            hosts_not_locked.append(host)
+        elif (datetime.now(timezone.utc) - host['igvm_locked']) > timedelta(minutes=max_minutes):
+            hosts_locked.append(host)
+        else:
+            hosts_not_locked.append(host)
+
+    console_out(hosts_locked, max_minutes)
```
Woellchen: is this a pure nsca check? why do we need to print something to the console then? and if we need to it should be the exact same text that is sent to nsca (`nagios_create`)

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+    nsca_out = nagios_create(hosts_locked, hosts_not_locked, max_minutes)
+    nagios_send(master, nsca_out)
+
+def nagios_create(hosts_locked, hosts_not_locked, max_minutes):
+    nsca_output = ""
+    for host in hosts_locked:
+        nsca_output += ('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer than {}h {}m\x17'
```
Woellchen: i am not sure if this is an acceptable nsca output string. maybe @lamaral can enlighten us

lamaral: Weird as it looks, NSCA uses it like that. 
```
Service Checks:
<host_name>[tab]<svc_description>[tab]<return_code>[tab]<plugin_output>[newline]

Host Checks:
<host_name>[tab]<return_code>[tab]<plugin_output>[newline]

When submitting multiple simultaneous results, separate each set with the ETB
character (^W or 0x17)
```

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+
+    parser = argparse.ArgumentParser(description='nagios check for long running igvm_locked attribute')
+    parser.add_argument("monitoring_master", type=str)
+    parser.add_argument("--time_in_minutes", type=int, default=480)
+    args = parser.parse_args()
+
+    max_minutes = args.time_in_minutes
+    master = args.monitoring_master
+
+    hosts = Query({'servertype': Any('hypervisor', 'vm'), 'no_monitoring': False, 'state': Not('retired'), 'hostname': 'luiz.test.ig.local'}, ['igvm_locked', 'hostname'])
```
lamaral: Please adjust this query, it still contains test data.

src/check_igvm_locked.py
```diff
@@ -0,0 +1,95 @@
...
+    if locked_hosts == "":
+       print("No igvm_locked Servers !!!")
+    else:
+       #print("This servers are locked longer than " + str(int(max_minutes/60)) + "h and " + str(max_minutes - (int(max_minutes/60) * 60)) + "m :" + str(locked_hosts))
+        hours = int(max_minutes/60)
+        minutes = max_minutes - (int(max_minutes/60) * 60)
+
+
+
+        print("({}) server/s are/is locked longer than {}h and {}m : \n{}".format(count_locked_servers, hours, minutes, locked_hosts))
```
lamaral: This line is too long. It can be broken down into multiple lines.

src/check_igvm_locked.py
lamaral: We could maybe toggle the output with a `-v` flag?

src/check_igvm_locked.py
Woellchen: but now it looks like this:
> (1) server/s are/is locked longer than 1h and 0m :
>***:####-##-## ##:##:##+##:##
>
>***	igvm_locked	1	WARNING - IGVM-locked longer than bla
is this still correct?!

Woellchen: ok scratch that. i had tested it locally and it was just because there were other prints to the console

Jowe1999:
changed the requested comments

src/check_igvm_locked.py
```diff
@@ -0,0 +1,117 @@
...
+import argparse
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='nagios check for long running igvm_locked attribute',
+    )
+    parser.add_argument("monitoring_master", type=str)
+    parser.add_argument("--time-in-minutes", type=int, default=480)
+    parser.add_argument("-v")
```
lamaral: This should be a boolean argument. The way it is, you need to give data to the argument.

src/check_igvm_locked.py
```diff
@@ -0,0 +1,117 @@
...
+import argparse
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='nagios check for long running igvm_locked attribute',
+    )
+    parser.add_argument("monitoring_master", type=str)
+    parser.add_argument("--time-in-minutes", type=int, default=480)
+    parser.add_argument("-v", type=bool, default=False)
```
lamaral: This still requires you to do something like `python3 check.py --time-in-minutes 60 -v True localhost`.

Try it like this:
```
parser.add_argument('-v', action='store_true')
```

Jowe1999: okey thx

Woellchen:
it is fine now

PR #172, opened 2019-09-13
Treat the ConnectionRefusedError exception on check_https_domains

lamaral:
In cases where a LBPool had no alive nodes behind it, this check would
return an exception because the LBPool would refuse the connection.
This PR fixes that problem by capturing the exception and returning
a proper message as well as an exit code.

### DIFF
--- a/src/check_https_domains.py
+++ b/src/check_https_domains.py
@@ -80,8 +80,8 @@ def main():
     try:
         state, output = get_check_result(domains, args.ip)
     except ConnectionRefusedError:
-        output = 'WARNING - The host refused the connection'
-        state = 2
+        output = 'CRITICAL - The host refused the connection'
+        state = 1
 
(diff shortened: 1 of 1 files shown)


src/check_https_domains.py
```diff
@@ -77,7 +77,12 @@ def main():
         print(message)
         sys.exit(3)
 
-    state, output = get_check_result(domains, args.ip)
+    try:
+        state, output = get_check_result(domains, args.ip)
+    except ConnectionRefusedError:
+        output = 'WARNING - The host refused the connection'
```
Woellchen: shouldn't this be critical then?

lamaral: I picked a warning because I judged it as something that didn't require immediate action from the responsible admin. 
Do you think it would be a good idea to make it a critical?

PR #173, opened 2019-09-13
Fix the systemd timer interval check

Woellchen:
this fixes the timer interval check by having an actual look at the unit it was configured to start and taking the last state change timestamp into account. it was previously ignored

### DIFF
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -15,17 +15,16 @@
 # Copyright (c) 2018 InnoGames GmbH
 #
 
+import logging
 from argparse import ArgumentParser
-from sys import exit
-
 from datetime import datetime
+from sys import exit
+from time import time
+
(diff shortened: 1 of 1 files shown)


src/check_systemd_units.py
```diff
@@ -196,63 +207,83 @@ def _check_timer(self, timer_warn, timer_crit):
...
 
-        # We can check only monotonic triggers for regular execution
-        min_interval = min(intervals) / m
-        inactivity = (
-            now - self.type_properties.LastTriggerUSec / m
-        )
-        last_execute = datetime.fromtimestamp(
-            self.type_properties.LastTriggerUSec / m
-        )
+        if len(intervals) > 1:
```
Felixoid: We still could have few triggers

Woellchen: you are right. i will put this in a loop

PR #175, opened 2019-09-25
Fix broken behavior on igvm_locked and use logging

lamaral:


src/check_igvm_locked.py
```diff
@@ -48,33 +69,53 @@ def main():
...
+            nsca_output = result
+        else:
+            nsca_output += result
+
+    ret = nagios_send(master, nsca_output)
+    if not ret:
+        exit_code = 1
+
+    if exit_code:
+        logger.warning('Failed to submit NSCA results')
```
Woellchen: error would fit better no?

src/check_igvm_locked.py
```diff
@@ -22,23 +22,44 @@
...
+    parser.add_argument(
+        '-v', action='store_true', dest='verbose',
+        help='Run the check in verbose mode'
+    )
     args = parser.parse_args()
 
+    if args.verbose:
+        logger.setLevel(logging.DEBUG)
+    else:
+        logger.setLevel(logging.FATAL)
```
Woellchen: usually we also want to see errors

src/check_igvm_locked.py
lamaral: I knew I was forgetting something :D

src/check_igvm_locked.py
```diff
@@ -48,33 +69,53 @@ def main():
...
+    for result in results:
+        if (len(nsca_output) + len(result)) >= 5000:
+            ret = nagios_send(master, nsca_output)
+            if not ret:
+                exit_code = 1
+            nsca_output = result
+        else:
+            nsca_output += result
+
+    ret = nagios_send(master, nsca_output)
```
TheFRedFox: After line 94 `exit_code could already be 1 .. is it necessary to do `nagios_send` again?

src/check_igvm_locked.py
```diff
@@ -97,20 +138,23 @@ def nagios_send(host, nsca_output):
 
 
 def console_out(hosts_locked, max_minutes):
-    count_locked_servers = 0
-    locked_hosts = ""
+    locked_hosts = []
     for host in hosts_locked:
```
TheFRedFox: Can be changed to:
``` python
locked_hosts = ['{}:{}'.format(host['hostname'], host['igvm_locked']) for host in hosts_locked]
```

src/check_igvm_locked.py
```diff
@@ -97,20 +138,23 @@ def nagios_send(host, nsca_output):
...
-        locked_hosts += '{}:{}\n'.format(host['hostname'], host['igvm_locked'])
-        count_locked_servers += 1
+        locked_hosts.append(
+            '{}:{}'.format(host['hostname'], host['igvm_locked'])
+        )
 
-    if locked_hosts == "":
-        print('No igvm_locked Servers !!!')
+    if not locked_hosts:
+        logger.info('No servers locked for more than 24 hours')
```
TheFRedFox: Return early here and get rid of the `else` indention.

src/check_igvm_locked.py
```diff
@@ -48,33 +69,53 @@ def main():
...
+def nagios_create(hosts_locked, hosts_not_locked, max_minutes):
+    nsca_output = []
     for host in hosts_locked:
-        nsca_output += ('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer'
-                        ' than {}h {}m\x17'
-                        .format(host['hostname'], 1, int(max_minutes / 60),
-                                int(max_minutes - 60 * (max_minutes / 60))))
+        nsca_output.append('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer'
+                           ' than {}h {}m\x17'
+                           .format(host['hostname'], 1, int(max_minutes / 60),
```
TheFRedFox: `int(max_minutes / 60)` and `int(max_minutes - 60 * (max_minutes / 60))` are used in the loop and are used in another method below. As `max_minutes` is not getting changed during the script runtime, you should rather calculate those things more globally.

src/check_igvm_locked.py
```diff
@@ -48,33 +69,53 @@ def main():
...
+        exit_code = 1
+
+    if exit_code:
+        logger.warning('Failed to submit NSCA results')
+    exit(exit_code)
+
+
+def nagios_create(hosts_locked, hosts_not_locked, max_minutes):
+    nsca_output = []
     for host in hosts_locked:
```
TheFRedFox: Can be changed in the same way like: https://github.com/innogames/igmonplugins/pull/175/files#r328110912

src/check_igvm_locked.py
```diff
@@ -48,33 +69,53 @@ def main():
...
-        nsca_output += ('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer'
-                        ' than {}h {}m\x17'
-                        .format(host['hostname'], 1, int(max_minutes / 60),
-                                int(max_minutes - 60 * (max_minutes / 60))))
+        nsca_output.append('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer'
+                           ' than {}h {}m\x17'
+                           .format(host['hostname'], 1, int(max_minutes / 60),
+                                   int(max_minutes - 60 * (max_minutes / 60))))
 
     for host in hosts_not_locked:
```
TheFRedFox: Can be changed in the same way like: https://github.com/innogames/igmonplugins/pull/175/files#r328110912

src/check_igvm_locked.py
lamaral: We have to break down the output to nsca in chunks smaller than 5000 characters. 
This last `nagios_send` should accommodate for the last chunk of data that didn't get sent in the loop.

src/check_igvm_locked.py
```diff
@@ -48,33 +69,53 @@ def main():
...
 
+    console_out(hosts_locked, max_minutes)
+
+    results = nagios_create(hosts_locked, hosts_not_locked, max_minutes)
 
-def nagios_create(hosts_locked, hosts_not_locked, max_minutes):
     nsca_output = ""
+    exit_code = 0
+    for result in results:
+        if (len(nsca_output) + len(result)) >= 5000:
```
TheFRedFox: Somehow that looks like we rather want to achieve something different.
For what exactly are you doing the check for. What do you want to achieve with it?

src/check_igvm_locked.py
lamaral: nsca_output is being built from two different lists with different logic for each. 

Making it a list comprehension would be nice, but I think would make it harder to understand what is going on.

src/check_igvm_locked.py
lamaral: nsca_output is being built from two different lists with different logic for each.

Making it a list comprehension would be nice, but I think would make it harder to understand what is going on.

src/check_igvm_locked.py
lamaral: The send_nsca has a limit on it's stdin buffer of roughly 5100 bytes. 
If we don't break it down in chunks, we end up with a string with around 300k characters, and some of the data simply gets silently dropped by send_nsca.

src/check_igvm_locked.py
TheFRedFox: Ah yeah, right. I dunno if performance would be actually better, but one could try to first make the two lists with list-comprehensions and then put them together to one big list.

src/check_igvm_locked.py
```diff
@@ -116,16 +116,21 @@ def main():
...
 def nagios_create(hosts_locked, hosts_not_locked, max_minutes, max_hours):
-    template = ('{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer'
-                'than {}h {}m\x17')
-    out_locked = [template.format(host['hostname'], 1, max_hours, max_minutes)
-                  for host in hosts_locked]
+    template = (
+        '{}\tigvm_locked\t{}\tWARNING - IGVM-locked longer than {}h {}m\x17'
+    )
+    out_locked = [
+        template.format(
```
TheFRedFox: Go for
``` python
out_locked = [
    template.format(host['hostname'], 1, max_hours, max_minutes)
    for host in hosts_locked
]
```

src/check_igvm_locked.py
```diff
@@ -116,16 +116,21 @@ def main():
...
+            host['hostname'], 1, max_hours, max_minutes
+        ) for host in hosts_locked
+    ]
 
     template = '{}\tigvm_locked\t{}\tOK\x17'
-    out_not_locked = [template.format(host['hostname'], 0)
-                      for host in hosts_not_locked]
+    out_not_locked = [
+        template.format(host['hostname'], 0) for host in hosts_not_locked
```
TheFRedFox: Go for 
``` python
out_not_locked = [
    template.format(host['hostname'], 0)
    for host in hosts_not_locked
]

File: src/check_igvm_locked.py [hunk 12]
```diff
@@ -101,6 +104,8 @@ def main():
         else:
             nsca_output += result
 
+    # This last nagios send, cover the remaining output that wasn't sent
```
TheFRedFox:
```
This last nagios sending covers the ...
```

src/check_igvm_locked.py
```diff
@@ -22,61 +22,115 @@
...
 
-import argparse
+logging.basicConfig(
+    format='%(levelname)-8s [%(filename)s:%(lineno)d] %(message)s'
+)
+logger = logging.getLogger(__name__)
 
 
 def main():
     parser = argparse.ArgumentParser(
```
TheFRedFox: If you want, you can separate the `ArgumentParser` part to an own method like all the other scripts. But it would be not in that PR scope.

PR #176, opened 2019-09-27
Add support for newer Extreme and HP switches

TuxPowered42:


src/check_switch.py
```diff
@@ -258,26 +258,21 @@ def check_ports(snmp, model, args):
                 continue
 
         elif model == 'powerconnect':
-            if (
-                port_names[port_index].startswith('Vl') or
-                port_names[port_index].startswith('CPU') or
-                port_names[port_index].startswith('Po')
+            if re.match(
+                '^(CPU|(Vl|Po)[0-9]+)$',
```
TheFRedFox: What's the reason for putting e.g. `Vl|Po` again in own braces? Same for the other regexes.

TheFRedFox: And what's the reason for not anymore checking for just the start but for the whole string?

TheFRedFox: For better performance (if needed) you can use `re.compile` before the `for-loop` and match with the compiled one then.

TuxPowered42: > What's the reason for putting e.g. `Vl|Po` again in own braces? Same for the other regexes.

No idea, this might be wrong. I will check.

> And what's the reason for not anymore checking for just the start but for the whole string?

Because that is the correct thing to do. It's a special interface of this particular name and with regexp I can correctly verify it and not risk that somebody created a normal interface of similar name.

> For better performance (if needed) you can use `re.compile` before the `for-loop` and match with the compiled one then.

Sure, I can do it.

TuxPowered42: > What's the reason for putting e.g. `Vl|Po` again in own braces? Same for the other regexes.

It is correct after all. Interface can be "CPU" or "(Vl|Po) with a number".

TheFRedFox: > > What's the reason for putting e.g. `Vl|Po` again in own braces? Same for the other regexes.
> 
> It is correct after all. Interface can be "CPU" or "(Vl|Po) with a number".

Ah yeah, that makes sense. Alright, keep it.

src/check_switch.py
```diff
@@ -67,6 +67,14 @@
     'force10_mxl': '1.3.6.1.4.1.6027.3.26.1.4.4.1.4.2.1.1',
 }
 
+PORT_SKIPS = {
```
TheFRedFox: Nice first approach to reduce the complexity of code by adding configuration. 👍

PR #177, opened 2019-10-22
Allow Auth and SSL for Elastic checks

elmatze79:


src/check_es_cluster_state.py
```diff
@@ -37,13 +37,52 @@ def parse_args():
...
+    )
 
     return parser.parse_args()
 
 def main():
     args = parse_args()
-    es = Elasticsearch(args.host)
+    if args.root_ca:
+        scheme = 'https'
+        from ssl import create_default_context
```
TheFRedFox: What was the reason to put the `from import` not on top of the code but here inside the code?

src/check_es_cluster_state.py
```diff
@@ -37,13 +37,52 @@ def parse_args():
...
+        context = create_default_context(cafile = args.root_ca)
+    else:
+        scheme = 'http'
+    if args.user and args.password:
```
TheFRedFox: Take the advantage of python and named parameters:
``` python
params = {}
if args.root_ca:
    params['scheme'] = 'https'
    params['context'] = create_default_context(...)

if args.user and args.password:  # is the `and` actually correctly here? what if user or password is actually an empty string? ^^
    params['http_auth'] = (args.user, args.password)

es = Elasticsearch([args.host], **params)
```

src/check_es_index_exists.py
```diff
@@ -42,12 +42,49 @@ def parse_args():
...
 def main():
     args = parse_args()
-    es = Elasticsearch(args.host)
+    if args.root_ca:
+        scheme = 'https'
+        from ssl import create_default_context
+        context = create_default_context(cafile = args.root_ca)
+    else:
+        scheme = 'http'
+    if args.user and args.password:
```
TheFRedFox: See above

src/check_es_index_lifecycle.py
```diff
@@ -40,13 +40,51 @@ def parse_args():
...
     failed_indices = []
     args = parse_args()
-    es = Elasticsearch(args.host)
+    if args.root_ca:
+        scheme = 'https'
+        from ssl import create_default_context
+        context = create_default_context(cafile = args.root_ca)
+    else:
+        scheme = 'http'
+    if args.user and args.password:
```
TheFRedFox: See above

src/check_es_cluster_state.py
elmatze79: because it's only needed for ssl

PR #178, opened 2019-12-11
Check for unused storage volumes

TuxPowered42:


src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python
```
Woellchen: what happened to python3? :(

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,115 @@
...
+    code = ExitCodes.ok
+    reason = 'All volumes are used by VMs'
+    
+    domain_names = [d.name() for d in domains]
+    mounts = [resolve_symlink(m.device) for m in psutil.disk_partitions()]
+
+    bad_volumes = []
+    for volume in volumes:
+        
+        if volume == 'test-vol':
```
Woellchen: is this internal specifics? if yes we should make this a parameter. otherwise it should be a constant

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,115 @@
...
+    return code, reason, bad_volumes
+
+
+def resolve_symlink(path):
+    if os.path.islink(path):
+        return os.path.realpath(os.readlink(path))
+    return path
+
+
+def print_volumes(bad_volumes):
```
Woellchen: i would vote for `print_bad_volumes`

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,115 @@
...
+        conn = openReadOnly(None)
+    except libvirtError as error:
+        print_nagios_message(
+            ExitCodes.unknown,
+            'Could not connect to libvirt: {0}'.format(str(error))
+        )
+        exit(ExitCodes.unknown)
+
+    domains = conn.listAllDomains()
+    pool = conn.storagePoolLookupByName('xen-data')
```
Woellchen: `xen-data` should definitely be a parameter

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def resolve_symlink(path):
+    if os.path.islink(path):
+        return os.path.realpath(os.readlink(path))
+    return path
+
+
+def print_bad_volumes(bad_volumes):
+    for bad_volume in bad_volumes:
```
TheFRedFox:
``` python
for path, mounted in bad_volumes:
    ...
```

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+        code = ExitCodes.warning
+        reason = 'Found volumes not corresponding to defined VMs'
+        bad_volumes.append((volume.path(), mounted))
+
+    return code, reason, bad_volumes
+
+
+def resolve_symlink(path):
+    if os.path.islink(path):
+        return os.path.realpath(os.readlink(path))
```
TheFRedFox: What would be the worst case when between the execution of the `if` and the `readlink` the link will get changed or removed?

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+from libvirt import openReadOnly, libvirtError
+from sys import exit
+
+import libvirt
+import psutil
+import re
+import os
+
+
+class ExitCodes:
```
TheFRedFox: All helper classes and methods to the very end. In the total order of:
* parse_args
* main
* helper_methods_directly_used_by_main
* helper_methods_used_more_internally
* helper_classes

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+    print("{0} - {1}".format(state_text, reason))
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument(
+        'pool', help='Storage pool', type=str,
+    )
+    parser.add_argument(
+        '-ignore', help='Ignore volumes by regexp', type=str, default=None
```
TheFRedFox: Since when do we use single dash for full written arguments? Shouldn't be rather `--ignore` than `-ignore`?

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+    code = ExitCodes.ok
+    reason = 'All volumes are used by VMs'
+    
+    domain_names = [d.name() for d in domains]
+    mounts = [resolve_symlink(m.device) for m in psutil.disk_partitions()]
+
+    bad_volumes = []
+    for volume in volumes:
+        
+        if ignore and re.match(
```
TheFRedFox: Compile the regex before the `for-loop`.

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return parser.parse_args()
+
+
+def check(domains, volumes, ignore):
+    code = ExitCodes.ok
+    reason = 'All volumes are used by VMs'
+    
+    domain_names = [d.name() for d in domains]
+    mounts = [resolve_symlink(m.device) for m in psutil.disk_partitions()]
```
TheFRedFox: I guess we can use `set`s here, can't we? The `in` operators would work much faster with `set`s. The order of the domains and disk_partitions do not matter, right?

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+
+    bad_volumes = []
+    for volume in volumes:
+        
+        if ignore and re.match(
+            ignore,
+            os.path.basename(volume.path())
```
TheFRedFox: Don't let us recall methods too often. There are calls for `volume.path()` 4 times and for `os.path.basename` 2 times. At least `volume.path()` can be first stored in a variable.

src/check_libvirt_volumes.py
TuxPowered42: Then the program will crash. I'm fine with that. Those symlinks are managed by libvirt and we can't lock on it's operations.

src/check_libvirt_volumes.py
TuxPowered42: `volume.path()` is now stored in additional variable.

src/check_libvirt_volumes.py
TuxPowered42: Fixed.

src/check_libvirt_volumes.py
TheFRedFox: Then it is good to go. Just general hint: To be more secure, you can just try to do stuff and then catch the Error. E.g. `os.readlink` will raise a `FileNotFoundError` if the file/directory does not exist and a `OSError` if the file is not a link. With that approach you don't fall into Race Condition Problems.

src/check_libvirt_volumes.py
```diff
@@ -0,0 +1,134 @@
...
+    print_bad_volumes(bad_volumes)
+    exit(code)
+
+
+def check(domains, volumes, ignore):
+    code = ExitCodes.ok
+    reason = 'All volumes are used by VMs'
+    
+    domain_names = set([d.name() for d in domains])
+    mounts = set([resolve_symlink(m.device) for m in psutil.disk_partitions()])
```
TheFRedFox: There is also `set-comprehension`:
``` python
mounts = {resolve_symlink(..) ...}
```

src/check_libvirt_volumes.py
TheFRedFox: `ExitCodes` is a helper class. So, let us move it to the very bottom. Just before the `__name__` statement.

src/check_libvirt_volumes.py
TuxPowered42: After giving it some thought I came to conclusion that indeed it is better to not crash. We can ignore errors for just a single storage volume and still correctly check all others.

PR #179, opened 2019-12-16
feat(memory): Add free memory nagios check

meikomeik:
This script will read /proc/meminfo and calculate the sum of:                   
* MemFree                                                                       
* SwapFree                                                                      
* Cached                                                                        
                                                                                
The sum will be compared agains the size of the swap partition.                 
                                                                                
Warning: if the free memory is lower than 75% of swap size                      
Critical: if the free memory is lower than 50% of swap size

### DIFF
--- a/src/check_free_memory_linux.py
+++ b/src/check_free_memory_linux.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python
+"""InnoGames Monitoring Plugins - Linux mem free check
+
+This script will read /proc/meminfo and calculate the sum of:
+* MemFree
+* SwapFree
+* Cached
+
+Will raise a warning/critical if a certain percentage or absolute value
+is reached.
+
+Copyright (c) 2020 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import argparse
+import os
+import sys
+
+
+def parse_args():
+    """Parse command line arguments"""
+    parser = argparse.ArgumentParser(
+        description=(
+            'Calculate the sum of MemFree, SwapFree and Cached. '
+            'Raise a warning or critical if the threshold is undercut.'
+        ),
+    )
+    # Our systems usually run with 1 GiB of swap.
+    # Raise a warning if the amount of free memory drops below 768 MiB
+    parser.add_argument(
+        '-w',
+        '--warning',
+        metavar='warning',
+        type=str,
+        default=str(768*1024*1024),  # 768 MiB
+        help=(
+            'Expects an absolute value in bytes or a percentage of '
+            'total memory (suffixed with %%)! '
+            'Raise a warning if this limit is exceeded. '
+        ),
+    )
+    # Raise a critical error if amount of free memory drops below 512 MiB
+    parser.add_argument(
+        '-c',
+        '--critical',
+        metavar='critical',
+        type=str,
+        default=str(512*1024*1024),  # 512 MiB
+        help=(
+            'Expects an absolute value in bytes or a percentage of '
+            'total memory (suffixed with %%)! '
+            'Raise a warning if this limit is exceeded.'
+        ),
+    )
+    args = parser.parse_args()
+
+    args.warning_percent = args.warning.endswith('%')
+    args.critical_percent = args.critical.endswith('%')
+    args.warning = int(args.warning.rstrip('%'))
+    args.critical = int(args.critical.rstrip('%'))
+    return args
+
+
+def main():
+    args = parse_args()
+
+    meminfo = read_proc_meminfo()
+    if not meminfo:
+        print('Could not parse /proc/meminfo')
+        sys.exit(3)
+
+    mem_free_kib = meminfo['MemFree'] + meminfo['SwapFree'] + meminfo['Cached']
+    mem_total_kib = meminfo['MemTotal'] + meminfo['SwapTotal']
+
+    output = (
+        'Memory low! Please increase memory! (MemFree + SwapFree + Cached) '
+        '< {:,} MiB'.format(mem_free_kib >> 10)
+    )
+    exit_status = 3
+
+    crit_value = (mem_free_kib << 10) / args.critical
+    crit_factor = (mem_total_kib << 10) / 100 if args.critical_percent else 1
+    warn_value = (mem_free_kib << 10) / args.warning
+    warn_factor = (mem_total_kib << 10) / 100 if args.warning_percent else 1
+    if crit_value < crit_factor:
+        exit_status = 2
+    elif warn_value < warn_factor:
+        exit_status = 1
+    else:
+        output = 'There is enough free memory :aw_yeah:'
+        exit_status = 0
+
+    print(output)
+    sys.exit(exit_status)
+
+
+def read_proc_meminfo():
+    """Return /proc/meminfo as a dict"""
+    meminfo_dict = {}
+    try:
+        with open('/proc/meminfo', 'r') as proc_file:
+            for line in proc_file.readlines():
+                name, value = line.strip().replace(' kB', '').split(':', 1)
+                meminfo_dict[name] = int(value)
+            return meminfo_dict
+    except (OSError, IOError):
+        return None
+
+
+if __name__ == '__main__':
+    main()


src/check_free_memory.py
```diff
@@ -0,0 +1,116 @@
...
+        return None
+
+
+def main():
+    args = parse_args()
+
+    meminfo = read_proc_meminfo()
+    if meminfo:
+        mem_free = meminfo['MemFree'] + meminfo['SwapFree'] + meminfo['Cached']
+        swap_total = meminfo['SwapTotal']
```
Felixoid: What will be with the disabled swap? I guess, the idea is very useful, but the swap size isn't the best choice for sticking to

src/check_free_memory.py
```diff
@@ -0,0 +1,116 @@
...
+        default=50,
+        help=(
+            'Percentage of the limit which may be reached until a critical is '
+            'thrown.'
+        ),
+    )
+    args = parser.parse_args()
+
+    if os.getuid() != 0:
+        raise Exception('I need to be run as root, really')
```
TuxPowered42: Really, though? Also this has nothing to do with parsing arguments.

src/check_free_memory.py
```diff
@@ -0,0 +1,116 @@
...
+
+def main():
+    args = parse_args()
+
+    meminfo = read_proc_meminfo()
+    if meminfo:
+        mem_free = meminfo['MemFree'] + meminfo['SwapFree'] + meminfo['Cached']
+        swap_total = meminfo['SwapTotal']
+
+    output = ' '.join([
```
TuxPowered42: This is not Puppet, there are better ways to have a long string split across multiple lines.

src/check_free_memory.py
```diff
@@ -0,0 +1,116 @@
...
+        'Low on memory. Please increase memory!',
+        '(MemFree + SwapFree + Cached) == {} kB'.format(format(mem_free, ',d'))
+    ])
+    if mem_free < swap_total * args.critical / 100:
```
TuxPowered42: Those variables are assigned only if meminfo was read. So this script will crash when it can't be read. In such case UNKNOWN Nagios status should be returned.

src/check_free_memory.py
```diff
@@ -0,0 +1,116 @@
...
+    args = parse_args()
+
+    meminfo = read_proc_meminfo()
+    if meminfo:
+        mem_free = meminfo['MemFree'] + meminfo['SwapFree'] + meminfo['Cached']
+        swap_total = meminfo['SwapTotal']
+
+    output = ' '.join([
+        'Low on memory. Please increase memory!',
+        '(MemFree + SwapFree + Cached) == {} kB'.format(format(mem_free, ',d'))
```
TuxPowered42: Please print everything with binary prefixes, as we always do at IG.

src/check_free_memory.py
meikomeik: I agree. I am rewriting the script and make better use of the ig::server variables

src/check_free_memory.py
meikomeik: I will move that block to a different place

src/check_free_memory.py
meikomeik: I moved it to the main class

src/check_free_memory.py
meikomeik: I deleted the code

src/check_free_memory.py
meikomeik: please review again

src/check_free_memory.py
meikomeik: sorry, code changed completly. comment no longer valid

src/check_free_memory.py
meikomeik: I converted all strings to byte-like objects

src/check_free_memory.py
meikomeik: please review multi line stuff

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+    mem_total_kB = meminfo['MemTotal'] + meminfo['SwapTotal']
+
+    output = (
+        'Memory low! Please increase memory! (MemFree + SwapFree + Cached) '
+        '< {} kB'.format(format(mem_free_kB, ',d'))
+    ).encode()
+    exit_status = 3
+
+    if args.critical_percent:
+        if mem_free_kB < mem_total_kB * args.critical / 100:
```
Felixoid: `mem_free_kB / args.critical` would be a variable, as well as warning

```
crit_value = (mem_free_kB << 10) / args.critical
crit_factor = (mem_total_kB << 10) / 100 if args.critical_percent else 1
if crit_value < crit_factor:
    exit_status = 2
else if warn_value < varn_factor:
    exit_status = 1
else:
    output = b'There is enough free memory :aw_yeah:'
    exit_status = 0
```

Not sure if it's more readable, but I prefer less nested if

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+        type=str,
+        default=524288000,
+        help=(
+            'Expects an absolute value in bytes or a percentage of total memory (suffixed with %)! '
+            'Raise a critical if this limit is exceeded.'
+        ),
+    )
+    args = parser.parse_args()
+
+    args.warning_percent = True if args.warning.endswith('%') else False
```
Felixoid: it returns True or False without ternary

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+    args.critical_percent = True if args.critical.endswith('%') else False
+    args.warning = int(args.warning.replace('%', ''))
```
Felixoid: `str.rstrip('%')`?

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+import sys
+
+
+def parse_args():
+    """The main program"""
+    parser = argparse.ArgumentParser(
+        description=(
+            'Check all running processes for the nofile limit, will throw '
+            'a warning if the limit is nearly reached and critical, if '
+            'the limit is reached'
```
kofrezo: Copy & paste error ? This is not the description of the check is it ?

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+        ),
+    )
+    parser.add_argument(
+        '-w',
+        '--warning',
+        metavar='warning',
+        type=str,
```
kofrezo: You could make it a float or integer and if it is above 100 see it as bytes value and otherwise as string. I think it is unlikely we specify values lower then 100 bytes.

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+        default=786432000,
```
kofrezo: - If you prefer to use string and suffixes why not support 1K, 1M, 1G ?
- Where does this value of 750 MB come from ?

src/check_free_memory.py
kofrezo: This is not a ternary operator but he is right. No need for if condition.

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+    args.warning_percent = True if args.warning.endswith('%') else False
+    args.critical_percent = True if args.critical.endswith('%') else False
+    args.warning = int(args.warning.replace('%', ''))
+    args.critical = int(args.critical.replace('%', ''))
+    return args
+
+
+def read_proc_meminfo():
+    """Return /proc/meminfo as a dict
+    """
```
kofrezo: No need for extra line

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+        return None
+
+
+def main():
+    args = parse_args()
+
+    meminfo = read_proc_meminfo()
+    if not meminfo:
+        print('Could not parse /proc/meminfo')
+        exit(3)
```
kofrezo: You may want to use sys.exit (https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python)

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+
+    mem_free_kB = meminfo['MemFree'] + meminfo['SwapFree'] + meminfo['Cached']
+    mem_total_kB = meminfo['MemTotal'] + meminfo['SwapTotal']
+
+    output = (
+        'Memory low! Please increase memory! (MemFree + SwapFree + Cached) '
+        '< {} kB'.format(format(mem_free_kB, ',d'))
```
kofrezo: No need to use format for digit formatting you cna use format syntax (https://pyformat.info/#number)

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+    ).encode()
+    exit_status = 3
+
+    if args.critical_percent:
+        if mem_free_kB < mem_total_kB * args.critical / 100:
+            exit_status = 2
+    elif not args.critical_percent:
+        if mem_free_kB * 1024 < args.critical:
```
kofrezo: args.critical is still a string this should raise a TypeError

src/check_free_memory.py
kofrezo: args.critical is still a string this should raise a TypeError

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
...
+            exit_status = 2
+    elif args.warning_percent:
+        if mem_free_kB < mem_total_kB * args.warning / 100:
+            exit_status = 1
+    elif not args.warning_percent:
+        if mem_free_kB * 1024 < args.critical:
+            exit_status = 1
+    else:
+        output = b'There is enough free memory :aw_yeah:'
```
kofrezo: Why use bytes string here ?

src/check_free_memory.py
meikomeik: Agree: `args.warning_percent = args.warning.endswith('%')` is all we need

src/check_free_memory.py
meikomeik: Yes, copy&paste error. Documentation is hard

src/check_free_memory.py
meikomeik: Puppet will calculate these into bytes. I didn't want to do K, M, G calculations myself as puppet can do it for us.

src/check_free_memory.py
meikomeik: I don't feel happy with "If value < 100 it must be percent else bytes"
I'd rather keep the % as indicator for percentages

src/check_free_memory.py
meikomeik: agree

src/check_free_memory.py
meikomeik: Agree. I actually wanted to use sys.exit here! Thanks

src/check_free_memory.py
meikomeik: I convert it to int during argument parsing:
`args.critical = int(args.critical.replace('%', ''))`

src/check_free_memory.py
meikomeik: Agree, this is shorter.

src/check_free_memory.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
meikomeik: Renamed the script to `check_free_memory_linux.py`

src/check_free_memory.py
meikomeik: I convert it to int during argument parsing:
`args.critical = int(args.critical.replace('%', ''))`

src/check_free_memory.py
meikomeik: This change was requested by Kajetan or I missunderstood him maybe?!
Please print everything with binary prefixes, as we always do at IG.

_Originally posted by @TuxPowered42 in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjIxMDE5MDQxOnYy/pull_request_review_threads/discussion_

src/check_free_memory.py
kofrezo: Use '{:,}'

meikomeik: Beautiful!

src/check_free_memory.py
meikomeik: I now use `KiB` as a unit

src/check_free_memory.py
meikomeik: Removed the b from the string. Was a missunderstanding

src/check_free_memory.py
meikomeik: Implemented your way

src/check_free_memory_linux.py
```diff
@@ -0,0 +1,124 @@
...
+    args = parser.parse_args()
+
+    args.warning_percent = args.warning.endswith('%')
+    args.critical_percent = args.critical.endswith('%')
+    args.warning = int(args.warning.rstrip('%'))
+    args.critical = int(args.critical.rstrip('%'))
+    return args
+
+
+def read_proc_meminfo():
```
TheFRedFox: Move helper methods to the bottom.

src/check_free_memory_linux.py
```diff
@@ -0,0 +1,124 @@
...
+            'total memory (suffixed with %)! '
+            'Raise a warning if this limit is exceeded.'
+        ),
+    )
+    parser.add_argument(
+        '-c',
+        '--critical',
+        metavar='critical',
+        type=str,
+        default=524288000,
```
TheFRedFox: Explain those magic numbers in a comment or define a CONSTANT with a descriptive variable name.

src/check_free_memory_linux.py
```diff
@@ -0,0 +1,124 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import argparse
+import os
+import sys
+
+
+def parse_args():
+    """The main program"""
```
TheFRedFox: That's not the main program.

src/check_free_memory_linux.py
```diff
@@ -0,0 +1,124 @@
...
+            'Calculate the sum of MemFree, SwapFree and Cached. '
+            'Raise a warning or critical if the threshold is undercut.'
+        ),
+    )
+    parser.add_argument(
+        '-w',
+        '--warning',
+        metavar='warning',
+        type=str,
+        default=786432000,
```
TuxPowered42: This is not exact 768MiB, but rather 1000 * 768KiB.

src/check_free_memory_linux.py
TuxPowered42: This is not exact 768MiB, but rather 1000 * 768KiB.

src/check_free_memory_linux.py
meikomeik: Moved main() function to the top. Is this how you expect it?

src/check_free_memory_linux.py
meikomeik: Indeed. Fixed

src/check_free_memory_linux.py
meikomeik: Isn't it 750 MiB?

src/check_free_memory_linux.py
meikomeik: Should be 500 MiB

meikomeik: Added a comment to the numbers

src/check_free_memory_linux.py
meikomeik: Changed it to ` default=805306368,  # 768 MiB`

src/check_free_memory_linux.py
meikomeik: Changed it to `default=536870912,  # 512 MiB`

src/check_free_memory_linux.py
TheFRedFox: For a lot of scripts we have this structure:
* parse_args
* main
* helper_methods_directly_called_by_main
* helper_methods_called_by_helper_methods
* helper_classes_directly_used_by_main
* helper_classes_used_by_other

But having the main method at the very top is fair enough.

src/check_free_memory_linux.py
TheFRedFox: You got me wrong. The question is, why is it exactly this number? Why not any other number? Why was that one chosen? Or no particular reason?

meikomeik: No particular reason. I think these are good defaults as our servers usually run with 1GiB of swap so if the ammout of free memory drops below 768MiB I'd like to have a warning and below 512MiB I would like it to become critical.

meikomeik: Added two comments clarifying the chosen defaults

src/check_free_memory_linux.py
```diff
@@ -0,0 +1,128 @@
...
+        ),
+    )
+    # Our systems usually run with 1 GiB of swap.
+    # Raise a warning if the amount of free memory drops below 768 MiB
+    parser.add_argument(
+        '-w',
+        '--warning',
+        metavar='warning',
+        type=str,
+        default='805306368',  # 768 MiB
```
TuxPowered42: You could just write it as 768 * 1024 * 1024 or 768 << 10, that would be way more readible.

src/check_free_memory_linux.py
meikomeik: Moved parse_args to top followed by main class. Thanks for the explanation

src/check_free_memory_linux.py
meikomeik: Added the multiplication as well to the comment

PR #184, opened 2020-02-20
Nagios checks for ClickHouse

Felixoid:


src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+    )
+    parser_replication.add_argument(
+        '--absolute-delay', action='store', type=int, default=40,
+        help='how big lag in seconds the current replica could have',
+    )
+
+    return parser.parse_args()
+
+
+def main():
```
TheFRedFox: Python Structure

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def parse_args() -> Namespace:
+    """
+    Parse the arguments
+
+    We are returning them as a dict for callers convenience.
+    """
+    parser = ArgumentParser(formatter_class=ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
```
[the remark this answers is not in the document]
Felixoid: Very nice trick, I'll use it

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+# Nagios clickhouse-server check
+#
+# This check tests multiple clickhouse-server's health metrics
+#
+# Copyright (c) 2020 InnoGames GmbH
+#
+
+from argparse import ArgumentParser, Namespace, ArgumentDefaultsHelpFormatter
+# https://pypi.org/project/clickhouse-driver/
+from clickhouse_driver import Client  # type: ignore
```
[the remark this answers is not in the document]
Felixoid: `Cannot find implementation or library stub for module named 'clickhouse_driver'`

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+        self.current = code
+
+    def reset(self, code: int = OK):
+        """
+        Resets 'current' property to optionally given code, 0 by default
+        """
+        self._check_code(code)
+        self.__current = code
+
+    def _check_code(self, code: int):
```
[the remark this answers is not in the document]
Felixoid: I'd raise exception explicitly to not allow stupid actions here ¯_(ツ)_/¯

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+            self.code.current = 1
+            return self.exit(e.message)
+
+    def execute_dict(self, *args, **kwargs) -> list:
+        """
+        Wraper around execute() to return list of rows as dict
+        """
+        kwargs['with_column_types'] = True
+        rows, columns = self.execute(*args, **kwargs)
+        result = []  # type: list
```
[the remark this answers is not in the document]
Felixoid: Yes, you are right

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+                        'code is {}, message is begining with {}'.format(
+                            cl, t, e.code, e.message.split('\n')[0]
+                        )
+                    )
+                    # No reason to continue if table is unavailable on cluster
+                    break
+        messages = messages or ['All clusters are fine: {}'.format(clusters)]
+        self.exit('; '.join(messages))
+
+    def replication(self, args: Namespace):
```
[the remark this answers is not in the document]
Felixoid: It's in the documentation for Argparse, how to proceed with functions for subparsers.

https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers

src/check_clickhouse.py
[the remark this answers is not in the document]
Felixoid: That's very cool trick

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+        '--inserts-in-queue', action='store', type=int, default=10,
+        help='how many inserts of blocks are delayed',
+    )
+    parser_replication.add_argument(
+        '--merges-in-queue', action='store', type=int, default=40,
+        help='how many merges are delayed, could be greater then 0 for a long',
+    )
+    parser_replication.add_argument(
+        '--log-delay', action='store', type=int, default=10,
+        help='how big difference between actions log in '
```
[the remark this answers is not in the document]
Felixoid: I don't quite see what do you mean. They are in the same columns, I guess

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+def main():
+    args = parse_args()
+    logger.setLevel(args.log_level)
+    logger.debug('Arguments are {}'.format(pformat(args.__dict__)))
+    ch = Check(
+        host=args.host, port=args.port, database='system',
+        user=args.user, password=args.password,
+    )
+    # Each subcommand executes appropriate method of Check
+    args.func(ch, args)
```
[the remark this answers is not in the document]
Felixoid: Discussed in private, it's in the documentation `how to use different function for each subparser` https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers

src/check_clickhouse.py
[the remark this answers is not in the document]
Felixoid: The part of https://www.python.org/dev/peps/pep-0484

src/check_clickhouse.py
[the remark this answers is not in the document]
TheFRedFox: Right now it is:
``` python
help='...'
'...'
```
Aligned it would be:
``` python
help='...'
          '...'
```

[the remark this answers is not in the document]
Felixoid: But they are both valid pep8
```
[felixoid@ig-fel src]$ sed '70,74p;d' check_clickhouse.py
    parser_replication = subparsers.add_parser(
        'replication', formatter_class=ArgumentDefaultsHelpFormatter,
        help='check replication health, see https://clickhouse.tech/docs/en/'
             'operations/system_tables/#system_tables-replicas',
    )
[felixoid@ig-fel src]$ flake8 check_clickhouse.py
[felixoid@ig-fel src]$ sed '70,74p;d' check_clickhouse.py
    parser_replication = subparsers.add_parser(
        'replication', formatter_class=ArgumentDefaultsHelpFormatter,
        help='check replication health, see https://clickhouse.tech/docs/en/'
        'operations/system_tables/#system_tables-replicas',
    )
[felixoid@ig-fel src]$ flake8 check_clickhouse.py
[felixoid@ig-fel src]$ 
```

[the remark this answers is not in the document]
TheFRedFox: Well, maybe it is just me, but I find
``` python
help='...'
'...'
```
not nice to read and would definitely prefer the other style. I might can get overruled there though.

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+            return
+
+
+class Check(object):
+    @staticmethod
+    def sanitize_name(name: str) -> str:
+        """
+        Return only characters allowed in database/table name
+        """
+        return ''.join(c for c in name if (c.isalnum() or c in '_'))
```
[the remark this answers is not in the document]
Felixoid: At least now, this is unnecessary. I've removed this, since every table now comes from `system.tables` CH table

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+            if t['is_readonly'] or t['is_session_expired']:
+                self.code.current = Code.CRITICAL
+                messages.append(
+                    'Table `{}`: is_readonly={}, is_session_expired={}'.format(
+                        t['name'], t['is_readonly'], t['is_session_expired']
+                    )
+                )
+                continue
+
+            for chk in ['future_parts', 'parts_to_check', 'queue_size',
```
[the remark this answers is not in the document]
Felixoid: I've rewrote this to comprehenser, so now it's one for less and much more compact.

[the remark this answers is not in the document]
Felixoid: Yes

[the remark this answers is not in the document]
Felixoid: Check. Gone as well

src/check_clickhouse.py
```diff
@@ -0,0 +1,342 @@
...
+            for i in range(len(row)):
+                rd[columns[i][0]] = row[i]
+            result.append(rd)
+        return result
+
+    def exit(self, message: str):
+        print(self.code.name + ': ' + message)
+        exit(self.code.current)
+
+    def clusters(self, args: Namespace):
```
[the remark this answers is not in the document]
Felixoid: done here, ready for next iteration

src/check_clickhouse.py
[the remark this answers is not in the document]
Felixoid: refactored

src/check_clickhouse.py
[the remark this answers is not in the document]
Felixoid: I almost don't see a difference but in this case, would prefer the defaults from vim pymode plugin `¯\_(ツ)_/¯`

src/check_clickhouse.py
```diff
@@ -0,0 +1,344 @@
...
+        help='cluster names, could be defined multiple times',
+    )
+
+    parser_replication = subparsers.add_parser(
+        'replication', formatter_class=ArgumentDefaultsHelpFormatter,
+        help='check replication health, see https://clickhouse.tech/docs/en/'
+        'operations/system_tables/#system_tables-replicas',
+    )
+    parser_replication.set_defaults(func=Check.replication)
+    parser_replication.add_argument(
```
Felixoid: @TheFRedFox Do you have any idea, how is it possible to pass warning and critical thresholds here (and other places) in an elegant way?

src/check_clickhouse.py
```diff
@@ -11,17 +11,112 @@
...
 from clickhouse_driver.errors import (  # type: ignore  # missing stubs
-    NetworkError, ServerException
+    NetworkError, ServerException,
+)
+from logging import (
+    getLogger, basicConfig, CRITICAL as l_CRITICAL, WARN as l_WARN,
+    INFO as l_INFO, DEBUG as l_DEBUG,
 )
 from pprint import pformat
 from sys import exit
```
TheFRedFox: I guess I would prefer to just import sys and use `sys.exit` in code as `exit` is also a spread command in interactive python shells. No strong opinion though.

src/check_clickhouse.py
TheFRedFox: Hm, not really. I guess the most elegant way without bloating it up even more is to accept one or two values per argument with `nargs` and to use the lower/higher value for the respectively thresholds. If that is not possible, you could end up with your own value format like: `20w;10c` or so. Both needs to be documented quite well.
Another possibility is also to introduce even more subparsers to every argument and allow `warning` and `critical` arguments in those subparsers.

src/check_clickhouse.py
```diff
@@ -19,8 +19,10 @@
...
+from typing import Tuple, List
 
-
+# exit code and message
+ExitStruct = Tuple[int, str]
```
TheFRedFox: Please put a line between ExitStruct and logging to separate them semantically.

src/check_clickhouse.py
```diff
@@ -108,180 +110,101 @@ def parse_args() -> Namespace:
 def main():
     args = parse_args()
     log_levels = [l_CRITICAL, l_WARN, l_INFO, l_DEBUG]
-    args.verbose = min(3, args.verbose)
-    logger.setLevel(log_levels[args.verbose])
+    logger.setLevel(log_levels[min(args.verbose, 3)])
     logger.debug('Arguments are {}'.format(pformat(args.__dict__)))
     ch = Check(
```
TheFRedFox: Be as explicit as possible with variable names. And to save 3 characters it is not necessary to sacrifice explicitness.

src/check_clickhouse.py
```diff
@@ -108,180 +110,101 @@ def parse_args() -> Namespace:
...
-            logger.debug('Tables for cluster `{}` are {}'
-                         .format(cl, pformat(tables)))
-            # If tables exist - request single row to check availability
-            for t in tables:
-                t = '{}.{}'.format(Check.sanitize_name(t[0]),
-                                   Check.sanitize_name(t[1]))
-                logger.debug('Sanitized table name: {}'.format(t))
+            logger.debug('Tables for cluster `{cluster}` are {tables}'
+                         .format_map(cl))
+            for t in cl['tables']:
```
TheFRedFox: Again, like in the `replication` method:
An easy way to improve readability of nested loops is to introduce an inline function with a meaningful name:
``` python
def doTables(tables):
    for table in tables:
        ...

for cluster in clusters:
    ...
    doTables(cluster['tables'])
    ...

```

src/check_clickhouse.py
```diff
@@ -0,0 +1,334 @@
...
+# exit code and message
+ExitStruct = Tuple[int, str]
+basicConfig(
+    format='%(levelname)-8s [%(filename)s:%(lineno)d]:\n%(message)s'
+)
+logger = getLogger(__name__)
+
+
+def parse_args() -> Namespace:
+    """
```
TheFRedFox: Tbh this doc block is quite redundant. No need to have it here.

src/check_clickhouse.py
```diff
@@ -0,0 +1,334 @@
...
+def main():
+    args = parse_args()
+    log_levels = [l_CRITICAL, l_WARN, l_INFO, l_DEBUG]
+    logger.setLevel(log_levels[min(args.verbose, 3)])
+    logger.debug('Arguments are {}'.format(pformat(args.__dict__)))
+    ch = Check(
+        host=args.host, port=args.port, database='system',
+        user=args.user, password=args.password,
+    )
+    # Each subcommand executes appropriate method of Check
```
TheFRedFox: This should go one line below, directly above the `args.func` call.

src/check_clickhouse.py
```diff
@@ -0,0 +1,334 @@
...
+                    'Table `{name}`: is_readonly={is_readonly}, '
+                    'is_session_expired={is_session_expired}'.format_map(t)
+                )
+                continue
+
+            t_messages = [
+                '{}={} (max < {})'.format(c, t[c], config[c]) for c in checks
+                if config[c] <= t[c]
+            ]
+            if t_messages:
```
TheFRedFox:
``` python
if not t_messages:
    continue

self.code.current = Code.WARNING
...
```

src/check_clickhouse.py
```diff
@@ -0,0 +1,334 @@
...
+                self.code.current = Code.WARNING
+                message = 'Table `{}`: {}'.format(t['name'],
```
TheFRedFox: How about:
``` python
message = (
    'Table `{}`: {}'
    .format(t['name'], ', '.join(t_messages))
)
```

src/check_clickhouse.py
Felixoid: Than `cl` [here](https://github.com/innogames/igmonplugins/pull/184/files#diff-9f8a84e2070abf9db6d0cf202d316e12R199) and [here](https://github.com/innogames/igmonplugins/pull/184/files#diff-9f8a84e2070abf9db6d0cf202d316e12R194) won't mean anything, and it doesn't make sense to pass it into function. And doTable will change external [message](https://github.com/innogames/igmonplugins/pull/184/files#diff-9f8a84e2070abf9db6d0cf202d316e12R196)

Don't really see yet, how to do it meaningfull.

src/check_clickhouse.py
Felixoid: Yep, makes sense to not overwrite exit()

src/check_clickhouse.py
```diff
@@ -11,17 +11,112 @@
 # https://pypi.org/project/clickhouse-driver/
 from clickhouse_driver import Client  # type: ignore  # missing stubs
 from clickhouse_driver.errors import (  # type: ignore  # missing stubs
-    NetworkError, ServerException
+    NetworkError, ServerException,
```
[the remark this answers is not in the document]
TheFRedFox: Redundant diffs won't happen with inline enumerations. Where it absolutely make sense is for "list"-enumerations:
``` python
flexible_list_of_stuff = [
    'b',
    'c',
    'a',
]
```

Now if you add something, the diff will look like that:
``` diff
flexible_list_of_stuff = [
    'b',
    'c',
    'a',
+   'd',
]
```

Without trailing comma:
``` diff
flexible_list_of_stuff = [
    'b',
    'c',
-   'a'
+   'a',
+   'd',
]
```

You see the cleaner diff with the trailing comma there.

For inline enumeration it would both look like that:
``` diff
flexible_list_of_stuff = [
-   'b', 'c', 'a'(,)
+   'b', 'c', 'a', 'd'
]
```
With our without the trailing comma it would delete one line and add one line. The diff can't be made cleaner.

So, when we should use "list"-enumerations? Exactly like the name of the variable is named: For very flexible and highly changeable lists/maps or other enumerations. Or where the order of items is very important to see.

For imports we definitely do not have those situations: Imports are not flexible, they are not highly changeable and the order of imports is (mostly) not important to see.

Therefore, no trailing commas, except for "list"-enumeration. It's not PEP8, therefore also no hard feelings, I would not reject a PR just because of that.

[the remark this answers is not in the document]
Felixoid: Well, technically true, besides the fact that this imports errors to catch. I could assume, that it can change eventually, and then

```diff
  from clickhouse_driver.errors import (  # type: ignore  # missing stubs
      NetworkError, ServerException, AnotherException,
+     YetAnotherException,
  )
```

Would be

```diff
  from clickhouse_driver.errors import (  # type: ignore  # missing stubs
-     NetworkError, ServerException, AnotherException
+     NetworkError, ServerException, AnotherException,
+     YetAnotherException
  )
```

Esthetically, I prefer the first one.

src/check_clickhouse.py
Felixoid: I'll postpone it for now. Here I'm fine with warning only yet

src/check_clickhouse.py
Felixoid: Have found a way to implement it

src/check_clickhouse.py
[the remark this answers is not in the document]
TheFRedFox: Possible and still it should not be so likely to be changed that often. Anyway. As I said in person: It's ok to keep it, it is no rejection point.

src/check_clickhouse.py
```diff
@@ -0,0 +1,341 @@
...
+            .format(pformat(clusters))
+        )
+
+        existing = {c['cluster'] for c in clusters}
+        missing = set(config['clusters']) - existing
+        if missing:
+            self.code.current = Code.WARNING
+            messages.append('Clusters not found on server: {}'.format(missing))
+
+        def checkTables(tables: List[str]) -> List[str]:
```
TheFRedFox: Nice solution, although it is not 100% the same code now. Is it ok to not break after an Exception? Wondering as you had it this way before.

src/check_clickhouse.py
```diff
@@ -0,0 +1,341 @@
...
+
+        for cl in clusters:
+            logger.debug('Tables for cluster `{cluster}` are {tables}'
+                         .format_map(cl))
+            failed_tables = checkTables(cl['tables'])
+            if not failed_tables:
+                continue
+            self.code.current = Code.CRITICAL
+            messages.append(
+                'Cluster `{}`, failed to query from tables {}'.format(
```
TheFRedFox: I suggest:
``` python
messages.append(
    'text'
    .format(...)
```
No rejection point though.

src/check_clickhouse.py
TheFRedFox: See my new comment: https://github.com/innogames/igmonplugins/pull/184#discussion_r384421900

src/check_clickhouse.py
Felixoid: I've found that it's possible to have different problems for different tables inside one cluster, so yes, it's intended.

src/check_clickhouse.py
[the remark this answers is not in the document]
Felixoid: Ok, thank you. Not likely, but now it's a little bit more homogenous.

src/check_clickhouse.py
Felixoid: Sure, it's nicer

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+    parser_replication.add_argument(
+        '--log-delay', type=int, default=10, help='how big difference between '
+        'actions log in ZK and locally could be',
+    )
+    parser_replication.add_argument(
+        '--absolute-delay', type=int, default=40,
+        help='how big lag in seconds the current replica could have',
+    )
+
+    # Return float in range (0, 1] or raise ArgumentTypeError
```
kofrezo: - [ ] Typo  range(0,1)

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+    def factor(arg: str) -> float:
+        try:
+            f = float(arg)
+        except ValueError:
+            raise ArgumentTypeError("Must be a floating point number")
```
kofrezo: - [ ] Just let it flow. The Python error itself already says the same

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+        if not 0 < f <= 1:
```
kofrezo: - [ ] Use floats 0.0 and 1.0 to make clear you want floats here

src/check_clickhouse.py
Felixoid: It's not a typo, it's a math range notation https://stackoverflow.com/questions/4396290/what-does-this-square-bracket-and-parenthesis-bracket-notation-mean-first1-last

src/check_clickhouse.py
Felixoid: Yes, but I explicitly raise another type of error to get output:

```
> python check_clickhouse.py parts --delay-factor 0.03 --throw-factor asd
usage: check_clickhouse.py parts [-h] [--delay-factor DELAY_FACTOR] [--throw-factor THROW_FACTOR]
check_clickhouse.py parts: error: argument --throw-factor: Must be a floating point number
```

src/check_clickhouse.py
Felixoid: no, it's any between, not 0.0 or 1.0

Felixoid:
Thank you for review, @kofrezo

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+                self.code.current = Code.CRITICAL
+                messages.append(
+                    'Inserts into the next table:partition:parts would be'
+                    ' failed soon since they contain {}<=parts: {}'
+                    .format(throw, ','.join(r['tpp']))
+                )
+        messages = messages or ['Inserts and merges are going fine']
+        return self.exit('; '.join(messages))
+
+    def _get_parts(self, config: dict) -> List[dict]:
```
TheFRedFox: Please write again in the doc block which keys the `config` dict needs to have.

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+    def parts(self, config: dict) -> ExitStruct:
+        """
+        Checks if there are tables:partitions with too many active parts
+        config should contain next keys:
+            - delay_factor
+            - throw_factor
+        """
+        logger.debug('Parts check, config={}'.format(pformat(config)))
+        keys = ['delay_factor', 'throw_factor']
+        keys_in_config = [k in keys for k in config.keys() if k in keys]
```
TheFRedFox:
```suggestion
        keys_in_config = config.keys() & keys
```

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+                continue
+
+            self.code.current = Code.WARNING
+            message = 'Table `{}`: {}'.format(t['name'], ', '.join(t_messages))
+            messages.append(message)
+
+        messages = messages or ['All replicated tables are fine']
+        return self.exit('; '.join(messages))
+
+    def parts(self, config: dict) -> ExitStruct:
```
TheFRedFox: How likely is it, that you will have more and more configuration factors? If not, I would suggest to rather use explicit parameters here instead of implicit config parameter.

TheFRedFox: The whole method has no single blank line. It is hard to find logical scopes. Please check and find some scopes to separate to make it easier to read through the code.

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+        print(message)
+        sys.exit(code)
+    except NetworkError as e:
+        check.code.current = Code.CRITICAL
+        code, message = check.exit(e.message)
+        print(message)
+        sys.exit(code)
+
+
+class Check(object):
```
TheFRedFox: Btw, the class is getting quite big now. I would already suggest to split it up into several smaller classes now. Question: How likely is it, that you will add some more logic into this class? If there is like even a small percentage, how about to separate this into several smaller Check classes?

.gitignore
```diff
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+.mypy_cache
```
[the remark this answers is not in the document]
Felixoid: Moved to #188

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+        keys = ['delay_factor', 'throw_factor']
+        keys_in_config = [k in keys for k in config.keys() if k in keys]
+        if not len(keys) == len(keys_in_config):
+            raise KeyError('Not all of {} presented in config: {}'
+                           .format(keys, config))
+        messages = []  # type: List[str]
+        rows = self._get_parts(config)
+        logger.debug('Problems from CH are {}'.format(pformat(rows)))
+        # current thresholds from first result
+        delay, throw = rows[0]['delay'], rows[0]['throw']
```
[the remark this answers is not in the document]
Felixoid: Quite good catch. It's **very** rare case that there's no any *MergeTree table on the server, but still possible. Thank you

src/check_clickhouse.py
Felixoid: Still, it's executed out of the `args.func(check, args.__dict__)`

Or could you rephrase it, if I understood it wrong?

src/check_clickhouse.py
Felixoid: It's very likely. I guess, it's better to have online call to discuss the changes

src/check_clickhouse.py
TheFRedFox: Ah no, I got that wrong. Sry, I first though that this is rather a helper function and not a check. Yeah, then forgot what I wrote for this line, but then same thing for the `_get_parts` function. Utility functions/methods are better to have explicit parameters.

src/check_clickhouse.py
Felixoid: Replaced it with arguments

src/check_clickhouse.py
Felixoid:
```
In [2]: ['asd', 'delay_factor'] & keys
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-2289fc7b1b2a> in <module>
----> 1 ['asd', 'delay_factor'] & keys

TypeError: unsupported operand type(s) for &: 'list' and 'list'
```

Doesn't seem to work

src/check_clickhouse.py
Felixoid: done

Felixoid: Should be better now

src/check_clickhouse.py
Felixoid: The class is split now with inheritance

src/check_clickhouse.py
Felixoid: This works though, and it's nicer. Thank you

TheFRedFox: `config.keys()` should already be set-like:
``` python
keys = {'delay_factor', 'throw_factor'}
config.keys() & keys
```
should work.

Felixoid: Ah, but keys was a list. Yes, it works with `config.keys() & keys`

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+# exit code and message
+ExitStruct = Tuple[int, str]
+
+logging.basicConfig(
+    format='%(levelname)-8s [%(filename)s:%(lineno)d]:\n%(message)s'
+)
+logger = logging.getLogger(__name__)
+
+
+def parse_args() -> Namespace:
```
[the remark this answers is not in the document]
Felixoid: I've put each subparser to dedicated function, it's much more readable now.

src/check_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+        '--log-delay', type=int, default=10, help='how big difference between '
+        'actions log in ZK and locally could be',
+    )
+    parser_replication.add_argument(
+        '--absolute-delay', type=int, default=40,
+        help='how big lag in seconds the current replica could have',
+    )
+
+    # Return float in range (0, 1] or raise ArgumentTypeError
+    def factor(arg: str) -> float:
```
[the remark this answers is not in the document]
Felixoid: Feel free to reopen in case if you are still into making dedicated class

src/check_clickhouse.py
```diff
@@ -0,0 +1,506 @@
...
+        '-u', '--user', default='default', help='ClickHouse username',
+    )
+    parser.add_argument(
+        '-p', '--password', default='', help='ClickHouse username',
+    )
+
+    subparsers = parser.add_subparsers(dest='check_name')
+    subparsers.required = True
+
+    parse_clusters(subparsers)
```
TheFRedFox: How about naming them `add_subparser_clusters` as `parse_` sounds like it is already parsing things.

src/check_clickhouse.py
kofrezo: Thats my point. I don't think your message is better: It says "Must be a floating point number" but does not show the user what (wrong) value he has entered. I think the original exception is better and it bloats your code without adding value or functionality.

TheFRedFox: A compromise could be:
``` python
except ValueError as e:
    raise ArgumentTypeError('Must be float, ' + str(e))
```

src/check_clickhouse.py
```diff
@@ -0,0 +1,500 @@
...
+            .format(pformat(clusters))
+        )
+
+        existing = {c['cluster'] for c in clusters}
+        missing = set(config['clusters']) - existing
+        if missing:
+            self.code.current = Code.CRITICAL
+            messages.append('Clusters not found on server: {}'.format(missing))
+
+        def checkTables(tables: List[str]) -> List[str]:
```
TheFRedFox: Sry, actually my bad as it was in the suggestion. Too many switches between programming languages. `pascalCase` for methods is not pythonic, so:
```suggestion
        def check_tables(tables: List[str]) -> List[str]:
```

Felixoid: Done in another commit.

src/check_clickhouse.py
```diff
@@ -4,6 +4,14 @@
 #
 # This check tests multiple clickhouse-server's health metrics
 #
+# Usage examples:
+#
+#   check_clikhouse.py clusters -c cluster_name
+#   check_clikhouse.py parts
+#   check_clikhouse.py replication
```
TheFRedFox:
```suggestion
#   check_clickhouse.py clusters -c cluster_name
#   check_clickhouse.py parts
#   check_clickhouse.py replication
```

src/check_clickhouse.py
```diff
@@ -4,6 +4,14 @@
...
+#
+# For details see --help/-h for each `check_clikhouse.py`a and each check_name
```
TheFRedFox:
```suggestion
# For details see --help/-h for each `check_clickhouse.py` and each check_name
```

PR #185, opened 2020-03-01
Introduce the BIRD protocols check

lamaral:
This is a python3 replacement for the check_bgp_peers.sh.

check_bgp_peers.sh has hardcoded paths and can only check protocols for
bird IPv4.
This check is capable of checking either protocol, requiring just that
you provide the proper path to the birdc/birdc6 binary, meaning that for
checking IPv4 and IPv6, you will need to run the check twice.

It will return critical if a protocol is not up and established, meaning
that even if you disabled a protocol on purpose, it will still return
critical.

### DIFF
--- a/src/check_bird_protocols.py
+++ b/src/check_bird_protocols.py
@@ -1,0 +1,226 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - BIRD BGP Protocols Health Check
+
+This is a Nagios script which queries BIRD to check if all protocols
+are up and established/running.
+
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import sys
+
+from argparse import ArgumentParser
+from subprocess import check_output, STDOUT, CalledProcessError
+
+
+def parse_args():
+    parser = ArgumentParser(
+        description='Check if all BIRD BGP protocols are up and established'
+    )
+    parser.add_argument('path', help='Path to the birdc/birdc6 binary')
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    birdc_path = args.path
+
+    code, reason = check_birdc_protocols(birdc_path)
+
+    print(format_nagios_message(code, reason))
+    sys.exit(code)
+
+
+def check_birdc_protocols(path):
+    try:
+        output = check_output([path, 'show', 'protocols'], stderr=STDOUT)
+    except (FileNotFoundError, PermissionError):
+        return (
+            ExitCodes.critical,
+            'The path to the birdc binary does not exist or is not executable'
+        )
+    except CalledProcessError:
+        return (
+            ExitCodes.critical,
+            'birdc returned a non-zero exit code'
+        )
+
+    if not output:
+        return (ExitCodes.critical, 'Could not retrieve BIRD\'s output')
+
+    protocols_parsed = parse_birdc_output(output.decode())
+
+    if not protocols_parsed:
+        return (ExitCodes.critical, 'Could not find any BGP protocol entries')
+
+    result = {
+        ProtocolStates.up: [],
+        ProtocolStates.down: [],
+        ProtocolStates.disabled: [],
+        ProtocolStates.unknown: [],
+    }
+
+    for protocol in protocols_parsed:
+        if protocol['type'] == 'BGP':
+            ret = check_bgp_protocol(protocol)
+        elif protocol['type'] == 'OSPF':
+            ret = check_ospf_protocol(protocol)
+        elif protocol['type'] in ['Device', 'Kernel', 'Static', 'Direct']:
+            ret = check_other_protocol(protocol)
+        else:
+            ret = ProtocolStates.unknown
+
+        result[ret].append(protocol['name'])
+
+    if result[ProtocolStates.down]:
+        return (
+            ExitCodes.critical,
+            'Some protocols are down: {}'
+            .format(', '.join(result[ProtocolStates.down]))
+        )
+    elif result[ProtocolStates.disabled]:
+        return (
+            ExitCodes.warning,
+            'Some protocols are disabled: {}'
+            .format(', '.join(result[ProtocolStates.disabled]))
+        )
+    elif result[ProtocolStates.unknown]:
+        return (
+            ExitCodes.unknown,
+            'Unknown protocols are seen: {}'
+            .format(', '.join(result[ProtocolStates.unknown]))
+        )
+    else:
+        return (ExitCodes.ok, 'All protocols are up')
+
+
+def check_bgp_protocol(protocol):
+    def is_bgp_up(protocol):
+        is_established = protocol['info'] == 'Established'
+        is_up = protocol['state'] == 'up'
+        return is_up and is_established
+
+    def is_bgp_disabled(protocol):
+        state_down = protocol['state'] == 'down'
+        return state_down
+
+    if is_bgp_up(protocol):
+        return ProtocolStates.up
+    elif is_bgp_disabled(protocol):
+        return ProtocolStates.disabled
+    else:
+        return ProtocolStates.down
+
+
+def check_ospf_protocol(protocol):
+    def is_ospf_up(protocol):
+        is_up = protocol['state'] == 'up'
+        is_running = protocol['info'] == 'Running'
+        return is_up and is_running
+
+    def is_ospf_disabled(protocol):
+        is_down = protocol['state'] == 'down'
+        return is_down
+
+    if is_ospf_up(protocol):
+        return ProtocolStates.up
+    elif is_ospf_disabled(protocol):
+        return ProtocolStates.disabled
+    else:
+        return ProtocolStates.down
+
+
+def check_other_protocol(protocol):
+    if protocol['state'] == 'up':
+        return ProtocolStates.up
+    elif protocol['state'] == 'down':
+        return ProtocolStates.disabled
+    else:
+        return ProtocolStates.down
+
+
+def parse_birdc_output(output):
+    # Matching protocol lines from BIRD:
+    # BIRD 1.6.3 ready.
+    # name     proto    table    state  since       info
+    # direct1  Direct   master   up     2020-02-27
+    # b_af     BGP      master   up     2020-02-28  Established
+    # b_lb     BGP      master   start  18:43:46    Active        Socket: conn
+    # b_home   BGP      master   down   2020-02-29
+    # We are extracting the protocol name, state, date/time since last
+    # change and the additional information about the state of the session.
+    lines = output.splitlines()
+    lines = lines[2:]
+
+    tokenized_lines = [
+        [field.strip() for field in line.split(None, 5)]
+        for line in lines
+    ]
+    protocols_dicts = [
+        build_dict(line) for line in tokenized_lines
+    ]
+
+    return protocols_dicts
+
+
+def build_dict(line):
+    # The next line ensures that we always have 6 elements in the line,
+    # as the tokenization of protocols that are disabled yield only 5 elements.
+    line += [None] * (6 - len(line))
+    return {
+        'name': line[0],
+        'type': line[1],
+        'table': line[2],
+        'state': line[3],
+        'since': line[4],
+        'info': line[5],
+    }
+
+
+def format_nagios_message(code, reason):
+    if code == ExitCodes.ok:
+        state_text = 'OK'
+    elif code == ExitCodes.warning:
+        state_text = 'WARNING'
+    elif code == ExitCodes.critical:
+        state_text = 'CRITICAL'
+    else:
+        state_text = 'UNKNOWN'
+    return '{0} - {1}'.format(state_text, reason)
+
+
+class ProtocolStates:
+    up = 0
+    disabled = 1
+    down = 2
+    unknown = 3
+
+
+class ExitCodes:
+    ok = 0
+    warning = 1
+    critical = 2
+    unknown = 3
+
+
+if __name__ == '__main__':
+    main()


lamaral:
@TuxPowered42 please check if this runs properly in our routers and HWLBs. Also please give me your comments regarding the fact that the check still returns critical for protocols that were disabled on purpose e.g. `birdc disable b_artfiles`. Should we maybe return a warning in those cases, since it's a somewhat expected behavior but not the normal operating behavior?

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+from subprocess import Popen, PIPE, STDOUT
+
+
+def parse_args():
+    parser = ArgumentParser(
+        description='Check if all BIRD BGP protocols are up and established'
+    )
+    parser.add_argument(
+        'path',
+        type=str,
```
TheFRedFox: `type=str` is default and also broadly known and logical. Explicitly is not needed here.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        help='Path to the birdc/birdc6 binary'
+    )
```
TheFRedFox: Without `type=str` I guess it will fit in one line.

```suggestion
    parser.add_argument('path', help='Path to the birdc/birdc6 binary')
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+    args = parse_args()
+    birdc_path = args.path
+
+    # Check if the path to birdc actually exists and is executable
+    if not check_birdc_executable(birdc_path):
+        print_nagios_message(
+            ExitCodes.critical,
+            'The path to the birdc executable does not exist/is not executable'
+        )
+        exit(ExitCodes.critical)
```
TheFRedFox: Please do not use `exit` but `sys.exit`, check also for other occasions.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+
+    print_nagios_message(code, reason)
+    exit(code)
+
+
+def check_birdc_peers(path):
+    protocols = get_birdc_protocols(path)
+
+    if not protocols:
+        return ExitCodes.critical, 'Could not retrieve BIRD\'s ' \
```
TheFRedFox: Please do not use `\` for multilines. Go for:
``` python
return (
    ExitCodes.critical, 'Could not retrieve BIRD\'s output'
)
```

In general, I am also more fan of stating return of tuples explicitly.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+
+    try:
+        r = re.compile(
+            r'^(?P<name>\w+)\s+BGP\s+(\w+)\s+(?P<state>\w+)\s+'
+            r'(?P<since>[\d\-:]+)\s+(?P<info>[\w]+)?\s*$',
+            re.MULTILINE
+        )
+
+        res = [m.groupdict() for m in r.finditer(protocols)]
+    except Exception:
```
TheFRedFox: Too broad catch here. Please check which Exceptions can be thrown here and except them explicitly.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        return ExitCodes.critical, 'Could not parse BIRD\'s output'
+
+    if not res:
+        return ExitCodes.critical, 'Could not parse BIRD\'s output'
+
+    down_protocols = [protocol['name'] for protocol in res
```
TheFRedFox: Btw, with switching the words you will get a better naming of the variable: `protocols_down`
`down` is an attribute of the scope of `protocols`, therefore first `protocols` then `down`.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+    print_nagios_message(code, reason)
+    exit(code)
+
+
+def check_birdc_peers(path):
+    protocols = get_birdc_protocols(path)
+
+    if not protocols:
+        return ExitCodes.critical, 'Could not retrieve BIRD\'s ' \
+                                   'output'
```
TheFRedFox:
```suggestion
        return (
            ExitCodes.critical, 'Could not retrieve BIRD\'s output'
        )
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+
+    try:
+        r = re.compile(
```
TheFRedFox: Please comment this regex. What input do you have and what do we want to match with this? Like this it is impossible for me to validate the correctness of the regex.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        return ExitCodes.critical, 'Could not parse BIRD\'s output'
+
+    if not res:
+        return ExitCodes.critical, 'Could not parse BIRD\'s output'
+
+    down_protocols = [protocol['name'] for protocol in res
+                      if protocol['info'] != 'Established']
+
+    if down_protocols:
+        return ExitCodes.critical, 'Some protocols are down!'
```
TheFRedFox: More a fan of explicit tuple returning:

```suggestion
        return (ExitCodes.critical, 'Some protocols are down!')
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+    else:
+        return ExitCodes.ok, 'All protocols are up'
```
TheFRedFox: Same as above.

```suggestion
        return (ExitCodes.ok, 'All protocols are up')
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+
+        res = [m.groupdict() for m in r.finditer(protocols)]
+    except Exception:
+        return ExitCodes.critical, 'Could not parse BIRD\'s output'
+
+    if not res:
+        return ExitCodes.critical, 'Could not parse BIRD\'s output'
+
+    down_protocols = [protocol['name'] for protocol in res
+                      if protocol['info'] != 'Established']
```
TheFRedFox: Is still in discussion, and I am more fan of handling multlines like that:
```suggestion
    down_protocols = [
        protocol['name'] for protocol in res
        if protocol['info'] != 'Established'
    ]
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        return ExitCodes.ok, 'All protocols are up'
+
+
+def get_birdc_protocols(path):
+    bird = Popen(
+        [
+            path,
+            'show',
+            'protocols',
+        ],
```
TheFRedFox: For "production" this is still absolutely readable and much less lines.

```suggestion
        [path, 'show', 'protocols'],
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        stdout=PIPE,
+        stderr=STDOUT,
+    )
```
TheFRedFox: Maybe the whole structure is not needed:
```suggestion
    bird = Popen([path, 'show', 'protocols'], stdout=PIPE, stderr=STDOUT)
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+    content = bird.communicate()[0].decode()
```
TheFRedFox: What will happen if `bird.communicate()` is empty?

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,141 @@
...
+def print_nagios_message(code, reason):
+    if code == ExitCodes.ok:
+        state_text = 'OK'
+    elif code == ExitCodes.warning:
+        state_text = 'WARNING'
+    elif code == ExitCodes.critical:
+        state_text = 'CRITICAL'
+    else:
+        state_text = 'UNKNOWN'
+    print("{0} - {1}".format(state_text, reason))
```
TheFRedFox: Even better would be to let this message return the message and the printing will be done by the caller.

src/check_bgp_protocols.py
lamaral: It was Pycharm formatting :sweat_smile:

src/check_bgp_protocols.py
lamaral: Hm, good catch. I'll treat that.

src/check_bgp_protocols.py
lamaral: I think I don't need this try catch anymore. I believe I had this try catch in place because I had a slightly different logic here before. 
Now, if the BIRD output that is being parsed by the regex doesn't have anything that matches, `res` will be empty and the `if` right after that will stop the execution.

src/check_bgp_protocols.py
```diff
@@ -69,27 +69,24 @@ def check_birdc_peers(path):
...
         )
 
-    try:
-        r = re.compile(
-            r'^(?P<name>\w+)\s+BGP\s+(\w+)\s+(?P<state>\w+)\s+'
-            r'(?P<since>[\d\-:]+)\s+(?P<info>[\w]+)?\s*$',
-            re.MULTILINE
-        )
+    r = re.compile(
+        r'^(?P<name>\w+)\s+BGP\s+(\w+)\s+(?P<state>\w+)\s+'
```
TheFRedFox: FYI if you want to group things without capturing them, you can use `(?:)` to define a non capturing group.

src/check_bgp_protocols.py
```diff
@@ -69,27 +69,24 @@ def check_birdc_peers(path):
...
+        r'(?P<since>[\d\-:]+)\s+(?P<info>[\w]+)?\s*$',
+        re.MULTILINE
+    )
 
-        res = [m.groupdict() for m in r.finditer(protocols)]
-    except Exception:
-        return ExitCodes.critical, 'Could not parse BIRD\'s output'
+    res = [m.groupdict() for m in r.finditer(protocols)]
```
TheFRedFox: Now I see it! It should break now. You were safe before because of the `except`. `m.groupdict()` was throwing an Exception for None items whenever `finditer` does not match.
Please check again my theory with your script. If it is breaking now, you can easily change it with:
```suggestion
    res = [
        m.groupdict() for m in r.finditer(protocols)
        if m is not None
    ]
```

src/check_bgp_protocols.py
```diff
@@ -67,6 +67,16 @@ def check_birdc_peers(path):
...
+    # Here we match the BGP protocol lines from BIRD.
+    # Below is an example of how they look like:
+    # direct1  Direct   master   up     2020-02-27
+    # static_bgp Static   master   up     2020-02-21
+    # b_firstroot BGP      master   up     2020-02-28  Established
+    # b_lb     BGP      master   start  18:43:46    Active        Socket: conn
+    # b_af     BGP      master   up     2020-02-29  Established
+    # b_home   BGP      master   down   2020-02-29
+    # We are extracting the protocol name, state, date/time since last
+    # change and the additional information about the state of the session.
```
TheFRedFox: Thx for the comment. It's good, can even be reduced.

src/check_bgp_protocols.py
```diff
@@ -67,6 +67,16 @@ def check_birdc_peers(path):
     if not protocols:
         return (ExitCodes.critical, 'Could not retrieve BIRD\'s output')
 
+    # Here we match the BGP protocol lines from BIRD.
+    # Below is an example of how they look like:
```
TheFRedFox:
```suggestion
    # Matching protocol lines from BIRD:
```

src/check_bgp_protocols.py
```diff
@@ -67,6 +67,16 @@ def check_birdc_peers(path):
...
+    # direct1  Direct   master   up     2020-02-27
+    # static_bgp Static   master   up     2020-02-21
+    # b_firstroot BGP      master   up     2020-02-28  Established
+    # b_lb     BGP      master   start  18:43:46    Active        Socket: conn
+    # b_af     BGP      master   up     2020-02-29  Established
+    # b_home   BGP      master   down   2020-02-29
```
TheFRedFox: I just need all possible different lines. With those lines given, we can reduce it to:
```suggestion
    # direct1  Direct   master   up     2020-02-27
    # b_firstroot BGP      master   down     2020-02-28  Established
    # b_lb     BGP      master   start  18:43:46    Active        Socket: conn
```

src/check_bgp_protocols.py
```diff
@@ -67,6 +67,16 @@ def check_birdc_peers(path):
...
+    # We are extracting the protocol name, state, date/time since last
+    # change and the additional information about the state of the session.
     r = re.compile(
         r'^(?P<name>\w+)\s+BGP\s+(\w+)\s+(?P<state>\w+)\s+'
         r'(?P<since>[\d\-:]+)\s+(?P<info>[\w]+)?\s*$',
```
TheFRedFox: Tbh., are you sure this will also match the line with `Socket: conn`?

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+    parser.add_argument('path', help='Path to the birdc/birdc6 binary')
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    birdc_path = args.path
+
+    # Check if the path to birdc actually exists and is executable
```
TheFRedFox: Redundant comment. Method name should already be meaningful enough.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+    if not check_birdc_executable(birdc_path):
+        print(format_nagios_message(
+            ExitCodes.critical,
+            'The path to the birdc executable does not exist/is not executable'
+        ))
+        sys.exit(ExitCodes.critical)
+
+    # Run the actual check and get the results
```
TheFRedFox: If one needs to comment a single line, one should think of renaming stuff as it is not meaningful enough.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+
+    if returncode > 0:
+        return None
+
+    return content
+
+
+def check_birdc_executable(path):
+    """Check if the birdc executable exists"""
+    return os.path.isfile(path) and os.access(path, os.X_OK)
```
TheFRedFox: Race Condition is actually possible here. However, in my opinion not in a risk scope as for accessing and modifying you will need to have server access anyway.

TheFRedFox: Just FYI: If you want to code it more secure. Do not check existence and executable flag before, but just try to run the script and catch the upcoming Exceptions.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+def format_nagios_message(code, reason):
+    if code == ExitCodes.ok:
+        state_text = 'OK'
+    elif code == ExitCodes.warning:
+        state_text = 'WARNING'
+    elif code == ExitCodes.critical:
+        state_text = 'CRITICAL'
+    else:
+        state_text = 'UNKNOWN'
+    return "{0} - {1}".format(state_text, reason)
```
TheFRedFox:
```suggestion
    return '{0} - {1}'.format(state_text, reason)
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def format_nagios_message(code, reason):
+    if code == ExitCodes.ok:
+        state_text = 'OK'
+    elif code == ExitCodes.warning:
+        state_text = 'WARNING'
+    elif code == ExitCodes.critical:
+        state_text = 'CRITICAL'
+    else:
```
TheFRedFox: Btw: 👍 for defensive programming. ;)

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+    ]
+
+    if protocols_down:
+        return (ExitCodes.critical, 'Some protocols are down!')
+    else:
+        return (ExitCodes.ok, 'All protocols are up')
+
+
+def get_birdc_protocols(path):
+    bird = Popen([path, 'show', 'protocols', ], stdout=PIPE, stderr=STDOUT)
```
TheFRedFox: How likely is it that we will extend/change the list here?
If not likely, please go for:
```suggestion
    bird = Popen([path, 'show', 'protocols'], stdout=PIPE, stderr=STDOUT)
```

src/check_bgp_protocols.py
lamaral: Hm, you're right. That slid through my testing.

src/check_bgp_protocols.py
lamaral: I'm now catching `TimeoutExpired`, in case the process took too long to run and `IndexError`, in case it doesn't return a proper tuple.

src/check_bgp_protocols.py
lamaral: I'm pushing a fix for this.

src/check_bgp_protocols.py
lamaral: That slid through from when I moved away from a multi-line. 

It's fixed in aa67e92.

lamaral: That slid through from when I moved away from a multi-line.

It's fixed in aa67e92.

src/check_bgp_protocols.py
lamaral: This is fixed with 90f31c5.

src/check_bgp_protocols.py
TheFRedFox: `TimeoutExpired` sounds correct. `IndexError` not really resp. you could just check `bird.communicate()` before accessing it:
``` python
content = bird.communicate()
if not content:
    return  # or whatever

content = content[0].decode()
decoded = content[0].decode() 
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,137 @@
...
+
+def get_birdc_protocols(path):
+    bird = Popen([path, 'show', 'protocols'], stdout=PIPE, stderr=STDOUT)
+    try:
+        content = bird.communicate()[0].decode()
+    except TimeoutExpired:
+        return None
+    except IndexError:
+        return None
+    return_code = bird.wait()
```
TuxPowered42: You can greatly simplify this code by uising subprocess.check_output.

src/check_bgp_protocols.py
```diff
@@ -67,19 +67,16 @@ def check_birdc_peers(path):
...
     # b_lb     BGP      master   start  18:43:46    Active        Socket: conn
-    # b_af     BGP      master   up     2020-02-29  Established
     # b_home   BGP      master   down   2020-02-29
     # We are extracting the protocol name, state, date/time since last
     # change and the additional information about the state of the session.
     r = re.compile(
-        r'^(?P<name>\w+)\s+BGP\s+(\w+)\s+(?P<state>\w+)\s+'
-        r'(?P<since>[\d\-:]+)\s+(?P<info>[\w]+)?\s*$',
+        r'^(?P<name>\w+)\s+BGP\s+(?:\w+)\s+(?P<state>\w+)\s+'
+        r'(?P<since>[\d\-:]+)\s*(?P<info>[\w]+)?.*$',
```
TheFRedFox: Talked about the possibility of a ReDOS attack. As the content is created by the bird program and for the case of having this content maliciously being modified we already have other problems.
Therefore the ReDOS attack vector is out of scope here and won't be protected.

src/check_bgp_protocols.py
TheFRedFox: Removal of catching resulted into another problem which will be tacked in another conversation: https://github.com/innogames/igmonplugins/pull/185#discussion_r386593160

src/check_bgp_protocols.py
lamaral: Tackled in 0edf4ee.

src/check_bgp_protocols.py
lamaral: I think this doesn't apply anymore, because of the subprocess.check_output simplifications from 0edf4ee.

src/check_bgp_protocols.py
lamaral: I have tested this extensively with some crafted output and couldn't hit any case where it contained None. 
I tried with empty lines, lines with just a space and couldn't make it happen. I think the `if m is not None` is a small change and doesn't really hurt, so I'll add it just in case.

lamaral:
@TheFRedFox @TuxPowered42 could you please give your blessing on this PR? :smile:

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,124 @@
...
+    # Matching protocol lines from BIRD:
+    # direct1  Direct   master   up     2020-02-27
+    # b_af     BGP      master   up     2020-02-28  Established
+    # b_lb     BGP      master   start  18:43:46    Active        Socket: conn
+    # b_home   BGP      master   down   2020-02-29
+    # We are extracting the protocol name, state, date/time since last
+    # change and the additional information about the state of the session.
+    r = re.compile(
+        r'^(?P<name>\w+)\s+BGP\s+(?:\w+)\s+(?P<state>\w+)\s+'
+        r'(?P<since>[\d\-:]+)\s*(?P<info>[\w]+)?.*$',
```
TheFRedFox: The `\s*` right before the optional info group needs to be ungreedy. I guess with this regex you could be safe now:
```suggestion
        r'(?P<since>[\d\-:]+)\s*?(?:(?P<info>[\w]+)|$).*?$',
```
And maybe we should consider to rather tokenise it as you also actually want to do it.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,124 @@
...
+        m.groupdict() for m in r.finditer(protocols.decode())
+        if m is not None
+    ]
+
+    if not res:
+        return (ExitCodes.critical, 'Could not find any BGP protocol entries')
+
+    protocols_down = [
+        protocol['name'] for protocol in res
+        if protocol['info'] != 'Established'
```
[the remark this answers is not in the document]
lamaral: Well spotted. It's updated.

src/check_bgp_protocols.py
lamaral: I applied this change to the regex. 

Regarding the tokenization, the data doesn't use tabs, but spaces instead. Could we still do it reliably?

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+        if protocol.get('info', '') != 'Established'
+    ]
+
+    if protocols_down:
+        return (ExitCodes.critical, 'Some protocols are down!')
+    else:
+        return (ExitCodes.ok, 'All protocols are up')
+
+
+def parse_birdc_output(protocols):
```
TheFRedFox: I would name the parameter `output`, especially as `protocols` sounds like it is a list.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+
+    tokenized_lines = [
+        [field.strip() for field in line.split(None, 5)]
+        for line in lines
+    ]
+    bgp_lines = [line for line in tokenized_lines if line[1] == 'BGP']
+
+    dicts = build_dicts(bgp_lines)
+
+    return dicts
```
TheFRedFox: I would suggest to make it a bit different. Make `build_dicts` rather to `build_dict` and make it a generic one without depending on being for `BGP`. This way, you can easily reuse it in the future.
```suggestion
    bgp_dicts = [
        build_dict(line) for line in tokenized_lines if line[1] == 'BGP'
    ]
    
    return bgp_dicts
```

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def build_dicts(bgp_lines):
```
TheFRedFox: I would rather suggest to make this taking just one line. And make it independent of being for `BGP`, so you can reuse it later easier.

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,138 @@
...
+    result = []
+    for line in bgp_lines:
+        result.append({
+            'name': line[0],
+            'state': line[3],
+            'since': line[4],
+            'info': line[5] if len(line) > 5 else None
```
TheFRedFox: I guess, you rather are searching for something like filling the list, so it is always 6 elements:
Use above that here:
``` python
line =+ [None] * (6 - len(line))
```

src/check_bgp_protocols.py
lamaral: Doesn't apply anymore

src/check_bgp_protocols.py
```diff
@@ -52,33 +52,36 @@ def main():
...
+    protocols_parsed = parse_birdc_output(output.decode())
 
-    if not parsed_protocols:
+    if not protocols_parsed:
         return (ExitCodes.critical, 'Could not find any BGP protocol entries')
 
     protocols_down = [
-        protocol['name'] for protocol in parsed_protocols
+        protocol['name'] for protocol in protocols_parsed
         if protocol.get('info', '') != 'Established'
```
TheFRedFox: The null safe is not needed anymore. If you want you can remove it again.

src/check_bgp_protocols.py
lamaral: This was simplified in a1a6d61

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,144 @@
...
+
+import sys
+
+from argparse import ArgumentParser
+from subprocess import check_output, STDOUT
+
+
+def parse_args():
+    parser = ArgumentParser(
+        description='Check if all BIRD BGP protocols are up and established'
```
TuxPowered42: Why only BGP protocols? What if we run OSPF too?

src/check_bgp_protocols.py
```diff
@@ -0,0 +1,144 @@
...
+    args = parse_args()
+    birdc_path = args.path
+
+    code, reason = check_birdc_peers(birdc_path)
+
+    print(format_nagios_message(code, reason))
+    sys.exit(code)
+
+
+def check_birdc_peers(path):
```
TuxPowered42: It's protocols, not peers.

src/check_bgp_protocols.py
lamaral: That was my next question. 
I will then adapt it to check other protocols.

TuxPowered42: I think it makes sense to check all protocol types. I think we still have OSPF in the office.

TheFRedFox: @TuxPowered42 do you have a list of protocols which would be interesting to check? Maybe we should make a choice argument to prevent typos. However, I am not sure about that.

lamaral: Each protocol needs to be handled different, due to the info field. 
OSPF for example, has `Running` or `Alone` in the info field, while BGP has `Established` and many other messages.

src/check_bird_protocols.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+    elif result[ProtocolStates.unknown]:
+        return (
+            ExitCodes.unknown,
+            'Unknown protocols are seen: {}'.format(
+                ', '.join(result[ProtocolStates.unknown])
+            )
+        )
+    else:
+        return (ExitCodes.ok, 'All protocols are up')
```
TheFRedFox: If there will come more types to check, we should reconsider to separate this code, as we could nicely put this into an own Class or the like.
For now, I would say it is fair enough.

src/check_bird_protocols.py
```diff
@@ -0,0 +1,223 @@
...
+            'Some protocols are disabled: {}'.format(
+                ', '.join(result[ProtocolStates.disabled])
+            )
+        )
+    elif result[ProtocolStates.unknown]:
+        return (
+            ExitCodes.unknown,
+            'Unknown protocols are seen: {}'.format(
+                ', '.join(result[ProtocolStates.unknown])
+            )
```
TheFRedFox:
```suggestion
            ExitCodes.unknown,
            'Unknown protocols are seen: {}'
            .format(', '.join(result[ProtocolStates.unknown]))
```

src/check_bird_protocols.py
```diff
@@ -0,0 +1,223 @@
...
+            'Some protocols are down: {}'.format(
+                ', '.join(result[ProtocolStates.down])
+            )
+        )
+    elif result[ProtocolStates.disabled]:
+        return (
+            ExitCodes.warning,
+            'Some protocols are disabled: {}'.format(
+                ', '.join(result[ProtocolStates.disabled])
+            )
```
TheFRedFox:
```suggestion
            ExitCodes.warning,
            'Some protocols are disabled: {}'
            .format(', '.join(result[ProtocolStates.disabled]))
```

src/check_bird_protocols.py
```diff
@@ -0,0 +1,223 @@
...
+            ret = ProtocolStates.unknown
+
+        result[ret].append(protocol['name'])
+
+    if result[ProtocolStates.down]:
+        return (
+            ExitCodes.critical,
+            'Some protocols are down: {}'.format(
+                ', '.join(result[ProtocolStates.down])
+            )
```
TheFRedFox:
```suggestion
            ExitCodes.critical,
            'Some protocols are down: {}'
            .format(', '.join(result[ProtocolStates.down]))
```

src/check_bird_protocols.py
```diff
@@ -0,0 +1,223 @@
...
+    ]
+    protocols_dicts = [
+        build_dict(line) for line in tokenized_lines
+    ]
+
+    return protocols_dicts
+
+
+def build_dict(line):
+    line += [None] * (6 - len(line))
```
TheFRedFox: As this is not too common, please comment this line with a short explanation.

src/check_bird_protocols.py
```diff
@@ -0,0 +1,223 @@
...
+    code, reason = check_birdc_protocols(birdc_path)
+
+    print(format_nagios_message(code, reason))
+    sys.exit(code)
+
+
+def check_birdc_protocols(path):
+    try:
+        # output = check_output([path, 'show', 'protocols'], stderr=STDOUT)
+        output = check_output(['./' + path], stderr=STDOUT)
```
TheFRedFox: I guess that's a forgotten 'debug' code or the like.

lamaral: You guessed right :smile:

src/check_bird_protocols.py
lamaral: I usually avoid this because pycharm's autoformat messes it up :/

src/check_bird_protocols.py
lamaral: I highly doubt that we will ever extend the list of types, but if the time for that comes, I will keep this in mind.

TheFRedFox: Sure, fair enough.

src/check_bird_protocols.py
```diff
@@ -0,0 +1,221 @@
...
+    protocols_dicts = [
+        build_dict(line) for line in tokenized_lines
+    ]
+
+    return protocols_dicts
+
+
+def build_dict(line):
+    # The next line ensures that we always have 6 elements in the line,
+    # as the tokenization of protocols that are disabled yeld only 5 elements.
```
TheFRedFox:
```suggestion
    # as the tokenization of protocols that are disabled yield only 5 elements.
```
of
```suggestion
    # as the tokenization of protocols that are disabled hold only 5 elements.
```
Not sure if you wanted to say `yield` or `hold`. ^^

src/check_bird_protocols.py
TheFRedFox: Mhm .. yeah, fair enough. I just like this style as it is very clear to read.

src/check_bird_protocols.py
lamaral: I meant yield, but my keyboard ate the I :joy:

PR #186, opened 2020-03-02
Add port argument to check_https_domains

lamaral:
We need to check domains on a service that runs a port different than 443. 

This PR add a new `-p` argument to check check_https_domains that allows you to define a custom port. It defaults to 443, in order to keep the previous expected behavior.

### DIFF
--- a/src/check_https_domains.py
+++ b/src/check_https_domains.py
@@ -57,6 +57,10 @@ def parse_args():
              'retrieved from.'
     )
     parser.add_argument(
+        '-p', dest='port', required=False, default=443, type=int,
+        help='Port on the host where Nagios will connect to. Defaults to 443.'
+    )
+    parser.add_argument(
         '-d', dest='domains', required=True,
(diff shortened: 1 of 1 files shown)


src/check_https_domains.py
```diff
@@ -56,6 +56,10 @@ def parse_args():
         help='IP of the host. The address where the certificate will be '
              'retrieved from.'
     )
+    parser.add_argument(
+        '-p', dest='port', required=False, default=443, type=int,
```
TheFRedFox: Instead of having `dest='port'`, we could also provide the possibility of `--port`. This will be then used as destination.
`required=False` is default. It is well known enough to not pass it explicitly.
Also the order of `type=int, default=443` makes more sense. First provide the type, then the default.
For all, no strong opinion. Up to you to change it or not.
```suggestion
        '-p', '--port', type=int, default=443,
```

lamaral: They make total sense. I'll change it :)

PR #188, opened 2020-03-19
Add ipython's cache to ignored files

Felixoid:


.gitignore
```diff
@@ -68,4 +69,4 @@ target/
 .ropeproject
 
 #VSCode
-.vscode/
\ No newline at end of file
+.vscode/
```
lamaral: I would say iPython is fine, but vscode should really be in your personal global gitignore, not in the project gitignore.

TheFRedFox: He didn't touch vscode, it was just about the new line. Therefore fine with me. We could blame the one who put vscode in the beginning. ^^

Felixoid: Yep, it's vim added the new line in EOF. Though, don't see a reason to delete it

TheFRedFox: One could argue about this that the RFC is not about cleanup and adding new line for consistency is cleanup. However, now we are getting even too picky for me as it would bloat up the workflow quite much. 😉

PR #190, opened 2020-03-27
Improve cluster check in clickhouse plugin

Felixoid:
The tiny change in `SELECT *` query improves performance almost 4 times.

### DIFF
--- a/src/check_clickhouse.py
+++ b/src/check_clickhouse.py
@@ -212,7 +212,7 @@ class CheckClusters(Check):
         clusters = self.execute_dict(
             r'''
             SELECT
-                splitByChar('\'', engine_full)[2] AS cluster,
+                splitByChar('\'', t.engine_full)[2] AS cluster,
                 groupArray(concat(t.database, '.', t.name)) AS tables
             FROM system.tables AS t
             INNER JOIN
@@ -238,22 +238,10 @@ class CheckClusters(Check):
             self.code.current = Code.CRITICAL
             messages.append('Clusters not found on server: {}'.format(missing))
 
-        def check_tables(tables: List[str]) -> List[str]:
-            failed = []
-            for t in tables:
-                logger.debug('Select from distributed table `{}`'.format(t))
-                try:
-                    self.execute('SELECT 1 FROM {} LIMIT 1'.format(t))
-                except ServerException as e:
-                    logger.warning('Fail to read from `{}`, exception is: {}'
-                                   .format(t, e.message))
-                    failed.append(t)
-            return failed
-
         for cl in clusters:
             logger.debug('Tables for cluster `{cluster}` are {tables}'
                          .format_map(cl))
-            failed_tables = check_tables(cl['tables'])
+            failed_tables = self._check_tables(cl['tables'])
             if not failed_tables:
                 continue
             self.code.current = Code.CRITICAL
@@ -264,6 +252,18 @@ class CheckClusters(Check):
 
         messages = messages or ['All clusters are fine: {}'.format(clusters)]
         return self.exit('; '.join(messages))
+
+    def _check_tables(self, tables: List[str]) -> List[str]:
+        failed = []
+        for t in tables:
+            logger.debug('Select from distributed table `{}`'.format(t))
+            try:
+                self.execute('SELECT 1 FROM {} LIMIT 1'.format(t))
+            except ServerException as e:
+                logger.warning('Fail to read from `{}`, exception is: {}'
+                               .format(t, e.message))
+                failed.append(t)
+        return failed
 
 
 class CheckParts(Check):
--- a/src/check_clickhouse.py
+++ b/src/check_clickhouse.py
@@ -25,7 +25,7 @@ from clickhouse_driver.errors import (  # type: ignore  # missing stubs
     NetworkError, ServerException,
 )
 from pprint import pformat
-from typing import Tuple, List, Union, Set
+from typing import List, Optional, Tuple, Union
 
 import logging
 import sys
@@ -198,27 +198,7 @@ class Check(object):
 
 
 class CheckClusters(Check):
-    # Here are ClickHouse data types that allow to optimize query by
-    # partition key `WHERE` clause
-    defaults_per_type_family = (
-        ({'Array'},
-         '[]'),
-        ({'DEC', 'Decimal', 'Decimal32', 'Decimal64', 'Decimal128',
-          'DOUBLE', 'FLOAT', 'Float32', 'Float64',
-          'BIGINT', 'INT', 'INTEGER', 'SMALLINT', 'TINYINT',
-          'Int8', 'Int16', 'Int32', 'Int64',
-          'UInt8', 'UInt16', 'UInt32', 'UInt64'
-          'IPv4'},
-         0),
-        ({'BINARY', 'BLOB', 'CHAR', 'LONGBLOB', 'LONGTEXT', 'MEDIUMBLOB',
-          'MEDIUMTEXT', 'TEXT', 'TINYBLOB', 'TINYTEXT', 'VARCHAR',
-          'FixedString', 'String',
-          'IPv6'},
-         "''"),
-        ({'Date', 'DateTime', 'DateTime64', 'TIMESTAMP'},
-         'now()',)
-    )  # type: Tuple[Tuple[Set[str], Union[int, str]], ...]
-    LocalTable = Union[Tuple[str, str], Tuple[()]]
+    LocalTable = Optional[Tuple[str, str]]
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -309,7 +289,7 @@ class CheckClusters(Check):
         failed = []
         # If there're no local_tables (the cluster is remote),
         # then we create tuple of empty tuples with the len of tables
-        local_tables = local_tables or ((),) * len(tables)
+        local_tables = local_tables or (None,) * len(tables)
         for t, lt in zip(tables, local_tables):
             logger.debug('Select from Distributed() table `{}`,'
                          ' local table is {}'.format(t, lt))
@@ -324,9 +304,9 @@ class CheckClusters(Check):
     def _optimized_request(self, table: str,
                            local_table: LocalTable) -> str:
         """
-        This method is trying to optimize request by adding 'WHERE' clause
-        with partitioning key. In the best case WHERE won't match anything
-        and SELECT query will finish almost instantly. Here the next steps are done:
+        This method is trying to optimize request by adding 'WHERE' clause with
+        partitioning key. In the best case WHERE won't match anything and the
+        request will finish almost instantly. Here the next steps are done:
             - See if the current host is the part of the checked cluster
             (local_tables is not empty)
             - Get columns in partitioning key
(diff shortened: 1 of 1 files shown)


Felixoid:
Here are results of hyperfine benchmark for the cluster check before PR (1), after changing `SELECT *` to `SELECT 1`, and after the latest commit with generating `WHERE` clause with partition key (3)

```
Benchmark #1: ./check_clickhouse.py clusters -c graphite_data
  Time (mean ± σ):      3.489 s ±  1.081 s    [User: 171.7 ms, System: 35.7 ms]
  Range (min … max):    2.812 s …  6.085 s    10 runs

Benchmark #2: ./check_clickhouse.py clusters -c graphite_data
  Time (mean ± σ):      1.178 s ±  0.101 s    [User: 197.7 ms, System: 38.0 ms]
  Range (min … max):    1.069 s …  1.329 s    10 runs

Benchmark #3: ./check_clickhouse.py clusters -c graphite_data
  Time (mean ± σ):     370.7 ms ±  24.7 ms    [User: 162.8 ms, System: 26.3 ms]
  Range (min … max):   339.7 ms … 425.0 ms    10 runs
```

Note: we have only 1(!) table in graphite_data cluster, so the time is mostly spent on the `SELECT * FROM {} LIMIT 1` query and must be optimized

src/check_clickhouse.py
```diff
@@ -196,64 +198,57 @@ def check_config(self, config: dict, keys: set):
...
+        ('DEC', 'Decimal', 'Decimal32', 'Decimal64', 'Decimal128',
+         'DOUBLE', 'FLOAT', 'Float32', 'Float64',
+         'BIGINT', 'INT', 'INTEGER', 'SMALLINT', 'TINYINT',
+         'Int8', 'Int16', 'Int32', 'Int64',
+         'UInt8', 'UInt16', 'UInt32', 'UInt64'
+         'IPv4'): 0,
+        ('BINARY', 'BLOB', 'CHAR', 'LONGBLOB', 'LONGTEXT', 'MEDIUMBLOB',
+         'MEDIUMTEXT', 'TEXT', 'TINYBLOB', 'TINYTEXT', 'VARCHAR',
+         'FixedString', 'String',
+         'IPv6'): '\'\'',
```
kofrezo: [...] When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.
[See PEP8](https://www.python.org/dev/peps/pep-0008/#string-quotes)

src/check_clickhouse.py
```diff
@@ -96,8 +96,10 @@ def add_subparser_clusters(subparsers: _SubParsersAction):
     )
     parser_clusters.set_defaults(check=CheckClusters)
     parser_clusters.add_argument(
-        '-c', '--clusters', action='append', default=[], required=True,
-        help='cluster names, could be defined multiple times',
+        '-c', '--clusters', action='append', default=[],
+        help='cluster names that MUST be presented on the server. '
+        'If the argument is omitted, then only clusters with Distributed '
```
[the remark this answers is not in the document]
Felixoid: This is an engine name https://clickhouse.tech/docs/en/operations/table_engines/distributed/

src/check_clickhouse.py
Felixoid: Thank you for the reference, I've lacked it

src/check_clickhouse.py
```diff
@@ -196,64 +198,57 @@ def check_config(self, config: dict, keys: set):
 
 
 class CheckClusters(Check):
+    # Here are ClickHouse data types that allow to optimize query by
+    # partition key `WHERE` clause
+    defaults_per_type_family = {
+        ('Array'): '[]',
+        ('DEC', 'Decimal', 'Decimal32', 'Decimal64', 'Decimal128',
```
TheFRedFox: Please check, if it is possible to use `set`s here. It would make the lookup with `in` faster.

src/check_clickhouse.py
```diff
@@ -262,9 +257,111 @@ def check_tables(tables: List[str]) -> List[str]:
...
+            - Get columns in partitioning key
+            - Add `WHERE` clause with compatible types, see
+            CheckClusters.defaults_per_type_family
+        """
+        query = r'SELECT 1 FROM {table} {where} LIMIT 1'
+        f_config = {'table': table, 'where': ''}
+        if not local_table:
+            return query.format_map(f_config)
+
+        p_keys = self.execute_dict(
```
TheFRedFox: In general be careful with such prefixes. I understand that the `p_` stands most probably for `partition`, however it is just clear **after** reading the whole code.
In general rather use explicit naming than implicit, so `partition_keys` is longer and much much more explicit/understandable. A good balance is not always easy to find and can also be used as an indicator to find new or separate scopes.
For here: IMO `partition_keys` is still short enough and much more readable than `p_keys`.

src/check_clickhouse.py
```diff
@@ -196,64 +198,57 @@ def check_config(self, config: dict, keys: set):
...
         super().__init__(*args, **kwargs)
 
     def __call__(self, config: dict) -> ExitStruct:
         """
-        Accepts args with 'clusters' attribute and checks every related
+        Accepts args with `clusters` attribute and checks every related
         Distributed table, if it is readable
+
+        If config['clusters'] is an empty list, than checks every presented
+        Distributed table
```
TheFRedFox:
```suggestion
        `Distributed`-engined  table.
```
```suggestion
        Distributed-engined table.
```

src/check_clickhouse.py
```diff
@@ -196,64 +198,57 @@ def check_config(self, config: dict, keys: set):
...
+    }
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def __call__(self, config: dict) -> ExitStruct:
         """
-        Accepts args with 'clusters' attribute and checks every related
+        Accepts args with `clusters` attribute and checks every related
         Distributed table, if it is readable
```
TheFRedFox:
```suggestion
        `Distributed`-engined table, if it is readable
```
```suggestion
        Distributed-engined table, if it is readable
```

src/check_clickhouse.py
```diff
@@ -196,64 +198,57 @@ def check_config(self, config: dict, keys: set):
...
-            ''',
-            {'clusters': tuple(config['clusters'])},
-        )
+        clusters = self._get_clusters(config['clusters'])
         logger.debug(
             'Existing clusters on ClickHouse server are: {}'
             .format(pformat(clusters))
         )
 
+        # Check if any of config['clusters'] doesn't have Distributed tables
```
TheFRedFox:
```suggestion
        # Check if any of config['clusters'] doesn't have Distributed-engined tables
```

src/check_clickhouse.py
```diff
@@ -262,9 +257,111 @@ def check_tables(tables: List[str]) -> List[str]:
...
         return self.exit('; '.join(messages))
 
+    def _get_clusters(self, clusters: List[str]) -> List[dict]:
+        """
+        Returns list of dicts with the following keys:
+            - cluster: cluster name
+            - tables: tables belongs to cluster
+            - local_tables: if the checking node belongs to cluster, then this
+                array contains tuples of (database, table) for tables behind
+                the Distributed tables
```
TheFRedFox:
```suggestion
                the Distributed-engined tables
```

src/check_clickhouse.py
```diff
@@ -262,9 +257,111 @@ def check_tables(tables: List[str]) -> List[str]:
...
+            return self.execute_dict(query)
+
+    def _check_tables(self, tables: List[str],
+                      local_tables: List[list]) -> List[str]:
+        failed = []
+        # If there're no local_tables (the cluster is remote),
+        # then we create list of empty lists with the len of tables
+        local_tables = local_tables or [[]] * len(tables)
+        for t, lt in zip(tables, local_tables):
+            logger.debug('Select from distributed table `{}`,'
```
TheFRedFox:
```suggestion
            logger.debug('Select from Distributed-engined table `{}`,'
```
?

src/check_clickhouse.py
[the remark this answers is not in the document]
TheFRedFox: Then I would suggest:
```suggestion
        'If the argument is omitted, then only clusters with Distributed-engined '
```

src/check_clickhouse.py
```diff
@@ -265,6 +251,107 @@ def check_tables(tables: List[str]) -> List[str]:
...
+            logger.debug('Select from distributed table `{}`'.format(t))
+            try:
+                self.execute(self._optimized_request(t, lt))
+            except ServerException as e:
+                logger.warning('Fail to read from `{}`, exception is: {}'
+                               .format(t, e.message))
+                failed.append(t)
+        return failed
+
+    def _optimized_request(self, table: str, local_table: list) -> str:
```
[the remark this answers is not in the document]
Felixoid:
```
In [8]: execute_dict(query)
Out[8]:
[{'cluster': 'graphite_data',
  'tables': ('graphite.data',),
  'local_tables': (('graphite', 'data_lr'),)}]
```

fixed, since it proceeds with a tuple in any case

src/check_clickhouse.py
```diff
@@ -265,6 +251,107 @@ def check_tables(tables: List[str]) -> List[str]:
...
+        """
+        This method is trying to optimize request by adding 'WHERE' clause
+        with partitioning key. In the best ever case it won't match anything
+        and SELECT won't return anything. Here the next steps are done:
+        - See if the current host is the part of the checked cluster
+        - Get columns in partitioning key
+        - Add
+        """
+        query = r'SELECT 1 FROM {table} {where} LIMIT 1'
+        f_config = {'table': table, 'where': ''}
```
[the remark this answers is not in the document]
Felixoid: Changed to `query_format`

src/check_clickhouse.py
```diff
@@ -265,6 +251,107 @@ def check_tables(tables: List[str]) -> List[str]:
...
+        return failed
+
+    def _optimized_request(self, table: str, local_table: list) -> str:
+        """
+        This method is trying to optimize request by adding 'WHERE' clause
+        with partitioning key. In the best ever case it won't match anything
+        and SELECT won't return anything. Here the next steps are done:
+        - See if the current host is the part of the checked cluster
+        - Get columns in partitioning key
+        - Add
```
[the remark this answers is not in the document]
Felixoid: Was fixed in the following commit

src/check_clickhouse.py
```diff
@@ -265,6 +251,107 @@ def check_tables(tables: List[str]) -> List[str]:
...
+                FROM system.clusters
+                WHERE cluster IN %(clusters)s
+                GROUP BY cluster
+            ) AS c USING (cluster)
+            WHERE t.engine = 'Distributed'
+            GROUP BY cluster
+            ''',
+            {'clusters': tuple(clusters)},
+        )
+        return data
```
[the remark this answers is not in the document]
Felixoid: already does

src/check_clickhouse.py
Felixoid: I've changed to the tuple of couples. I wasn't happy with dict, now it looks a bit more proper

src/check_clickhouse.py
```diff
@@ -223,7 +223,7 @@ def __call__(self, config: dict) -> ExitStruct:
         Accepts args with `clusters` attribute and checks every related
         Distributed table, if it is readable
 
-        If config['clusters'] is an empty list, than checks every presented
+        If config['clusters'] is an empty list, then checks every presented
```
TheFRedFox:
```suggestion
        If config['clusters'] is an empty list it checks every presented
```

src/check_clickhouse.py
```diff
@@ -320,17 +320,20 @@ def _optimized_request(self, table: str, local_table: list) -> str:
         """
         This method is trying to optimize request by adding 'WHERE' clause
         with partitioning key. In the best case WHERE won't match anything
-        and SELECT won't return anything. Here the next steps are done:
+        and SELECT will over almost immediately. Here the next steps are done:
```
TheFRedFox: 'SELECT will over almost immediately' .. there is something wrong with the sentence. Not sure, what you meant, maybe:
```suggestion
        and SELECT will be over almost everything. Here the next steps are done:
```
?

src/check_clickhouse.py
```diff
@@ -329,13 +335,14 @@ def _optimized_request(self, table: str, local_table: list) -> str:
         """
 
         query = r'SELECT 1 FROM {table} {where} LIMIT 1'
-        f_config = {'table': table, 'where': ''}
-        unoptimized_query = query.format_map(f_config)
+        query_format = {'table': table, 'where': ''}
```
TheFRedFox: `query_format` is definitely better. I could also think of `query_config` as `config` is actually quite suitable for it.

src/check_clickhouse.py
```diff
@@ -299,12 +304,12 @@ def _get_clusters(self, clusters: List[str]) -> List[dict]:
...
-    def _check_tables(self, tables: List[str],
-                      local_tables: List[list]) -> List[str]:
+    def _check_tables(self, tables: Tuple[str, ...],
+                      local_tables: Tuple[LocalTable, ...]) -> List[str]:
         failed = []
         # If there're no local_tables (the cluster is remote),
-        # then we create list of empty lists with the len of tables
-        local_tables = local_tables or [[]] * len(tables)
+        # then we create tuple of empty tuples with the len of tables
+        local_tables = local_tables or ((),) * len(tables)
```
TheFRedFox: Btw, to fill it with `None`s is not possible? That would be more suitable.

PR #193, opened 2020-04-06
Add check for the oldest message in queue

Felixoid:
Before these changes, if there's only one message, but from days ago, monitoring did stay silent.

Now bot issues with `too old message` and `too many messages` in the queue are checked.

### DIFF
--- a/src/check_exim_delivery.sh
+++ b/src/check_exim_delivery.sh
@@ -24,19 +24,33 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


Felixoid:
@kofrezo is it better now?

kofrezo:
Yes. Now it makes sense to. Approval remains.

TheFRedFox:
I'm out here. I trust in the colleagues measurements. 🙂

PR #194, opened 2020-04-09
Add a special "diff" keyword to make checks based on absolute ACKs

Woellchen:


src/check_rabbitmq.py
```diff
@@ -216,10 +229,19 @@ def __call__(self, data):
...
+            # We also support a special keyword "diff" that calculates the
+            # actual sample difference.
+            if part == 'diff':
+                if 'samples' not in data:
+                    raise ValueError('No samples are found to calculate diff')
 
-            data = data[part]
+                # Ignore the oldest sample. Check _build_url for the reasoning.
+                samples = data['samples']
+                data = samples[0]['sample'] - samples[-2]['sample']
```
TheFRedFox, line 240: How likely is it that `samples` is empty or not having the wanted entries and what will happen if so?

Woellchen: it will not happen because thats how the rabbitmq stats api works. below in `_build_url` i pass the appropriate values so that it will always return 4 samples. if that stops working then the rabbitmq api had a breaking change

Woellchen:
thanks @lamaral for the updated comment and the simpler logic :)

src/check_rabbitmq.py
TheFRedFox: Possibilities are that script will either break then, or it will provide wrong data. `samples` with 3 entries, `samples[-2]` would actually take `samples[0]`, so in the end it would be `samples[0]['sample'] - samples[0]['sample']` which I guess would just provide wrong data.

Woellchen: in your case the result would be zero. and you would notice when suddenly your checks go critical right? or do you want me to check the length and raise?

Woellchen: you are right, ill just add an exception

TheFRedFox: Question was first "how likely" and "what" will happen. Question now would be if we can live with that. I am not using the script, I am just making aware of that possible problem. ;)

Woellchen: well. given the fact that i had to introduce this "hack" to get reliable data out of the api it makes total sense to be skeptical about this

PR #195, opened 2020-04-16
fix(check_plugin_vulnerability): add cookies to requests

seqizz:


src/check_plugin_vulnerability.py
```diff
@@ -83,20 +81,20 @@ def fetch_csrf_token(host, auth):
     if len(crumb) != 2:
         raise KeyError("Wrong number of tokens for csrf response")
 
-    return {crumb[0]: crumb[1]}
+    return {crumb[0]: crumb[1]}, response.cookies
```
TheFRedFox, line 84: What's the reason to return all cookies and not the necessary ones?

seqizz: I am not sure what is necessary to what degree. Some said JSESSIONID is needed and some advised sending the whole cookie jar. I've tested sending only that one but didn't work (maybe I did something wrong) but instead of digging and theorizing just sent the whole thing. Does it harm?

TheFRedFox: It should not harm. It's just destroying the scope of the method being called `fetch_csrf_token` as it is now also getting most probably any authentication cookies or the like. In general, I would also prefer to know what we actually need.
3 possibilities to proceed:
1. Checking what we actually need and just returning it.
2. Renaming the function to the scope fits again.
3. Leaving it like that with the comment that we don't know what we actually need, to give it a context, why we are returning everything right now and that it might not be needed like that.

seqizz: Renamed with last push + fixed some trivial pep stuff.

src/check_plugin_vulnerability.py
```diff
@@ -83,20 +79,20 @@ def fetch_csrf_token(host, auth):
...
 
 
-def fetch_plugin_issues(url, script, headers, auth):
-    res = post_script_data(url, script, headers, auth)
-    if res.status_code == 403:
-        raise CsrfNeededError(response=res)
+def fetch_plugin_issues(url, script, headers, auth, cookies):
+    res = post_script_data(url, script, headers, auth, cookies)
+    if res.status_code != 200:
+        print('Unknown - Request failed: {}'.format(res.text))
```
TheFRedFox, line 88: What makes this `print` being needed additionally to the `res.raise_for_status()` 2 lines below?

seqizz: Good catch, I am not raising/catching stuff anymore. If it's not 200, I intend to bail out, so added new exit line.

TheFRedFox: That's ok, however, it is bad practice to exit inside utility function resp outside the `main` function. Rather let this function raise an error and let the `main` function catch it and exit accordingly.

seqizz: While it's true on importable functions and bigger programs, in this case I'd disagree. This is system.exit(): A decision that I don't want this binary to do any further action after printing what I need.
Will check Monday :+1:

TheFRedFox: It's still bad practice. You can decide to ignore it. The disadvantages are that the script has several points to exit hard, which is hard to maintain. The utility function is also hiding that it is exiting hard, furthermore a utility function should never exit on it's own but let the user of the function decide what to happen. Like that this function can also directly be reused. And for python every function is importable.
What I am asking for is to raise an Error her, e.g. just a HTTPError with this exact message maybe and catch it in the `main` function to `exit` there. Advantages then is that every possible `exit` is handled inside the `main` function, so just in one single place. Like this it is way easier to maintain in the future.
Think about the scope of the function and not the scope for a single scenario.
What's for you against making the responsibility of functions clear and let the `main` function handle the exits and not some utility function?

seqizz: I am not against it, I already told it's true. Just saying this whole thing should be simple enough. I hope my last push tells what I mean.
I can revert and add the exception, but at least for me this is way easier to maintain.

TheFRedFox: Your last change is giving the responsibility of exiting back from the utility function to the `main` function, as it calls the `print_and_exit` function and the naming is clear enough to mark it as having this responsibility and the function is still designed to be reusable. This is absolutely ok like that, as the responsibilities are clear now again. 👍

src/check_plugin_vulnerability.py
```diff
@@ -49,30 +47,32 @@ def main():
...
-    pass
-
-
-def print_and_exit(plugins):
-    if not plugins:
+def print_and_exit(result):
+    if result.status_code != 200:
+        print('Unknown - Request failed: {}'.format(result.text))
+        exit(1)
+    plugin_list = result.json()
```
TheFRedFox, line 61: As it was also not checking with the former version, I see it as out-of-scope for this PR. Just for the future, if the response body is not `json` then `result.json()` would raise an Error here. If it will happen in the future, it might be interesting to catch it.

seqizz: Added `TODO` line for it, surprisingly force-pushing didn't remove your approval :thinking: :shield:

TheFRedFox: It's a setting and I don't think we need it. The one who can merge PRs should be responsible enough to check for new commits and to tell the reviewers to check again. Feel free to revoke or request my review again if you need it again. ;)

PR #196, opened 2020-04-16
Allow to ignore table on check_mysql_table_status

meikomeik:
Add the ability to ignore certain tables via --ignore-tables argument.

### DIFF
--- a/src/check_mysql_table_status.py
+++ b/src/check_mysql_table_status.py
@@ -56,11 +56,15 @@ def parse_arguments():
     parser.add_argument('--modes', type=options, default=DEFAULT_MODES)
     parser.add_argument('--warnings', type=options, help='warning limits')
     parser.add_argument('--criticals', type=options, help='critical limits')
-    parser.add_argument('--tables', type=options, help='show selected tables')
     parser.add_argument('--perf', action='store_true', help='performance data')
     parser.add_argument('--avg', action='store_true', help='show averages')
     parser.add_argument('--max', action='store_true', help='show maximums')
     parser.add_argument('--min', action='store_true', help='show minimums')
+
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--tables', type=options, help='show selected tables')
+    group.add_argument('--ignore-tables', nargs='*',
+                       help='ignore selected tables')
 
     return parser.parse_args()
 
@@ -93,7 +97,8 @@ def main():
         arguments.perf,
     )
     messages = get_messages(
-        database, arguments.tables, arguments.modes, list(outputs)
+        database, arguments.tables, arguments.ignore_tables, arguments.modes,
+        list(outputs)
     )
     joined_message = join_messages(**messages)
 
@@ -123,10 +128,12 @@ def get_outputs(output_classes, modes, warnings, criticals, perf):
             yield output_class(mode, warning_limit, critical_limit, perf)
 
 
-def get_messages(database, filter_tables, attributes, outputs):
+def get_messages(database, filter_tables, ignore_tables, attributes, outputs):
     """Check all tables for all output instances"""
     for table, values in database.get_table_values(attributes):
-        if filter_tables and not any(t == table for t in filter_tables):
+        if filter_tables and table not in filter_tables:
+            continue
+        elif ignore_tables and table in ignore_tables:
             continue
         for output in outputs:
             attribute = output.attribute


src/check_mysql_table_status.py
```diff
@@ -123,10 +126,12 @@ def get_outputs(output_classes, modes, warnings, criticals, perf):
             yield output_class(mode, warning_limit, critical_limit, perf)
 
 
-def get_messages(database, filter_tables, attributes, outputs):
+def get_messages(database, filter_tables, ignore_tables, attributes, outputs):
     """Check all tables for all output instances"""
     for table, values in database.get_table_values(attributes):
-        if filter_tables and not any(t == table for t in filter_tables):
+        if filter_tables and table not in filter_tables:
```
axxelG, line 132: if you set filter_tables and ignore_tables ignore_tables has no effect

src/check_mysql_table_status.py
```diff
@@ -57,6 +57,8 @@ def options(value):
     parser.add_argument('--warnings', type=options, help='warning limits')
     parser.add_argument('--criticals', type=options, help='critical limits')
     parser.add_argument('--tables', type=options, help='show selected tables')
+    parser.add_argument('--ignore-tables', type=options,
```
TheFRedFox, line 60: Historical question:
What's the reason to use that custom `options` type and not `nargs='*'`. Change would be to provide it like that: `... --ignore-tables table1 table2` and not `... --ignore-tables table1,table2`

src/check_mysql_table_status.py
TheFRedFox: `ignore_tables` just has no effect if table is *not* in `filter_tables`. I guess it is on purpose that we don't get further if the table is not in `filter_tables`. If the table is in `filter_tables` the `ignore_tables` is processed too.
Guess @meikomeik can tell more for the purpose.

src/check_mysql_table_status.py
meikomeik: The historic context is that Emre cloned this from somewhere on the internet around 2017 and since then it didn't got much attention. I came here to add this --ignore-tables thingy

meikomeik: I can look into changing it. Sounds like a good improvement and the change is rather minior.

src/check_mysql_table_status.py
axxelG: Ah, --tables goes to filter_tables. Than I need to change my comment. I read filter_tables as tables that will be filtered out. Maybe this parameter should be renamed to tables to keep it consistent with the parameter name.

TheFRedFox: Normally you use `filter` in English to keep them. I guess the confusion comes, that we mostly use `rausfiltern` in German, which then gets associated like `filter` is to get rid of something. So, here it is actually correct and the confusion might come from German usage.

src/check_mysql_table_status.py
TheFRedFox: Fair enough. Might be interesting in the future to change it to native `nargs` as this is also more safe to use as it already trims whitespaces.

meikomeik: I was able to change it for the --tables and --ignore-tables parameter as these are currently not in use. The other ones I can not easily change as this wierd options syntax is used in production.

Here is an example on how the script is currently called:

> check_mysql_table_status.py --modes rows,data_free,auto_increment --warnings 1G,1750M,80%

src/check_mysql_table_status.py
meikomeik: Setting both filter_tables and ignore_tables doesn't make much sense as they are the opposite of each other. I just left the old code to "whitelist" tables in and added the ability to ignore tables. Unsure if I should change something now or leave it like it is?!

src/check_mysql_table_status.py
TheFRedFox: How sure are you that `--tables` is not used anywhere yet? For now I am cool with using `options` to keep consistent here for now.

src/check_mysql_table_status.py
TheFRedFox: If it should not be possible to set both, we might set those arguments as `mutual_exclusive` ones: https://docs.python.org/3/library/argparse.html#mutual-exclusion and then changing the code to have an `if-elseif` so it is clear code-wise that just one is getting checked.

However, it's not up to me to decide if it should be mutual-exclusive or if there are cases where both will make sense.

meikomeik: I now made the --tables / --ignore-tables mutual exclusive.

I also changed the if to else if but I'm not sure if this is really necessary. It looks wierd to me to do an elseif at that place. Two ifs would be shorter that if/else if

src/check_mysql_table_status.py
TheFRedFox: Fair enough for me. Up to you if you wanna change it already for `--ignore-tables` or make it `type=options` for now to be consistent.

src/check_mysql_table_status.py
TheFRedFox: The `elif` is to show that it is mutual exclusive by design (also the 2nd `if` won't get processed). If we are not 100% sure about it we should leave it with 2 `if`s. Both is fine for me. Decision is up to you.

src/check_mysql_table_status.py
meikomeik: Thank you, I'll keep it like this

src/check_mysql_table_status.py
meikomeik: I'll leave the elif in as they are mutualy exclusive. Thank you

PR #197, opened 2020-04-21
check_cron_runtime: add exclude file support

seqizz:
Also removed hardcoded `am_stockpile_distribution` cron which I believe we don't need

### DIFF
--- a/src/check_cron_runtime.py
+++ b/src/check_cron_runtime.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """InnoGames Monitoring Plugins - Cron Runtime Check
 
-Copyright (c) 2018 InnoGames GmbH
+Copyright (c) 2020 InnoGames GmbH
 """
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the 'Software'), to deal
@@ -24,11 +24,20 @@ Copyright (c) 2018 InnoGames GmbH
 from subprocess import Popen, PIPE, STDOUT
 from argparse import ArgumentParser
 from sys import exit
+from os.path import isfile
 
-# A list of cron names that shall get ignored by this check.
-exclude_list = [
-    'am_stockpile_distribution',
-]
+
+def parse_exclude_file(exclude_file):
+    """Parses the exclude file and and adds given list
+
+    Returns a combined list
+    """
+
+    if not isfile(exclude_file):
+        return []
+
+    with open(exclude_file) as fd:
+        return list(filter(None, [line.rstrip('\n') for line in fd]))
 
 
 def parse_args():
@@ -49,10 +58,16 @@ def parse_args():
         action='store_true',
         help='Increase verbosity',
     )
+    parser.add_argument(
+        '--exclude-file',
+        default='',
+        help='File to get list of cron names to exclude (line separated)',
+    )
+
     return vars(parser.parse_args())
 
 
-def main(verbose=False):
+def main(verbose=False, exclude_file=''):
     """The main program"""
 
     # Dictionary of pids containing th command that is running
@@ -91,6 +106,8 @@ def main(verbose=False):
     cron_child_list = []
     cron_child_time = None
 
+    exclude_list = parse_exclude_file(exclude_file)
+
     for child in cron_childs.split('\n'):
         cron_child_list.append(child)
 
@@ -98,7 +115,7 @@ def main(verbose=False):
         print('Cron Childs: ' + str(cron_child_list))
 
     cron_pid_more_than1 = []
-    ok_flag = warn_flag = 0
+    warn_flag = 0
     for pid in cron_child_list:
         if pid:
             try:
@@ -111,7 +128,7 @@ def main(verbose=False):
                 continue
 
             # check for 1 hour time
-            if int(cron_child_time) > 3600:
+            if int(cron_child_time) >= 3600:
                 # Check if cron name is on ignore list
                 include = True
                 child_pid = execute('pgrep -P ' + pid).strip()
@@ -124,7 +141,7 @@ def main(verbose=False):
                     cron_pid_more_than1.append(pid)
                     warn_flag += 1
             elif int(cron_child_time) < 3600:
-                ok_flag = 1
+                continue
             else:
                 print('UNKNOWN PID execution TIME')
                 exit(int(un_code))
@@ -136,7 +153,7 @@ def main(verbose=False):
         for i in cron_pid_more_than1:
             get_child(i)
         exit(int(warn_code))
-    if ok_flag:
+    else:
         print('No CRON running more than 1 hour')
         exit(int(ok_code))
 


src/check_cron_runtime.py
```diff
@@ -24,11 +24,25 @@
...
+
+    Returns a combined list
+    """
+
+    if isfile(exclude_file):
+        with open(exclude_file) as fd:
+            excluded_crons = (
+                [i.rstrip('\n') for i in fd.readlines()]
+            )
+    return excluded_crons
```
TheFRedFox, line 45: `excluded_crons` is not defined if `exclude_file` is not a file.

src/check_cron_runtime.py
```diff
@@ -24,11 +24,25 @@
...
 from argparse import ArgumentParser
 from sys import exit
+from os.path import isfile
 
-# A list of cron names that shall get ignored by this check.
-exclude_list = [
-    'am_stockpile_distribution',
-]
+
+# A list of cron names that shall get ignored by this check will go into here
```
TheFRedFox, line 30: The comment is misleading:
```suggestion
# Path to file with a list of cron names (line separated) that shall get ignored by this check
```

src/check_cron_runtime.py
```diff
@@ -24,11 +24,25 @@
...
+exclude_file = '/etc/check_cron_runtime_exclude'
+
+
+def parse_exclude_file():
+    """Parses the exclude file and and adds given list
+
+    Returns a combined list
+    """
+
+    if isfile(exclude_file):
```
TheFRedFox, line 40: What if not? What if the file does not exist at all? Will there be a fallback for it?

src/check_cron_runtime.py
```diff
@@ -24,11 +24,25 @@
...
 from sys import exit
+from os.path import isfile
 
-# A list of cron names that shall get ignored by this check.
-exclude_list = [
-    'am_stockpile_distribution',
-]
+
+# A list of cron names that shall get ignored by this check will go into here
+exclude_file = '/etc/check_cron_runtime_exclude'
```
TheFRedFox, line 31: What's the reason for not already going and making it configurable via arguments?

src/check_cron_runtime.py
```diff
@@ -24,11 +24,25 @@
...
+    """Parses the exclude file and and adds given list
+
+    Returns a combined list
+    """
+
+    if isfile(exclude_file):
+        with open(exclude_file) as fd:
+            excluded_crons = (
+                [i.rstrip('\n') for i in fd.readlines()]
+            )
```
TheFRedFox, line 44:
```suggestion
            excluded_crons = (
                [line.rstrip('\n') for line in fd]
            )
```
`fd.readlines()` will read the whole file and put it in memory. Although, I don't believe that we need to worry about that, we also don't need to worry that we have everything in memory. Using `for line in fd` will use `fd` as a generator, having just one line in memory all the time.

Adding `filter` to it would also make it empty line safe:
```suggestion
            excluded_crons = (
                list(filter(None, [line.rstrip('\n') for line in fd]))
            )
```
or sticking to comprehensions and generators:
```suggestion
            excluded_crons = (
                line for line in (l.rstrip('\n') for l in fd) if line
            )
```

src/check_cron_runtime.py
```diff
@@ -91,6 +105,9 @@ def main(verbose=False):
     cron_child_list = []
     cron_child_time = None
 
+    # Parse exclude list
```
TheFRedFox, line 108: Unnecessary comment. The function name is already meaningful enough.

TheFRedFox:
What will happen if `am_stockpile_distribution` is still there and we will merge this change and resp. what needs to be done then?

src/check_cron_runtime.py
seqizz: Some checks use one method, some other. Not a real reason but I simply didn't want to change the monitoring check itself after this change. If you think this is worth it for some reason please reopen :thinking:

src/check_cron_runtime.py
seqizz: Hmm TIL filter, thanks.

seqizz:
> What will happen if `am_stockpile_distribution` is still there and we will merge this change and resp. what needs to be done then?

Since this check doesn't kill or do something invasive, they'll just see a warning on nagios. In that case it can be added to exclude file via project's puppet.

src/check_cron_runtime.py
TheFRedFox: We should go for arguments for all scripts in the future. Especially a consistent way and arguments is then the best option. I see a huge benefit already to make that an argument. If you don't want to change the way this script works, you can set `/etc/check_cron_runtime_exclude` as default value for now. Then the script works as normal now and we still have the change to configure the path of the file freely.
In general I would say for this script it also might make sense to add a `nargs` argument to pass in a list of cron names, like:
`script.py --exclude am_stockpile_distribution some_other_cron_name`

src/check_cron_runtime.py
```diff
@@ -24,11 +24,28 @@
...
+    """
+
+    excluded_crons = []
+
+    if isfile(exclude_file):
+        with open(exclude_file) as fd:
+            for cron in list(filter(None, [line.rstrip('\n') for line in fd])):
+                excluded_crons.append(cron)
+
+    return excluded_crons
```
TheFRedFox, line 48: Actually we can make it shorter, as we now already get the list we want:
```suggestion
    if not isfile(exclude_file):
        return []
    
    with open(exclude_file) as fd:
            return list(filter(None, [line.rstrip('\n') for line in fd]))
```

src/check_cron_runtime.py
seqizz: Ah, but I don't like the style of last example. That's how we get `check_logrotate_missing` in our environment; single line but 700+ character checks..
But I am ok with `--exclude-file` parameters, will add :+1:

TheFRedFox: The last style was also meant as addition to the file one. Then you don't have the need to always provide an extra file if you want to exclude something. Might be useful in general, not specifically for now already.

seqizz:
There was a logical weirdness also, which led script to print nothing if only an excluded cron found. Should be better now, please re-check when you have time.

src/check_cron_runtime.py
```diff
@@ -124,7 +141,7 @@ def main(verbose=False):
                     cron_pid_more_than1.append(pid)
                     warn_flag += 1
             elif int(cron_child_time) < 3600:
-                ok_flag = 1
+                continue
```
TheFRedFox, line 144: Btw, not part of PR and just saying, there is no check for `== 3600`.

TheFRedFox:
> There was a logical weirdness also, which led script to print nothing if only an excluded cron found. Should be better now, please re-check when you have time.

This should have been the case even before this PR.

src/check_cron_runtime.py
seqizz: Aaaah I'll change > to >=

PR #201, opened 2020-05-27
check_multiple: open subprocess in text mode

elmatze79:


src/check_multiple.py
```diff
@@ -36,7 +36,7 @@
 subs = {}
 
 for iterate_param in args.iterate_over:
-    subs[iterate_param] = subprocess.Popen(args.origin_command + ' ' + iterate_param , stdout=subprocess.PIPE, shell=True)
+    subs[iterate_param] = subprocess.Popen(args.origin_command + ' ' + iterate_param , stdout=subprocess.PIPE, shell=True, universal_newlines=True)
```
TuxPowered42, line 39: Why is this needed? And it would be nice if we could get rid of `shell=True` if possible.

kofrezo: From the documenation:

> If encoding or errors are specified, or text is true, file objects for stdin, stdout and stderr are opened in text mode using the specified encoding and errors or the io.TextIOWrapper default. The universal_newlines argument is equivalent to text and is provided for backwards compatibility. By default, file objects are opened in binary mode.

If we don't specify this we sometimes get bytes and sometimes strings. This will give us always strings and requires no further encoding and decoding in the Python code.

PR #205, opened 2020-07-01
Make multiple checks Python3 compatible

lamaral:


src/check_haproxy.py
```diff
@@ -26,53 +26,69 @@
...
+        f = haproxy_socket.makefile()
+        lbpools = {}
+        lbstatuses = {}
+        status = ''
+
+        for line in f.readlines():
+            lbstatus = {}
+            if line.startswith('#') or line == "\n":
+                continue
+            split = line.split(',')
```
Woellchen, line 61: is there a `decode()` call needed? not sure

lamaral: It's running properly without the `decode()`, no errors given. 

I double-checked and this codeline is being executed, so I think it's not needed. 

I also tried a `print(type(line))` on line 58 and it's giving me str. I couldn't find official docs on the return type of readlines() to back it up, tho.

PR #208, opened 2020-08-14
Add check for removed file handles

seqizz:
Not sure if that does make sense, but this would have rescued few minutes of confusion from time to time.

### DIFF
--- a/src/check_deleted_filesize.py
+++ b/src/check_deleted_filesize.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - Deleted files size check
+
+Copyright (c) 2020 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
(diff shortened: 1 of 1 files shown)


src/check_deleted_filesize.py
```diff
@@ -0,0 +1,121 @@
...
+from shutil import disk_usage
+
+
+def walklevel(path, depth=1):
+    """It works just like os.walk, but you can pass it a level parameter
+       that indicates how deep the recursion will go.
+
+       If depth is 1, the current directory is listed.
+       If depth is 0, nothing is returned.
+       If depth is -1 (or less than 0), the full depth is walked.
```
TheFRedFox, line 36: What about if `None` is passed?

src/check_deleted_filesize.py
```diff
@@ -0,0 +1,121 @@
...
+        return
+
+    # path.count(os.path.sep) is safe because on UNIX
+    # "/" is never allowed in the name of a file or directory
+    base_depth = path.rstrip(os.path.sep).count(os.path.sep)
+    for root, dirs, files in os.walk(path):
+        yield root, dirs[:], files
+        cur_depth = root.count(os.path.sep)
+        if base_depth + depth <= cur_depth:
+            del dirs[:]
```
TheFRedFox, line 56: I am not sure how this is actually working. To me it still looks like iterating through the whole `os.walk`?

src/check_deleted_filesize.py
```diff
@@ -0,0 +1,121 @@
...
+
+    total_size = 0
+
+    for (dirpath, _, filenames) in walklevel('/proc', depth=2):
+        if dirpath.endswith('fd'):
+            for link in filenames:
+                real_file = os.path.realpath('{}/{}'.format(dirpath, link))
+                full_link = '{}/{}'.format(dirpath, link)
+                if 'deleted' in real_file:
+                    total_size += Path(full_link).stat().st_size
```
TheFRedFox, line 73:
```suggestion
        if not dirpath.endswith('fd'):
            continue
            
        for link in filenames:
            real_file = os.path.realpath('{}/{}'.format(dirpath, link))
            if 'deleted' not in real_file:
                continue
             
            full_link = '{}/{}'.format(dirpath, link)
            total_size += Path(full_link).stat().st_size
```

src/check_deleted_filesize.py
```diff
@@ -0,0 +1,121 @@
...
+def get_total_deleted_size():
+    """Walks into /proc/*/fd and sums filesizes of deleted objects
+
+    :return: byte count, float
+    """
+
+    total_size = 0
+
+    for (dirpath, _, filenames) in walklevel('/proc', depth=2):
+        if dirpath.endswith('fd'):
```
TheFRedFox, line 68: If you are interested in all `*fd` folders, maybe there is also a possibility with `glob`?

PR #211, opened 2020-09-23
Remove warnings from redis_cluster_status check

meikomeik:
Improve code quality of check_redis_cluster.py


src/check_redis_cluster.py
```diff
@@ -32,24 +32,26 @@
...
     """Main entrypoint for script"""
 
     args = get_parser().parse_args()
 
-    master_addr, master_state, cluster_state_master, failed_master = _get_cluster_status(
-        args.master_port, args.password)
-    slave_addr, slave_state, cluster_state_slave, failed_slave = _get_cluster_status(
-        args.slave_port, args.password)
+    master_addr, master_state, cluster_state_master, failed_master = \
+        _get_cluster_status(args.master_port, args.password)
```
TheFRedFox, line 42: `\` is not a preferred way in PEP8 resp. braces is the preferred way for multilining. I am not completely sure if the below is the correct syntax, but try it with that:
```suggestion
    master_addr, master_state, cluster_state_master, failed_master = (
        _get_cluster_status(args.master_port, args.password)
    )
```

src/check_redis_cluster.py
```diff
@@ -32,24 +32,26 @@
...
+    slave_addr, slave_state, cluster_state_slave, failed_slave = \
```
TheFRedFox, line 43: Same here.

src/check_redis_cluster.py
```diff
@@ -32,24 +32,26 @@
...
     if master_state != 'unknown' and slave_state != 'unknown':
+        failed_hosts = failed_master + failed_slave
         if cluster_state_master != 'ok' and cluster_state_slave != 'ok':
             print('CRITICAL - cluster is broken')
             code = 2
-        elif len(failed_master + failed_slave) > 0:
+        elif failed_hosts:
             print('WARNING - cluster status is degraded')
-            for host in failed_master + failed_slave:
+            code = 1
```
TheFRedFox, line 52: Good one.

src/check_redis_cluster.py
```diff
@@ -114,7 +116,8 @@ def _get_cluster_status(port, password):
             'redis-cli -p {0} -a {1} cluster nodes'.format(
                 port, password), shell=True).decode().split()
         state_index = [i for i, s in enumerate(role) if 'myself' in s][0]
-        failed_hosts = [role[i-1] for i, s in enumerate(role) if 'fail' in s and str(port) in role[i-1]]
+        failed_hosts = [role[i-1] for i, s in enumerate(role)
+                        if 'fail' in s and str(port) in role[i-1]]
```
TheFRedFox, line 120: `str(port)` could be moved outside, so it won't happen for every iteration.

src/check_redis_cluster.py
```diff
@@ -109,14 +111,33 @@ def _get_cluster_status(port, password):
...
+    The string ",fail" is optional. It will only exists on failed hosts
+    """
     try:
         role = subprocess.check_output(
             'redis-cli -p {0} -a {1} cluster nodes'.format(
                 port, password), shell=True).decode().split()
+        # Find keyword myself in the output
         state_index = [i for i, s in enumerate(role) if 'myself' in s][0]
-        failed_hosts = [role[i-1] for i, s in enumerate(role) if 'fail' in s and str(port) in role[i-1]]
+        # Remove "myself," from the string. It will be eigther master or slave
```
TheFRedFox, line 133:
```suggestion
        # Remove "myself," from the string. It will be either master or slave
```

src/check_redis_cluster.py
meikomeik: I agree. We only need the port as string anyway. Changed it on arg parser to string

src/check_redis_cluster.py
meikomeik: replaced as well

src/check_redis_cluster.py
meikomeik: Thank you. Fixed

TheFRedFox (approved):

PR #212, opened 2020-09-27
Allow check_bird_protocols to work with short or long date format

lamaral:
The current version of the check only works with uptimes in ISO short format. 

I have an use case where I actually needed the ISO long format for the protocol uptimes and there the check broke, as we had an extra field. 

I have tested these changes with the ISO long format. Testing with ISO short is still pending.

### DIFF
--- a/src/check_bird_protocols.py
+++ b/src/check_bird_protocols.py
@@ -193,7 +193,7 @@ def parse_birdc_output(output):
     lines = lines[2:]
 
     tokenized_lines = [
-        [field.strip() for field in line.split(None, 5)]
+        [field.strip() for field in line.split(None, 6)]
         for line in lines
     ]
     protocols_dicts = [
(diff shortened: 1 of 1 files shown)


src/check_bird_protocols.py
```diff
@@ -204,18 +204,26 @@ def parse_birdc_output(output):
...
+    structure = {
         'name': line[0],
         'type': line[1],
         'table': line[2],
         'state': line[3],
-        'since': line[4],
-        'info': line[5],
     }
 
+    if ':' in line[5]:
```
TheFRedFox, line 218: I would recommend to add a short comment about what you are solving here. Like what does the line look like so it is understandable why this is needed.

lamaral: I pushed a commit with an explanatory comment.

TheFRedFox: 👍

TheFRedFox (approved):

PR #214, opened 2020-10-19
Fix the wrong level for non-active service units

Felixoid:
It looks like during the latest refactoring the logic of crit/warn was messed up a little bit. Here is an example of running the master code against the branch one

```
$ /usr/bin/python3 /usr/lib/nagios/igmonplugins/check_systemd_units.py -u 'nginx.service'
OK
$ /usr/bin/python3 check_systemd_units.py -u 'nginx.service'
OK
$ sudo systemctl stop nginx
$ /usr/bin/python3 /usr/lib/nagios/igmonplugins/check_systemd_units.py -u 'nginx.service'
WARNING: the service is inactive: nginx.service
$ /usr/bin/python3 /usr/lib/nagios/igmonplugins/check_systemd_units.py -a -i proc-sys-fs-binfmt_misc.automount
OK
$ /usr/bin/python3 check_systemd_units.py -u 'nginx.service'
CRITICAL: the service is inactive: nginx.service
$ /usr/bin/python3 check_systemd_units.py -a
WARNING: the service is inactive: systemd-fsckd.service
```

### DIFF
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -200,11 +200,12 @@ class SystemdUnit:
         else:
             if self.unit_properties.ActiveState != 'active':
                 return (
-                    self._warn_level, 'the service is inactive'
+                    self._crit_level, 'the service is inactive'
                 )
             if self.unit_properties.SubState == 'exited':
+                # Non one-shot services should not exit
                 return (
-                    self._warn_level, 'the service is exited'
+                    self._crit_level, 'the service is exited'
                 )
 
         return (Codes.OK, '')
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -211,7 +211,7 @@ class SystemdUnit:
                     self._crit_level, 'the service is inactive'
                 )
             if self.unit_properties.SubState == 'exited':
-                # Non one-shot services should not exit
+                # Non oneshot services should not exit
                 return (
                     self._crit_level, 'the service is exited'
                 )
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -161,6 +161,13 @@ class SystemdUnit:
             # services silently without raising a warning which requires no
             # manual action.
             return (Codes.OK, '')
+        if (
+            hasattr(self.unit_properties, 'ConditionResult') and
+            not self.unit_properties.ConditionResult
+        ):
+            # systemd on Debian Buster contains a lot of services in inactive
+            # state by "Condition*" parameters, it's fine to ignore them
+            return (Codes.OK, '')
 
         # Most probably, oneshot is related to some timer
         if self.type_properties.Type == 'oneshot':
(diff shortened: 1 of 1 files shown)


kofrezo (approved):
I can't judge if it makes sense to return "OK" for units that are inactive be evaluating the condition parameters but code wise it looks good to me.

Felixoid:
I could clarify Contitions. A good example of the subj is `systemd-timesyncd.service`.

So, it's a normal service, but contains additional configuration:

```
> cat /lib/systemd/system/systemd-timesyncd.service.d/disable-with-time-daemon.conf
[Unit]
# don't run timesyncd if we have another NTP daemon installed
ConditionFileIsExecutable=!/usr/sbin/ntpd
ConditionFileIsExecutable=!/usr/sbin/openntpd
ConditionFileIsExecutable=!/usr/sbin/chronyd
ConditionFileIsExecutable=!/usr/sbin/VBoxService
```

So, it runs unless any of the other time syncing daemons is installed. So, `ConditionResult` tells exactly about `if the service should run`.

Felixoid:
Will merge and build this tomorrow. The internal adjustments will be right in place then.

src/check_systemd_units.py
```diff
@@ -200,11 +208,12 @@ def _check_service(self, timer=False):
         else:
             if self.unit_properties.ActiveState != 'active':
                 return (
-                    self._warn_level, 'the service is inactive'
+                    self._crit_level, 'the service is inactive'
                 )
             if self.unit_properties.SubState == 'exited':
+                # Non one-shot services should not exit
```
TheFRedFox, line 214: Above it is called `oneshot` in one word. Let's stick to one version.
```suggestion
                # Non oneshot services should not exit
```

src/check_systemd_units.py
```diff
@@ -481,4 +490,7 @@ def gen_output(results):
 
 
 if __name__ == '__main__':
-    main()
+    with ExitStack() as stack:
```
TheFRedFox, line 493: Not gonna reject on that, but let's discuss it maybe internally: Is it ok for this to be outside of main or should that be part of the main function, too?

TheFRedFox (approved):
Just one minor comment and one internal discussion. Approved anyway.

src/check_systemd_units.py
Felixoid: tl;dr: TBH, I don't know for 100%. But the current state makes more sense to me.

In my imagination, if somebody executes `main()`, he should take care of the Manager by himself, and it isn't our busyness to close it.

Maybe it makes sense to clarify what problem do I solve here. When the code is executed with `python -m pdb src/check_systemd_units.py *args **kwargs`, it's impossible to restart it because the manager is already subscribed. I could imagine the case when the module is imported in an ipython for debugging, and there I wouldn't expect if it will stop working after executing `main()`

TheFRedFox: Maybe then it would be interesting to pass an instance (or the like) to the `main` function actually (if that is possible) to tell the main function which to use. Then it could be reused everywhere.
But again, nothing for the PR right now, rather something to discuss for the generality.

PR #215, opened 2020-10-22
Use fnmatch to check unit against glob

Felixoid:
Python contains a standard function to check if a file matches glob, so let's use it.

### DIFF
--- a/src/check_systemd_units.py
+++ b/src/check_systemd_units.py
@@ -35,6 +35,7 @@ import logging
 from argparse import ArgumentParser
 from contextlib import ExitStack
 from datetime import datetime
+from fnmatch import fnmatch
 from sys import exit
 from time import time
 
@@ -104,10 +105,8 @@ class SystemdUnit:
 
     def match(self, pattern):
         name = str(self)
-        if pattern.endswith('@*') and '@' in name:
-            return pattern[:-len('@*')] == name.split('@', 1)[0]
         logger.debug('Pattern is {}, name is {}'.format(pattern, name))
-        return pattern == name
+        return fnmatch(name, pattern)
 
     def check(self, timer_warn, timer_crit, critical=True, timer=False):
         """


src/check_systemd_units.py
```diff
@@ -104,10 +105,8 @@ def type_properties(self):
 
     def match(self, pattern):
         name = str(self)
-        if pattern.endswith('@*') and '@' in name:
-            return pattern[:-len('@*')] == name.split('@', 1)[0]
```
kofrezo, line 108: Removing this line and switching to fnmatch module will possibly require us to adjust usage of the script no ?

Felixoid: The fnmatch extends the current behavior. Let me show some examples:

```python
In [1]: from fnmatch import fnmatch

In [2]: def old(name, pattern):
   ...:     if pattern.endswith('@*') and '@' in name:
   ...:         return pattern[:-len('@*')] == name.split('@', 1)[0]
   ...:     return name == pattern
   ...:

In [3]: def new(name, pattern):
   ...:     return fnmatch(name, pattern)
   ...:

In [4]: cases = (('name@1.service', 'name@*.service'),
   ...: ('name@1.service', 'name@*'),
   ...: ('name@1.timer', 'name@*.timer'),
   ...: ('name@1.timer', 'name@*')
   ...: )

In [5]: for c in cases:
   ...:     print(old(c[0], c[1]), new(c[0], c[1]))
   ...:
False True
True True
False True
True True
```

I'd say, it implements more expected behavior.

kofrezo: Your example clearly shows it does not only extend the behaviour it changes the current behaviour (Case 1 and 3) therefore my question in the first place.

I am fine with changing the behaviour if it provides as more possibilities or "more intuitive" usage but please make sure current usages are adjusted if necessary and that it is clear for externals using the script what has changed.

kofrezo (approved):

src/check_systemd_units.py
TheFRedFox: I agree here. I like the use of `glob` style matching more and it is especially also more widely used and known. However, we need to bring that to the users that for some occasion the behavior is changing. It looks like it should not be a problem except someone was matching for `*` or other glob styled special characters in the old one, still we should first make people aware of the change.

Felixoid: I've checked our puppet modules, the only place where it's used is the multiserver, and it matches the `name@*`, but should match `name@*.service`.

This is a patch, that I'll send next week.

Thank you fir review!

PR #216, opened 2020-10-23
Add check for ZFS pool state

TuxPowered42:


src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+
+from subprocess import check_output, STDOUT, CalledProcessError
+from argparse import ArgumentParser
+from sys import exit
+import re
+import platform
+
+
+# Nagios plugin exit codes
+class ExitCodes:
```
lamaral, line 32: Per our Python script coding style, helper classes should be moved to the end, before the `if __name__ == '__main__':` block.

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+    tank_130T_75.9T_54.1T_-_22%_58%_1.00x_ONLINE_-
+    ...
+    """
+
+    line_split = line.split('\t')
+
+    if line_split[8] == 'ONLINE':
+        ret_code = ExitCodes.ok
+    else:
+        ret_code = ExitCodes.critical
```
lamaral, line 93: Do we really want to go straight critical on all other cases?

Our raid_state check goes to Warning in some cases of degradation.

lamaral (changes requested):

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+def main():
+    parser = ArgumentParser(
+        description=('Check if all ZFS pools are online')
+    )
+    
+    args = parser.parse_args()
+
+    try:
+        proc = check_output(
+            ['/sbin/zpool', 'list', '-H'],stderr=STDOUT,
```
TheFRedFox, line 48:
```suggestion
            ['/sbin/zpool', 'list', '-H'], stderr=STDOUT
```

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+        exit_header = 'All zpools are fine'
+        for line in lines:
+            zpool_code, zpool_name, zpool_state = parse_line(line)
+            exit_multiline.append('{}: {}'.format(zpool_name, zpool_state))
+            if zpool_code > exit_code:
+                exit_code = zpool_code
+                exit_header = 'Some zpools are not online!'
+    else:
+        exit_code = ExitCodes.unknown
+        exit_header = 'No zpools found!'
```
TheFRedFox, line 71: Early return can be used here:

```suggestion
    if not lines:
        print('No zpools found!')
        return ExitCodes.unknown
    
    exit_code = ExitCodes.ok
    exit_header = 'All zpools are fine'
    for line in lines:
        zpool_code, zpool_name, zpool_state = parse_line(line)
        exit_multiline.append('{}: {}'.format(zpool_name, zpool_state))
        if zpool_code > exit_code:
            exit_code = zpool_code
            exit_header = 'Some zpools are not online!'
```

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+
```
TheFRedFox, line 72: Seems to be an unneeded extra line
```suggestion
```

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from subprocess import check_output, STDOUT, CalledProcessError
+from argparse import ArgumentParser
+from sys import exit
```
TheFRedFox, line 26: Like that the builtin `exit` function will be shadowed. For that we might rather just want to import `sys` and use `sys.exit`.

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if line_split[8] == 'ONLINE':
+        ret_code = ExitCodes.ok
+    else:
+        ret_code = ExitCodes.critical
+
+    return (ret_code, line_split[0], line_split[8])
+
+if __name__ == '__main__':
+    exit(main())
```
TheFRedFox, line 98: It's actually not dumb to do it like that. ^^

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+    if exit_multiline:
+        print('\n'.join(exit_multiline))
+    return exit_code
+
+def parse_line(line):
+    """
+    The output looks like this, with -H there are real tabs separating fields:
+
+    # sudo zpool list -H | tr '\t' _
+    tank_130T_75.9T_54.1T_-_22%_58%_1.00x_ONLINE_-
```
TheFRedFox, line 84: :+1: for the tabs you could also just write it `tank<\tr>130T` so it should be self explaining. But nothing to do, it's good like that already. ;)

src/check_zfs.py
```diff
@@ -0,0 +1,98 @@
...
+    ...
+    """
+
+    line_split = line.split('\t')
```
TheFRedFox, line 88: It could also possible like that (if it fits with the line limit):
```suggestion
    name, _, _, _, _, _, _, _, code = line.split('\t')
```
`name` would hold `line_split[0]` then and `code` `line_split[8]`. `_` would hold the last occurrence, so `line_split[7]`

Both is fine for me here, you can choose.

src/check_zfs.py
TheFRedFox: 👍

TheFRedFox (changes requested):

src/check_zfs.py
TuxPowered42: Fixed.

src/check_zfs.py
TuxPowered42: Fixed.

lamaral (approved):

src/check_zfs.py
```diff
@@ -0,0 +1,126 @@
...
+    except CalledProcessError as e:
+        print('UNKNOWN: can\'t read zpool status: {}'.format(e.output))
+        return ExitCodes.unknown
+    
+    exit_multiline = []
+    lines = proc.splitlines()
+
+    if not lines:
+        print('No zpools found!')
+        return(ExitCodes.unknown)
```
TheFRedFox, line 69:
```suggestion
        return ExitCodes.unknown
```

src/check_zfs.py
```diff
@@ -0,0 +1,126 @@
...
+    ):
+        ret_code = ExitCodes.warning
+
+    if (
+        frag > args.fragmentation[1] or
+        cap > args.capacity[1]
+    ):
+        ret_code = ExitCodes.critical
+
+    return (ret_code, name, "health: {}, fragmentation: {}%, capacity: {}%".format(health, frag, cap))
```
TheFRedFox, line 114: By separating this it's easier to read and understand for me.
```suggestion

    message = 'health: {}, fragmentation: {}%, capacity: {}%'
    return (ret_code, name, message.format(health, frag, cap))
```

TheFRedFox, line 114: Looks like single quotes should rather be used.
```suggestion

    return (ret_code, name, 'health: {}, fragmentation: {}%, capacity: {}%'.format(health, frag, cap))
```

TheFRedFox (approved):
Just 1 `return(..)` with braces and 1 thing which I would change. Will give it an early approve already though.

src/check_zfs.py
```diff
@@ -0,0 +1,126 @@
...
+def parse_zpool(line, args):
+    """
+    The output looks like this, with -H there are real tabs separating fields:
+
+    # sudo zpool list -H
+    tank\t130T\t75.9T\t54.1T\t-\t22%\t58%\t1.00x\tONLINE\t-
+    ...
+    """
+
+    name, _, _, _, _, frag, cap, _, health, _ = line.split('\t')
```
kofrezo, line 95:
```suggestion
    status = line.split('\t')
    name, frag, cap, health = status[0], status[5], status[6], status[8]
```

kofrezo (approved):

src/check_zfs.py
TuxPowered42: I've done it the way @TheFRedFox suggested. Which one is more pythonic?

src/check_zfs.py
TuxPowered42: Fixed.

TuxPowered42: Fixed.

src/check_zfs.py
TuxPowered42: Fixed.

src/check_zfs.py
TuxPowered42: I've used a table before and @TheFRedFox asked me to change it to this format with `_`. Which is more pythonic?

TheFRedFox: I would say mine and @kofrezo's is more pythonic than your original one. Now, I guess, it is more about readability, for which I find @kofrezo's actually even more readable than mine.

TuxPowered42: I see one advantage of the current code: if the number of columns changes, the code will crash, no matter if we're using those columns or not. Which is good, becasue that allows us to easily spot modified output of zfs command.

kofrezo: As @TheFRedFox pointed out my point was to increase readability. 

Yes, your way would cover if the amount of columns changes but also not if the ordering or format changes so if you really want validation I think this should be a dedicated step. 

If you want to cover the length you could additionally add a assert statement. This would not break with a cryptic error message such as "Too many values to unpack" or "Too few values to unpack" but rather make clear what you expect.

```
assert len(data) == 9, "Expected 9 columns from zfs pool list command"
```

TuxPowered42: How abouta now?

kofrezo: I think it is good this way :+1:

PR #218, opened 2020-12-09
Add timeout to api_reachability

seqizz:


src/check_api_reachability.py
```diff
@@ -74,7 +75,7 @@ def check_urls(url_list, codes_list):
     unreachable_urls = []
     for url in url_list:
         try:
-            request = urlopen(url)
+            request = urlopen(url, timeout=10)
```
kofrezo, line 78: Keep in mind that this is the timeout for everything (establishing the connection, processing the request etc.) not only the initial establishment.

kofrezo (approved):

src/check_api_reachability.py
seqizz: Since we're putting the reason into the exception, should be ok. E.g.:
```
CRITICAL - 1 URLs with problem. https://portal-api.innogames.de/status.php - <urlopen error timed out>
```

kofrezo: Sounds sane.  I was just worried that we have connections that may take a little longer than 10 seconds overall.

TheFRedFox (approved):

PR #219, opened 2021-01-06
libvirt_hosts: Use also Serveradmin data

lamaral:


src/check_libvirt_hosts.py
```diff
@@ -100,58 +102,133 @@
...
+def query_serveradmin_vms():
+    hostname = socket.gethostname()
+
+    vms = Query(
+        {'servertype': 'vm', 'hypervisor': hostname},
+        ['hostname', 'state']
+    )
+    # This is a trick to get pure dicts out of the serveradmin data, since
+    # we don't need to edit the queried objects.
+    vms = [vm.copy() for vm in vms]
```
TuxPowered42, line 164: If you feel like materializing the query, you can do `vms = list(vms)`

src/check_libvirt_hosts.py
```diff
@@ -100,58 +102,133 @@
...
-              .get(reason, "unknown")))
+    parsed_domains = parse_libvirt_domains(domains)
+
+    vm_object_ids = set([vm['object_id'] for vm in vms])
+    domain_object_ids = set([domain['object_id'] for domain in parsed_domains])
 
+    #  - Check if same object ID is defined twice in libvirt
+    if len(domain_object_ids) != len(parsed_domains):
+        # TODO: Implement logic properly
+        print('Duplicated domain defined')
```
TuxPowered42, line 132: It does not say which domain is duplicated.

src/check_libvirt_hosts.py
```diff
@@ -100,58 +102,133 @@
...
 
-def check(domains):
+    #  - Check if there are VMs defined in libvirt that are not associated to
+    #  this hypervisor in Serveradmin
+    if len(domain_object_ids - vm_object_ids) > 0:
+        # TODO: Implement logic properly
+        print('Running VM not associated to the hypervisor in serveradmin')
```
TuxPowered42, line 138: It does not say which domain is not associated in Serveradmin. Also any multiple-running VM could be easily found by just removing every found VM from the list of VMs from Serveradmin.

src/check_libvirt_hosts.py
```diff
@@ -100,58 +102,133 @@
...
+
+    #  - Check if there are VMs associated to this HV in Serveradmin but not
+    #  defined in libvirt
+    if len(vm_object_ids - domain_object_ids) > 0:
+        # TODO: implement logic properly
+        print('VM associated in serveradmin not running on this HV')
```
TuxPowered42, line 144: It does not say which VM.

src/check_libvirt_hosts.py
```diff
@@ -100,58 +102,133 @@
...
     return ExitCodes.ok, 'All defined domains are running'
 
 
-def main():
+def query_serveradmin_vms():
+    hostname = socket.gethostname()
+
+    vms = Query(
+        {'servertype': 'vm', 'hypervisor': hostname},
+        ['hostname', 'state']
```
TuxPowered42, line 160: Is object_id always given to us?

src/check_libvirt_hosts.py
```diff
@@ -100,58 +102,133 @@
...
+    return parsed_domains
+
+
+def print_domains(domains):
+    for d in domains:
+        state, reason = d.state()
+        m = re.match(r'(?P<object_id>\d+_)?(?P<hostname>[\w\.\d-]+)', d.name())
+        name = m.group('hostname')
+        print("{0} - {1}".format(
+            name, reason_names.get(state, "unknown").get(reason, "unknown")
```
TuxPowered42, line 211: The reason/state of VMs is not really human-readible. It could be better described as "running as defined in Serveradmin", "not running as defined in Serveradmin because of retired state", "running even though not defined in Serveradmin" and so on, maybe with some shorter forms.

TuxPowered42 (changes requested):

src/check_libvirt_hosts.py
lamaral: Keep in mind it's heavy WIP. There is a TODO to make it proper :)

src/check_libvirt_hosts.py
lamaral: Keep in mind it's heavy WIP. There is a TODO to make it proper :)

src/check_libvirt_hosts.py
lamaral: Yes, it's always given, but I'll make it explicit to be on the safe side.

src/check_libvirt_hosts.py
lamaral: This part is leftover. It will get redone.

src/check_libvirt_hosts.py
lamaral: Doing like that, would result in a list of DatasetObjects. 
This way, I get a list of dictionaries, containing the DatasetObject data.

src/check_libvirt_hosts.py
lamaral: This is fixed in the newer version.

lamaral:
I'm closing this PR as this functionality will be covered by  the new IGVM consistency check.

lamaral:
It's reopened as we preferred to go with the simpler solution implemented here.

PR #220, opened 2021-02-10
Add check_barman

Phantuman:
At the moment we are checking every host twice,
if there is an error or the host ist a standby node.
This check will prevent this.

### DIFF
--- a/src/check_barman.py
+++ b/src/check_barman.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - Barman Check
+
+This is a Nagios script which checks, if the given host is currently backing
+up currently on barman.
+
+The script will exit with:
+ - 0 (OK) All checks are ok, or this is a secondary node which can not be
+    backed up
+ - 2 (CRITICAL) Any error was detected during the backup check
+
+Copyright (c) 2021 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import sys
+from subprocess import Popen, PIPE, STDOUT
+from argparse import ArgumentParser
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--host',
+                        help='The hostname for which to check the barman status'
+                             '(e.g. "am0db1.moc.ig.local")')
+    return parser.parse_args()
+
+
+def main(args):
+    with Popen(
+        '/usr/bin/sudo -u barman '
+        '/usr/bin/barman check --nagios {}'.format(args.host),
+        shell=True,
+        stderr=STDOUT,
+        stdout=PIPE,
+        encoding='UTF-8'
+    ) as result:
+        # NRPE has a 30 seconds timeout. so there is no need to wait
+        # longer here
+        output = result.communicate(timeout=30)[0].strip()
+        return_code = result.returncode
+
+    # Standby nodes will always throw an error. We will ignore them
+    if 'cannot perform exclusive backup on a standby' in output:
+        print('BARMAN OK - {} is a standby'.format(args.host))
+        sys.exit(0)
+
+    print(output)
+    sys.exit(return_code)
+
+
+if __name__ == '__main__':
+    main(parse_args())


src/check_barman.py
```diff
@@ -0,0 +1,71 @@
...
+
+def main(args):
+    result = Popen(
+        '/usr/bin/sudo -u barman '
+        '/usr/bin/barman check --nagios {}'.format(args.host),
+        shell=True,
+        stderr=STDOUT,
+        stdout=PIPE
+    )
+    output = result.communicate()[0].decode('UTF-8').strip()
```
TheFRedFox, line 53: Btw, you can pass in an encoding to `Popen` with the parameter `encoding`, then the outputs of the result is already decoded.

TheFRedFox, line 53: Another btw: `utf-8` is default for `decode`. Can still be put in as explicit is better than implicit.

TheFRedFox, line 53: How much do we need to care for a timeout in this case as stated here: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

src/check_barman.py
```diff
@@ -0,0 +1,71 @@
...
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--host',
+                        help='The hostname for which to check the barman status'
+                             '(e.g. "am0db1.moc.ig.local"')
+    return parser.parse_args()
+
+
+def main(args):
+    result = Popen(
```
TheFRedFox, line 46: `Popen` also has a context manager since 3.2, example: https://stackoverflow.com/a/53498797/2625488

TheFRedFox (commented):
Just some notes actually so far.

src/check_barman.py
Phantuman: moved the encoding to the Popen command

src/check_barman.py
```diff
@@ -0,0 +1,71 @@
...
+    # Standby nodes will always through an error. We will ignore them
+    if 'cannot perform exclusive backup on a standby' in output:
+        print('BARMAN OK - {} is a standby'.format(args.host))
+        sys.exit(0)
+
+    print(output)
+
+    # There was an error with the check of the given host
+    if return_code != 0:
+        sys.exit(2)
```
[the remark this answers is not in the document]
Phantuman: The output is done once above this line.

src/check_barman.py
```diff
@@ -0,0 +1,71 @@
...
+        '/usr/bin/sudo -u barman '
+        '/usr/bin/barman check --nagios {}'.format(args.host),
+        shell=True,
+        stderr=STDOUT,
+        stdout=PIPE
+    )
+    output = result.communicate()[0].decode('UTF-8').strip()
+    return_code = result.returncode
+
+    # Standby nodes will always through an error. We will ignore them
```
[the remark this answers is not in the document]
Phantuman: fixed

src/check_barman.py
```diff
@@ -0,0 +1,71 @@
...
+import sys
+from subprocess import Popen, PIPE, STDOUT
+from argparse import ArgumentParser
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--host',
+                        help='The hostname for which to check the barman status'
+                             '(e.g. "am0db1.moc.ig.local"')
```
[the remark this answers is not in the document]
Phantuman: fixed

src/check_barman.py
Phantuman: switched to context maneger

src/check_barman.py
Phantuman: added timeout and comment

src/check_barman.py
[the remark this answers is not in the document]
TheFRedFox: Right, ignore that. ^^

src/check_barman.py
```diff
@@ -0,0 +1,74 @@
...
+        print('BARMAN OK - {} is a standby'.format(args.host))
+        sys.exit(0)
+
+    print(output)
+
+    # There was an error with the check of the given host
+    if return_code != 0:
+        sys.exit(2)
+
+    sys.exit(0)
```
TheFRedFox, line 70:
```suggestion
    sys.exit(0 if return_code == 0 else 2)
```
Was just checking if that might be nicer to read. Not sure though. ^^

src/check_barman.py
```diff
@@ -0,0 +1,74 @@
...
+        '/usr/bin/sudo -u barman '
+        '/usr/bin/barman check --nagios {}'.format(args.host),
+        shell=True,
+        stderr=STDOUT,
+        stdout=PIPE,
+        encoding='UTF-8'
+    ) as result:
+        # NRPE has a 30 seconds timeout. so there is no need to wait
+        # longer here
+        output = result.communicate(timeout=30)[0].strip()
```
TheFRedFox, line 56: Not 100% sure right now, but the documentation also says something about catch the Timeout error to clean up the process? Is that something we need to do here, too?

src/check_barman.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - Barman Check
+
+This is a Nagios script which checks, if the given host is currently backing
+up currently on barman.
+
+The script will exit with:
+ - 0 (OK) All checks are ok, or this is a secondary node which can not be
+    backed up
```
TheFRedFox, line 9: Isn't there the one for standby missing?

TheFRedFox (commented):
One more thing about the timeout.

src/check_barman.py
```diff
@@ -0,0 +1,74 @@
...
+ - 2 (CRITICAL) Any error was detected during the backup check
+
+Copyright (c) 2020 InnoGames GmbH
```
lamaral, line 12: 2021 please

src/check_barman.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Standby nodes will always throw an error. We will ignore them
+    if 'cannot perform exclusive backup on a standby' in output:
+        print('BARMAN OK - {} is a standby'.format(args.host))
+        sys.exit(0)
+
+    print(output)
+
+    # There was an error with the check of the given host
+    if return_code != 0:
```
lamaral, line 67: If the intended behavior is that any error of the barman check nomatter what results in a critical, then I'm fine with this. 
On the other hand, we could probably just honor the barman check return code. If it was made for Nagios, it would return proper codes already.

lamaral (approved):

src/check_barman.py
TheFRedFox: Ok, I understand now, that these are explanations and not examples. Then fine.

src/check_barman.py
TheFRedFox: After researching both type of context manager again, it came clear that the context manager build as class will always execute `__exit__` method. So, this should be safe and correct like that.

TheFRedFox (approved):

PR #222, opened 2021-04-20
Fix nagios warnings for new replication

elmatze79:


axxelG:
I don't think we should ignore the read only setting everywhere. We should have this configurable.

elmatze79:
needs some rework to make the setting configureable

PR #224, opened 2021-04-27
Split LBPool check logic from LBPool data export

lamaral:


src/lbpool_export.py
```diff
@@ -0,0 +1,148 @@
...
+        help='Password for Redis server authentication'
+    )
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+
+    with \
```
TuxPowered42, line 54: Does `with` work with `(…)`? That would look better IMO.

src/lbpool_export.py
```diff
@@ -0,0 +1,148 @@
...
+        results = compare_pools(
+            lbpools, carps_states, lbpools_states, args.hwlb_group
+        )
+        if args.redis_servers:
+            try:
+                send_to_redis(
+                    args.redis_servers, args.redis_password,
+                    results, args.hwlb_group
+                )
+            except redis.exceptions.AuthenticationError:
```
TuxPowered42, line 71: There's no reason to catch it, if sending to Redis is impossible, there is no way to recover. Maybe unless the error message is totally unreadible to user.

src/lbpool_export.py
```diff
@@ -0,0 +1,148 @@
...
+            open('/var/run/iglb/carp_state.json') as carps_states_f, \
+            open('/var/run/iglb/lbpools_state.json') as lbpools_states_f \
+            :
+        lbpools = json.load(lbpools_f)
+        carps_states = json.load(carps_states_f)
+        lbpools_states = json.load(lbpools_states_f)
+        results = compare_pools(
+            lbpools, carps_states, lbpools_states, args.hwlb_group
+        )
+        if args.redis_servers:
```
TuxPowered42, line 65: What if redis servers are not given?

src/lbpool_export.py
```diff
@@ -0,0 +1,148 @@
...
+
+    for pool_k, pool_v in lbpools.items():
+        if not pool_v['nodes']:
+            results.append(get_object_output(pool_k, pool_v, hwlb_group))
+            continue
+
+        # Skip LB Pools for which this HWLB is not master carp
+        first_lbnode = list(pool_v['nodes'].values())[0]
+        if not (
+                first_lbnode['route_network'] in carps_states
```
TuxPowered42, line 89: This is 8 spaces, not 4.

src/lbpool_export.py
```diff
@@ -0,0 +1,148 @@
...
+                and
+                carps_states[first_lbnode['route_network']]['carp_master']
+        ):
+            print(
+                'Skipping LB Pool {} because carp is not master {}'.format(
```
TuxPowered42, line 94: Let's switch to modern Python `f'{var}'` formatting while we are at it.

src/lbpool_export.py
```diff
@@ -0,0 +1,148 @@
...
+        results.append(
+            get_object_output(pool_k, pool_v, hwlb_group, in_testtool)
+        )
+
+    return results
+
+
+def send_to_redis(redis_servers, redis_password, results, hwlb_group):
+    for server in redis_servers:
+        r = redis.Redis(host=server, port=6379, db=0, password=redis_password)
```
TuxPowered42, line 116: What if sending to one of servers fail? An exception thrown here will stop data from being sent to the other one. Is there any timeout for those operations? The cron job is supposed to run every minute, there are multiple servers to send data to, it would be unwise to just hang on one server. Is there a way to send in parallel, using modern Python async io?

src/lbpool_nodes.py
```diff
@@ -0,0 +1,209 @@
...
+        help='Redis servers to report the results to'
+    )
+
+    parser.add_argument(
+        '-p', dest='redis_password',
+        help='Password for Redis server authentication'
+    )
+
+    parser.add_argument(
+        '-n', dest='lbpool_name', required=True,
```
TuxPowered42, line 44: Is this script run on Nagios server? Then the LB Pool checked is the "host" checked by Nagios. So `-H` should be the LB Pool.

src/lbpool_nodes.py
```diff
@@ -0,0 +1,209 @@
...
+    exit_codes = []
+
+    if len(data) < 1:
+        return (
+            [ExitCodes.critical],
+            ['No data could be retrieved for given LBPool']
+        )
+
+    for pool_k, pool_v in data.items():
+        hwlb_group = pool_v.get('hwlb_group', None)
```
TuxPowered42, line 92: Do we expect this key missing in incoming data? I thought the dict is well defined and if yes, then it's safe to crash when keys are missing.

src/lbpool_nodes.py
```diff
@@ -0,0 +1,209 @@
...
+            )
+            exit_codes.append(ExitCodes.critical)
+            continue
+
+        exit_code = ExitCodes.ok
+        optimal_nodes = len(pool_v['nodes'])
+
+        # We want to keep "None" text as it is nicer than 0 for human
+        # perception.
+        min_nodes = pool_v.get('min_nodes', None)
```
TuxPowered42, line 105: Do we expect this key missing in incoming data? I thought the dict is well defined and if yes, then it's safe to crash when keys are missing.

src/lbpool_nodes.py
```diff
@@ -0,0 +1,209 @@
...
+        if min_nodes == 0:
+            min_nodes = None
+        max_nodes = pool_v.get('max_nodes', None)
```
TuxPowered42, line 108: Do we expect this key missing in incoming data? I thought the dict is well defined and if yes, then it's safe to crash when keys are missing.

src/lbpool_nodes.py
```diff
@@ -0,0 +1,209 @@
...
+        if max_nodes == 0:
+            max_nodes = None
+        alive_nodes = pool_v.get('alive_nodes', 0)
```
TuxPowered42, line 111: Do we expect this key missing in incoming data? I thought the dict is well defined and if yes, then it's safe to crash when keys are missing.

TuxPowered42 (changes requested):

src/lbpool_export.py
lamaral: Sadly doesn't, I tried

src/lbpool_export.py
lamaral: Then the script doesn't run. This argument is required, see line 34.

src/lbpool_export.py
lamaral: Argh, forgot to change from the old script. Will adjust.

src/lbpool_nodes.py
lamaral: Yes. 
Actually makes sense, I'll adjust it.

src/lbpool_export.py
lamaral: The message is actually quite clear. I'll remove the try catch.

src/lbpool_export.py
lamaral: It's aligned to the parentheses because of pyCharm.

src/lbpool_nodes.py
lamaral: I have updated it to be `-s` and `-H`

src/lbpool_nodes.py
lamaral: Yes, export has a well defined dict, so we can clean this up.

src/lbpool_nodes.py
lamaral: Yes, export has a well defined dict, so we can clean this up.

src/lbpool_nodes.py
lamaral: Yes, export has a well defined dict, so we can clean this up.

src/lbpool_nodes.py
lamaral: Yes, export has a well defined dict, so we can clean this up.

src/lbpool_export.py
lamaral: I think setting proper timeouts could mitigate the issue.
Implementing asyncio might be a little bit of overhead for shipping data to two redis servers.

lamaral: I just pushed some changes to address this. 

I'm now setting up proper connect timeouts as well as wrapping each redis connection on a try catch, so we can skip failed ones.

src/lbpool_export.py
TuxPowered42: If it is always given, there is no need to check it.

lamaral: This is outdated, I removed it already.

TuxPowered42 (approved):

PR #226, opened 2021-05-18
Allow check for daily elastic index

zados:
As some indices are rotated daily, one should check that the index of the current day exists. 
Otherwise old indices would make this check green.

### DIFF
--- a/src/check_es_index_exists.py
+++ b/src/check_es_index_exists.py
@@ -2,7 +2,7 @@
 """InnoGames Monitoring Plugins - Check if an index exists with the
 provided prefix
 
-Copyright (c) 2019 InnoGames GmbH
+Copyright (c) 2021 InnoGames GmbH
 """
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -24,8 +24,10 @@ Copyright (c) 2019 InnoGames GmbH
 
 import sys
 from argparse import ArgumentParser, RawTextHelpFormatter
+from datetime import datetime
 from elasticsearch import Elasticsearch
 from ssl import create_default_context
+
 
 def parse_args():
     parser = ArgumentParser(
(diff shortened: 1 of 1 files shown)


src/check_es_index_exists.py
```diff
@@ -65,16 +74,19 @@ def main():
...
+    index = args.index
+    if args.daily:
+        index = index + "-" + datetime.now().strftime(args.date_format)
     es = Elasticsearch([args.host], **connect_params)
 
     # suppress print output from es function
     sys.stderr = None
-    if not es.indices.exists(args.index + '*', allow_no_indices=False):
-        print('Index not found in cluster: {}'.format(args.index))
+    if not es.indices.exists(index + '*', allow_no_indices=False):
```
lamaral, line 85: I know it's out of the scope of this change, but I think having the star in here would cause the following situation:

Let's say we have indexes `nginx` and `nginx_project`. The nginx index could be silently dropped and the check for it would never complain, because `nginx*` would match `nginx_project`, no?

lamaral (approved):
LGTM. There is just this one comment that we could improve the check, since we are already touching this part.

src/check_es_index_exists.py
zados: Afaik this may happen, I would be fine with removing that star. 
I think it was added by @elmatze79 because he wanted to catch the dates behind the index name.

So with this MR we could actually remove the star. Any objections @elmatze79 ?

elmatze79: As a quick-fix we could use `index + '-*'` here as we use the dash to separate the date, for our separations we use underscores

zados: It would not bring any value to have 
`index-2021.05.18-*` as there is nothing which could come behind it.
Where `index-*` still can cause some trouble, if people don't know about this culprit. 
If there are custom formats e.g. `index-2021.05.18.12` (if someone wants to add the hour for example) then this can be set by using `--date-format`.

So I am still on board with removing the star.

lamaral: You could:
```
    if args.daily:
        index = index + "-" + datetime.now().strftime(args.date_format)
    else:
        index = index + "-*"
```
on line 79 and then remove the star on line 85.

zados: For your proposed solution: If the `--daily` flat is used, the `*` is not applied. But if you do not have daily rotated indices, why would you need the star anyway? You know how your index is called then, I would assume.

Keeping it would be cutting corners and would be error prone for users being unaware of the issue.

PR #227, opened 2021-06-21
Clickhouse query logging

Felixoid:


src/check_clickhouse.py
```diff
@@ -172,6 +172,13 @@ def execute(self, *args, **kwargs) -> List[tuple]:
         """
         Wrapper to execute ClickHouse SQL
         """
+        if len(args) == 1:
+            logger.info('Execute query: {}'.format(args[0]))
+        elif len(args) == 2:
```
TheFRedFox, line 177: What about if this gets called with more than 2 arguments? Maybe you want to have a default `logger.info` in an `else` block. Or an error-raising if this is actually not possible.

TheFRedFox (approved):
One discussion point but early approve.

src/check_clickhouse.py
Felixoid: Good catch. It should be `>=2`

The default one is no logging, I'd say.

There's as well a way to parse kwargs, but I'd like to not cover the edge case

TheFRedFox: Sure, fair enough. Just saw that the same question would also work for `<= 1`, so if there are no args at all. It seems that it is not designed for that, so error raising might be interesting. Maybe that will already happen during the actual statement execution then though.

PR #228, opened 2021-06-30
check_switch: Ignore oobm interfaces on Procurve

lamaral:
The oobm interfaces in the procuve switches can't be "unnamed", so we
need to ignore it on the check.


lamaral:
I tested it in one of the workers. Working 10/10

TuxPowered42 (approved):

PR #230, opened 2021-08-13
Ignore unnamed ports on Cumulus

TuxPowered42:


src/check_switch.py
```diff
@@ -74,7 +74,7 @@
     'force10_mxl': re.compile(
         '^(NULL 0|ManagementEthernet [0-9]+/[0-9]+|(Vlan|Port-channel) [0-9]+)$'
     ),
-    'cumulus': re.compile('^(bridge|mgmt|lo)$'),
+    'cumulus': re.compile('^(bridge|mgmt|lo|eth)$'),
```
lamaral, line 77: Isn't this gonna ignore eth0 also?

TuxPowered42: Yes. But do we care? It's the management port. If it's down, you can't monitor the system anyway :)

lamaral: Ok, valid point.

lamaral (approved):

PR #231, opened 2021-08-17
Add ignore support to jenkins plugin checker

seqizz:


src/check_plugin_vulnerability.py
```diff
@@ -51,21 +53,36 @@ def main():
     csrf, cookie_jar = fetch_token_and_cookies(args.host, auth)
     result = post_script_data(url, SCRIPT, csrf, auth, cookie_jar)
 
-    print_and_exit(result)
+    evaluate_and_exit(result)
 
 
-def print_and_exit(result):
+def _parse_ignore_file():
+    if not isfile(IGNORE_FILE):
```
TheFRedFox, line 60: Is possible, just want to mention that try-except would also be possible. try-except would also be more secure, but that risk should not be attackable here. So, just showing the alternative:
``` python
try:
    with open(IGNORE_FILE) as fd:
        return (
            [i.rstrip('\n') for i in fd.readlines()]
        )
except FileNotFoundError:
    return []
```

src/check_plugin_vulnerability.py
```diff
@@ -51,21 +53,36 @@ def main():
     csrf, cookie_jar = fetch_token_and_cookies(args.host, auth)
     result = post_script_data(url, SCRIPT, csrf, auth, cookie_jar)
 
-    print_and_exit(result)
+    evaluate_and_exit(result)
 
 
-def print_and_exit(result):
+def _parse_ignore_file():
```
TheFRedFox, line 59: This method is right now returning different types. An empty tuple or a list.
Either change 61 returning a list too, or in 63 changing it too:
``` python
    return tuple(i.rstrip('\n') for i in fd.readlines())
```

TheFRedFox: Actually it would probably make more sense to return a `set` instead of a tuple/list.

seqizz: Now it's converted to set, either empty or filled.

src/check_plugin_vulnerability.py
seqizz: Alright, why not.

src/check_plugin_vulnerability.py
```diff
@@ -34,6 +35,7 @@
     'def critical_plugins = plugin_vulnerabilities.keySet()*.longName\n'
     'print(new groovy.json.JsonBuilder(critical_plugins))'
 )
+IGNORE_FILE = '/etc/check_plugin_vulnerability_ignore'
```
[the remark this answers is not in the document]
seqizz: There is nagios-plugins folder which comes via other packages which is not ensured by igmonplugins package as a dependency. But apparently it's used already, so I've put there.

src/check_plugin_vulnerability.py
```diff
@@ -34,6 +34,7 @@
     'def critical_plugins = plugin_vulnerabilities.keySet()*.longName\n'
     'print(new groovy.json.JsonBuilder(critical_plugins))'
 )
+IGNORE_FILE = '/etc/nagios-plugins/check_plugin_vulnerability_ignore'
```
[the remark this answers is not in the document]
seqizz: That's an implicit dependency I'd say. Without this folder, multiple nagios-related checks would be affected. I'll pass that, considering we already caught the exception of non-existence.

src/check_plugin_vulnerability.py
```diff
@@ -51,21 +52,35 @@ def main():
...
-def print_and_exit(result):
+def _parse_ignore_file():
+    try:
+        with open(IGNORE_FILE) as fd:
+            return set(i.rstrip('\n') for i in fd.readlines())
+    except FileNotFoundError:
+        return set()
+
+
+def evaluate_and_exit(result):
```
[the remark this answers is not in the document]
seqizz: Agreed, I did a slight change on wording + at least the exiting is not done in the function now.

src/check_plugin_vulnerability.py
```diff
@@ -51,21 +52,35 @@ def main():
     csrf, cookie_jar = fetch_token_and_cookies(args.host, auth)
```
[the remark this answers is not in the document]
kofrezo: We agreed in the past to not use external modules like requests to keep the dependencies to a minimum. But since there are already plenty scripts using it I think this is obsolete.

PR #233, opened 2021-08-23
Add logfile values changes check

v0devil:
Can be used to check if integration with tideways is working by checking measurements values in /var/log/tideways/daemon.log file. See https://support.tideways.com/documentation/setup/troubleshooting/checklist.html


src/check_logfile_values_change.py
```diff
@@ -0,0 +1,97 @@
...
+
+    print(f'UNKNOWN - No values in {logfile_path} or wrong regex')
+    sys.exit(3)
+
+
+def tail(f, n):
+    proc = subprocess.Popen(
+        ['tail', '-n', f'{n}', f], stdout=subprocess.PIPE
+    )
+    lines = proc.stdout.readlines()
```
TheFRedFox, line 92: Most probably we won't have any problem with it. So, just on a general note: If it would return a lot of lines, it would be better to change to function to a generator one by using `yield` and yielding the lines separately.

TheFRedFox (changes requested):
The regex usage is still not nice IMO. Additionally, some more medium comments and 2 side notes.

src/check_logfile_values_change.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    parser.add_argument(
+        '--path', help='Path to logfile', type=str
+    )
+
+    parser.add_argument(
+        '--regex',
+        help=(
+            'Regex for value to compare. Should contain named group "value", '
```
TheFRedFox, line 51: This is a must have, so it should be communicated like one:
```suggestion
            'Regex for value to compare. Must contain named group "value", '
```

src/check_logfile_values_change.py
```diff
@@ -0,0 +1,104 @@
...
+    args = parse_args()
+    logfile_path = args.path
+    num = args.num
+    lines = tail(logfile_path, num)
+    values = set()
+    p = re.compile(rf'{args.regex}')
+
+    if 'value' not in p.groupindex:
+        print('Provide regex with named group "value", e.g. v=(?P<value>\d+?)')
+        return
```
TheFRedFox, line 75: A `return` would end up in a `sys.exit(0)`, so for nagios a "OK" sign. Is that correct? For me it sounds more like it should end up in a Critical or Unknown one:
```suggestion
        print('UNKNOWN - Regex without named group "value" provided, e.g. v=(?P<value>\d+?)')
        sys.exit(3)
```
or 
```suggestion
        print('CRITICAL - Regex without named group "value" provided, e.g. v=(?P<value>\d+?)')
        sys.exit(2)
```

TheFRedFox (approved):
One minor phrasing comment and one major exit change.
Early approve though.

PR #235, opened 2021-09-07
Introduce the script to export VM data from HVs

lamaral:
This is phase 1 of the IGVM consistency check. 

We will export data from the Hypervisors to the Redis servers. On phase 2, we will use this data together with serveradmin data to find problematic VMs/HVs.

### DIFF
--- a/src/vm_export.py
+++ b/src/vm_export.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - VM Data Export Script
+
+Copyright (c) 2021 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the 'Software'), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import sys
+
+from argparse import ArgumentParser
+from libvirt import openReadOnly
+from redis import Redis
+from redis.exceptions import ConnectionError
+
+
+def parse_args():
+    parser = ArgumentParser()
+
+    parser.add_argument(
+        '-H', dest='redis_servers', action='append', required=True,
+        help='Redis servers to report the results to',
+    )
+
+    parser.add_argument(
+        '-n', dest='hostname', required=True,
+        help='Hypervisor hostname to suffix the keys with',
+    )
+
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


src/vm_export.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def get_domains_info():
+    conn = openReadOnly(None)
+    libvirt_domains = conn.listAllDomains()
+
+    domains = {}
+    for d in libvirt_domains:
+        state, reason = d.state()
+        m = re.match(r'((?P<vm_id>\d+)_)?(?P<vm_name>[\w\.\d-]+)', d.name())
```
lamaral, line 87: I'll compile this RE since it's used many times inside the loop.

lamaral:
@Woellchen also brought up one edge case of a VM defined with the same hostname but different object IDs. Both would end up with the same key in Redis, so we need to account for that.

src/vm_export.py
```diff
@@ -0,0 +1,125 @@
...
+                print(f'Results successfully sent to {server}')
+            else:
+                print(f'Errors while sending results to {server}')
+                error = True
+        except ConnectionError as e:
+            print(str(e))
+            error = True
+
+    if error:
+        exit(1)
```
TheFRedFox, line 75: We should use `sys.exit` instead of `exit`. `exit` and `quit` are more just for the interactive shell.

src/vm_export.py
```diff
@@ -0,0 +1,125 @@
...
+
+    return
```
TheFRedFox, line 77: Rather use one way. Return or exit. Not both. You can also easily use `return 1` for the `exit` above.

src/vm_export.py
TheFRedFox: I recommend to comment regular expressions, so also after months and years it is clear what it should match and what not.

TheFRedFox, line 87: Is it correct that `vm_id` is optional?

TheFRedFox, line 87: Also .. is it possible to completely get rid of regular expressions here? If `d.name()` is already returning something like `32_foo.43`, then it would also be possible to just split on `_`.

src/vm_export.py
```diff
@@ -0,0 +1,125 @@
...
+
+    error = False
+    for server in args.redis_servers:
+        try:
+            result = send_to_redis(
+                server, args.redis_password,
+                args.hostname, domains
+            )
+            if result:
+                print(f'Results successfully sent to {server}')
```
TheFRedFox, line 66: Don't we need the nagios `OK`, `WARNING`, `CRITICAL` prefixes here?

src/vm_export.py
TheFRedFox, line 75: Is it really just a warning for nagios here? Sounds to me here right now it should actually be critical, no?

TheFRedFox (commented):
Some discussion points.

src/vm_export.py
lamaral: I added a clarifying comment.

lamaral: It used to be optional, but we don't have these legacy cases anymore. I made it mandatory.

src/vm_export.py
lamaral: No, because this is not a monitoring check on it's own. It just exports data to be later evaluated by Nagios.

src/vm_export.py
lamaral: This is not a Nagios check. It returns one just to ensure it's a non-zero exit code, in case we ever start to monitor cron executions.

src/vm_export.py
lamaral: I exchanged it for a `split()`. We should not have any more hosts that don't have an object_id in the domain name.

src/vm_export.py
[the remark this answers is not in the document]
lamaral: I updated in the latest commit.

src/vm_export.py
```diff
@@ -0,0 +1,122 @@
...
+            else:
+                print(f'Errors while sending results to {server}')
+                error = True
+        except ConnectionError as e:
+            print(str(e))
+            error = True
+
+    if error:
+        sys.exit(1)
```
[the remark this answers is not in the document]
lamaral: I updated in the latest commit.

src/vm_export.py
TheFRedFox: I see .. can we then specify that explicitly in the top doc block? To really make sure that it won't get seen as a check in igmonplugins.

src/vm_export.py
TheFRedFox: Now it is actually without anything. Is that correct?

TheFRedFox (approved):
Reopened 2 conversations. Except of that looks good to me. So, early approve.

PR #237, opened 2021-09-14
lbpool_nodes: Warn if not all nodes are alive

lamaral:
The case where more than min_nodes but less than optimal_nodes are alive was returning an OK state instead of Warning. 

It was mistakenly defined like that to cover the case of clusters, but at that level, a single HWLB group is being checked and therefore, shouldn't be aware of the cluster logic.

### DIFF
--- a/src/lbpool_nodes.py
+++ b/src/lbpool_nodes.py
@@ -66,9 +66,9 @@ def main():
         if exit_code != ExitCodes.ok and i != ExitCodes.ok:
             exit_code = get_worse_exit_code(exit_code, i)
         elif exit_code != ExitCodes.ok and i == ExitCodes.ok:
-            exit_code = 0
+            exit_code = ExitCodes.ok
 
-    if exit_code == 0:
+    if exit_code == ExitCodes.ok:
         outputs.insert(0, 'LBPool is healthy')
     else:
         outputs.insert(0, 'LBPool is unhealthy')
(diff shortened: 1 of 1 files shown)


src/lbpool_nodes.py
```diff
@@ -112,14 +112,21 @@ def evaluate_state(data):
...
+                elif min_nodes is not None and (alive_nodes < min_nodes):
+                    exit_code = ExitCodes.critical
+                # More than max_nodes are alive
                 elif max_nodes is not None and (alive_nodes > max_nodes):
                     exit_code = ExitCodes.warning
+                # All nodes are alive
                 elif alive_nodes == optimal_nodes:
                     exit_code = ExitCodes.ok
+                # More than min_nodes are alive, but not all of them
                 elif min_nodes is not None and (alive_nodes >= min_nodes):
```
Woellchen, line 128: actually this case became equal to the `else` block now. so we could simply remove the `elif` now

src/lbpool_nodes.py
```diff
@@ -112,14 +112,21 @@ def evaluate_state(data):
...
         if pool_v['health_checks']:
             if pool_v['in_testtool']:
+                # All nodes are down
                 if alive_nodes == 0:
                     exit_code = ExitCodes.critical
+                # Less than min_nodes are alive
+                elif min_nodes is not None and (alive_nodes < min_nodes):
+                    exit_code = ExitCodes.critical
+                # More than max_nodes are alive
                 elif max_nodes is not None and (alive_nodes > max_nodes):
```
Woellchen, line 122: while at it, this case can also be removed because it is also handled in `else`

src/lbpool_nodes.py
TheFRedFox: In general and by syntax you are right .. but I would actually suggest to leave it here for the approach of rather be explicit than implicit. To find out in the code what it would return for that case would be horrible without that `elif`.

Woellchen: sure we can do that. but then lets rethink the `else`. is it even needed or should it throw some exception like "we shouldn't ever get here because all cases are handled above"?

Woellchen: i also just realized that these cases dont produce a user-facing message that is put into `output` array. so for the user it really doesn't make a difference. if we keep all cases then we should at least add some messages here no?

TheFRedFox: Yep, that's rather the direction I guess we should go. I think the pattern here goes to the Strategy pattern. But to do that now would be way out of scope of the PR.
But, that's why I would just not simplify the if-else right now, to still show this structure.

lamaral: The message for the user is produced on line 126. 
This part is running within the scope of a single HWLB group and we could have more. After we evaluate all HWLB groups, we produce a final message on line 71.

I will make the last else case an exception.

src/lbpool_nodes.py
lamaral: I would rather keep those cases explicitly defined, to make it better for other admins to understand as well as help if we decide to change the exit code for this one case.

src/lbpool_nodes.py
TheFRedFox: For me that's gtg. In general this function is doing too much and is hard to read, but that's not in scope of PR. So, for future development, check how we can make that code nicer to read. 
For me that PR is approvable though.

TheFRedFox (approved):

Woellchen (approved):

src/lbpool_nodes.py
Woellchen: think the code matches our discussion now :)

PR #239, opened 2021-09-22
Include hidden warnings too on Jenkins

seqizz:
Users can "ignore" some warnings on Jenkins management interface. We'd like to have our ignore file to be the only source of truth in this case. New groovy code doesn't care if user has ignored it on UI.

### DIFF
--- a/src/check_plugin_vulnerability.py
+++ b/src/check_plugin_vulnerability.py
@@ -28,12 +28,20 @@ import argparse
 import requests
 from sys import exit
 
-SCRIPT = (
-    'def warnings_monitor = new jenkins.security.UpdateSiteWarningsMonitor()\n'  # NOQA E501 Don't wrap because of groovy code
-    'def plugin_vulnerabilities = warnings_monitor.getActivePluginWarningsByPlugin()\n'  # NOQA E501 Don't wrap because of groovy code
-    'def critical_plugins = plugin_vulnerabilities.keySet()*.longName\n'
-    'print(new groovy.json.JsonBuilder(critical_plugins))'
-)
+SCRIPT = """
+    import jenkins.security.*
+    ExtensionList<UpdateSiteWarningsConfiguration> configurations = ExtensionList.lookup(UpdateSiteWarningsConfiguration.class);
+    UpdateSiteWarningsConfiguration configuration = configurations.get(0);
+    problematic_plugins = []
+    current_warnings = configuration.getApplicableWarnings()
+    current_warnings.each {
+        if (it.type.name() == 'PLUGIN') {
+            problematic_plugins.add(it.component)
+        }
+    }
+    print(new groovy.json.JsonBuilder(problematic_plugins.unique()))
+"""  # NOQA E501 Don't wrap because of groovy code
+
 IGNORE_FILE = '/etc/nagios-plugins/check_plugin_vulnerability_ignore'
 
 


src/check_plugin_vulnerability.py
```diff
@@ -28,12 +28,20 @@
...
-    'def warnings_monitor = new jenkins.security.UpdateSiteWarningsMonitor()\n'  # NOQA E501 Don't wrap because of groovy code
-    'def plugin_vulnerabilities = warnings_monitor.getActivePluginWarningsByPlugin()\n'  # NOQA E501 Don't wrap because of groovy code
-    'def critical_plugins = plugin_vulnerabilities.keySet()*.longName\n'
-    'print(new groovy.json.JsonBuilder(critical_plugins))'
-)
+SCRIPT = """
+    import jenkins.security.*
+    ExtensionList<UpdateSiteWarningsConfiguration> configurations = ExtensionList.lookup(UpdateSiteWarningsConfiguration.class);
+    UpdateSiteWarningsConfiguration configuration = configurations.get(0);
+    problematic_plugins = []
```
TheFRedFox, line 35: I think the def is missing at the beginning:
```suggestion
    def problematic_plugins = []
```

src/check_plugin_vulnerability.py
```diff
@@ -28,12 +28,20 @@
...
+    current_warnings = configuration.getApplicableWarnings()
```
TheFRedFox, line 36: Same as above:
```suggestion
    def current_warnings = configuration.getApplicableWarnings()
```

src/check_plugin_vulnerability.py
```diff
@@ -28,12 +28,20 @@
...
+    current_warnings.each {
+        if (it.type.name() == 'PLUGIN') {
+            problematic_plugins.add(it.component)
+        }
+    }
```
TheFRedFox, line 41: You can also write it like that:
```suggestion
    def problematic_plugins = current_warnings.findAll { it.type.name() == 'PLUGIN') }.unique()
```

src/check_plugin_vulnerability.py
```diff
@@ -28,12 +28,20 @@
...
+    print(new groovy.json.JsonBuilder(problematic_plugins.unique()))
```
TheFRedFox, line 42: Better Practice is to have the import at the beginning and explicit:
``` groovy
import groovy.json.JsonBuilder
...
print(new JsonBuilder(...))
```

src/check_plugin_vulnerability.py
```diff
@@ -28,12 +28,20 @@
...
 from sys import exit
 
-SCRIPT = (
-    'def warnings_monitor = new jenkins.security.UpdateSiteWarningsMonitor()\n'  # NOQA E501 Don't wrap because of groovy code
-    'def plugin_vulnerabilities = warnings_monitor.getActivePluginWarningsByPlugin()\n'  # NOQA E501 Don't wrap because of groovy code
-    'def critical_plugins = plugin_vulnerabilities.keySet()*.longName\n'
-    'print(new groovy.json.JsonBuilder(critical_plugins))'
-)
+SCRIPT = """
+    import jenkins.security.*
```
TheFRedFox, line 32: It's good practice to use explicit non grouped imports:
```suggestion
    import jenkins.security.UpdateSiteWarningsConfiguration
    import jenkins.security.ExtensionList
```

TheFRedFox (commented):
Some Groovy tips.

src/check_plugin_vulnerability.py
seqizz: Which is less readable imho. I'll pass.

src/check_plugin_vulnerability.py
seqizz: Done.

PR #242, opened 2021-11-10
Use HEAD method to test reachability

lacunoc:


TheFRedFox:
Can't we even better add an optional script argument for the http method which defaults to 'HEAD'? I have a feeling that in the future there will be a request we need to extra do with GET maybe.

lacunoc:
Sure, added the argument and set the default to HEAD

lamaral (approved):

PR #243, opened 2022-01-13
Add check for huge pages

meikomeik:


src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+
+import sys
+
+from argparse import ArgumentParser
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--value',
+                        help='The amout of huge pages to expect')
```
lamaral, line 34: Please make the argument required, as the script can't run without it.

src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+    if not meminfo:
+        print('Could not parse /proc/meminfo')
+        sys.exit(3)
+
+    meminfo_huge_pages = meminfo['HugePages_Total']
+
+    if int(meminfo_huge_pages) < int(args.value):
+        exit_status = 2
+        output = (
+            f"/proc/meminfo['HugePages_Total'] == {meminfo_huge_pages} "
```
lamaral, line 49: Please prefix the output with `CRITICAL - `.

src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+            f"which is smaller than {args.value}"
+        )
+    else:
+        exit_status = 0
+        output = (
+            f"/proc/meminfo['HugePages_Total'] is {meminfo_huge_pages} "
```
lamaral, line 55: Please prefix the output with `OK - `.

src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+        print('Could not parse /proc/meminfo')
+        sys.exit(3)
+
+    meminfo_huge_pages = meminfo['HugePages_Total']
+
+    if int(meminfo_huge_pages) < int(args.value):
+        exit_status = 2
+        output = (
+            f"/proc/meminfo['HugePages_Total'] == {meminfo_huge_pages} "
+            f"which is smaller than {args.value}"
```
lamaral, line 50: Please make this line consistent with line 56 below.

src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+    else:
+        exit_status = 0
+        output = (
+            f"/proc/meminfo['HugePages_Total'] is {meminfo_huge_pages} "
+            f"which is >= {args.value}"
```
lamaral, line 56: Please make this line consistent with line 50 above.

src/check_huge_pages.py
lamaral: I would also go as far as saying let's make this `--min_value`, so if we ever decide to check for max value, we don't have to change the script calls.

TheFRedFox, line 34:
```suggestion
                        help='The amount of huge pages to expect')
```

lamaral: Also a typo in `amount`

src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+# THE SOFTWARE.
+
+import sys
+
+from argparse import ArgumentParser
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--value',
```
TheFRedFox, line 33: If there is no default, but based on the script it looks like a required argument, maybe it should be a positional parameter then? At least it should be configured as required.

lamaral (changes requested):

src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+
+def main(args):
+    meminfo = read_proc_meminfo()
+    if not meminfo:
+        print('Could not parse /proc/meminfo')
+        sys.exit(3)
+
+    meminfo_huge_pages = meminfo['HugePages_Total']
+
+    if int(meminfo_huge_pages) < int(args.value):
```
TheFRedFox, line 46: If `args.value` is of type `int` a cast is not needed anymore.

src/check_huge_pages.py
```diff
@@ -0,0 +1,77 @@
...
+    """Return /proc/meminfo as a dict"""
+    meminfo_dict = {}
+    try:
+        with open('/proc/meminfo', 'r') as proc_file:
+            for line in proc_file.readlines():
+                name, value = line.strip().replace(' kB', '').split(':', 1)
+                meminfo_dict[name] = int(value)
+            return meminfo_dict
+    except (OSError, IOError):
+        return None
```
TheFRedFox, line 73: Outside you are interpreting `None` return as parsing error. However, a `ValueError` should also then get caught here. `int(value)` and also the `name, value = line.[...].split(':', 1)` could raise a `ValueError` here.

src/check_huge_pages.py
TheFRedFox, line 33: Also, shouldn't it be of type `int`?

src/check_huge_pages.py
TheFRedFox, line 49: Prefixes like already Luiz said.

src/check_huge_pages.py
meikomeik: prefixed

src/check_huge_pages.py
meikomeik: Added your suggestions

src/check_huge_pages.py
meikomeik: Added prefix

src/check_huge_pages.py
meikomeik: Should be consistent now

src/check_huge_pages.py
meikomeik: Should be consistent now

src/check_huge_pages.py
meikomeik: fixed typo

src/check_huge_pages.py
meikomeik: Removed not needed int

src/check_huge_pages.py
meikomeik: changed it to int

src/check_huge_pages.py
meikomeik: Done

src/check_huge_pages.py
meikomeik: Added ValueError as part of the possible errors though each line always contains a colon folowed by a number so it should never throw a ValueError I think. Anyhow, why not also add ValueError

src/check_huge_pages.py
meikomeik: Made the argument required

src/check_huge_pages.py
```diff
@@ -0,0 +1,75 @@
...
+    parser = ArgumentParser()
+    parser.add_argument('--min_value', type=int, required=True,
+                        help='The amount of huge pages to expect')
+    return parser.parse_args()
+
+
+def main(args):
+    meminfo = read_proc_meminfo()
+    if not meminfo:
+        print('Could not parse /proc/meminfo')
```
lamaral, line 41: I missed the `UNKNOWN - ` prefix here.

lamaral (approved):
There is just one more thing that I didn't see before. 
I'm pre-approving. Feel free to merge after fixing this last comment.

src/check_huge_pages.py
meikomeik: fixed this one as well

PR #245, opened 2022-02-01
Add check_postgres_waiting_changes

seqizz:
Just a simple check to catch manually touched pg instances, _normally_ not the case thanks to puppet.


src/check_postgres_waiting_changes.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def main():
+    """Main entrypoint for script"""
+
+    pending = _get_configs_pending_restarts()
+
+    if pending:
+        print('WARNING - Waiting for restart for following configs: {}'.format(
+            ', '.join(pending)))
```
TheFRedFox, line 53: Just a suggestion how to style it IMO nicer.
```suggestion
        print(
            'WARNING - Waiting for restart for following configs: {}'
            .format(', '.join(pending))
        )
```

src/check_postgres_waiting_changes.py
```diff
@@ -0,0 +1,61 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import subprocess
+
+from sys import exit
+
+
+def _get_configs_pending_restarts():
```
TheFRedFox, line 32: As a helper function this needs to go under the main function.

TheFRedFox (approved):
The helper function needs to go under the main function. But I early approve as that should be a trivial fix and the rest looks good. :)

src/check_postgres_waiting_changes.py
seqizz: Arrrgh, I thought "oh, the new coding style said... above, I guess" :facepalm:

src/check_postgres_waiting_changes.py
seqizz: Pushed the suggestion. This looks like black's styling, didn't use it for a while though.

PR #246, opened 2022-02-23
Extend redis check for identical roles assignment

konstantinbest:
Currently, we do not know if a node from the Redis cluster has two master or slave roles. This is not important for the functionality of the cluster, but the load on the double master is increased and we would like to make this visible in Nagios with this change.


seqizz (changes requested):
Could you tell why this is a warning, in the warning?
E.g. `'WARNING - node has two {master_state} roles, this will cause increased load on ...'`

Or at least as an inline comment, moving the PR description into here.

seqizz (approved):

PR #247, opened 2022-04-03
Add script to check for Atlassian license seats

TheFRedFox:


src/check_atlassian_license_seats.py
```diff
@@ -0,0 +1,202 @@
...
+    parser.add_argument(
+        '--consumer-secret',
+        help='consumer secret for oauth authentication'
+    )
+    parser.add_argument('--private-key', help='private key for oauth')
+    parser.add_argument(
+        '--passphrase',
+        help='possible passphrase for the private key'
+    )
+    warn_group = parser.add_mutually_exclusive_group()
```
kofrezo, line 79: Whats the benefit of using a mutually exclusive group when here is just one argument ?

src/check_atlassian_license_seats.py
```diff
@@ -0,0 +1,202 @@
...
+    warn_group.add_argument('--warn-perc-threshold', type=float)
+    crit_group = parser.add_mutually_exclusive_group()
+    crit_group.add_argument('--crit-perc-threshold', type=float)
+    return parser.parse_args()
+
+
+def main(args) -> 'ExitCode':
```
kofrezo, line 86: Are the single quotes for `'ExitCode'` necessary here ?

src/check_atlassian_license_seats.py
```diff
@@ -0,0 +1,202 @@
...
+
+    return requests.request(
+        method, base_url + endpoint, auth=auth,
+        params=params
+    )
+
+
+def get_oauth1session(consumer_key, consumer_secret, private_key, passphrase):
+    from Crypto.PublicKey import RSA
+    from requests_oauthlib import OAuth1
```
kofrezo, line 162: If these are non standard modules required for the script you could think of mentioning them in the header of the file or even the argument parser description to make it easier for people to view the requirements.

src/check_atlassian_license_seats.py
```diff
@@ -0,0 +1,202 @@
...
+    with open(private_key, 'r') as fd:
+        rsa_key = RSA.importKey(fd.read(), passphrase)
+
+    return OAuth1(
+        client_key=consumer_key, client_secret=consumer_secret,
+        signature_method='RSA-SHA1', rsa_key=rsa_key
```
kofrezo, line 168: Can we use something different than `RSA-SHA1` ?

I think this is not considered safe anymore, we should use SHA256 if possible.

kofrezo: Would it be possible to use ECDSA instead of RSA keys. They are more secure and faster (in computation for the crypto).

src/check_atlassian_license_seats.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python
```
kofrezo, line 1: While all our systems should default to Python 3 by default one could still override this. So if you want to be 100% sure to use Python3 you are better off using Python3.

kofrezo, line 1:
```suggestion
#!/usr/bin/env python3
```

src/check_atlassian_license_seats.py
TheFRedFox: No, it is given by the Atlassian systems not by me.

src/check_atlassian_license_seats.py
TheFRedFox: Yes, as the class is not defined yet here.

src/check_atlassian_license_seats.py
TheFRedFox: It is mentioned in line 41.

src/check_atlassian_license_seats.py
TheFRedFox: None, I started with more arguments and decided to remove them for now. Seems I forgot to get rid of the mutual exclusive group though.

kofrezo (approved):

PR #248, opened 2022-04-06
bird_protocols: Add support for Pipe protocols

lamaral:


src/check_bird_protocols.py
```diff
@@ -86,7 +86,9 @@ def check_birdc_protocols(path):
             ret = check_ospf_protocol(protocol)
         elif protocol['type'] == 'RPKI':
             ret = check_rpki_protocol(protocol)
-        elif protocol['type'] in ['Device', 'Kernel', 'Static', 'Direct']:
+        elif protocol['type'] in [
+            'Device', 'Direct', 'Kernel', 'Pipe', 'Static'
+        ]:
```
TheFRedFox, line 91: When you are already touching it: As the content is static here, we could use a set and would make it a tiny bit faster.

```suggestion
        elif protocol['type'] in {
            'Device', 'Direct', 'Kernel', 'Pipe', 'Static'
        }:
```

TuxPowered42 (approved):
Dorime

src/check_bird_protocols.py
lamaral: Done.

PR #249, opened 2022-04-06
Add script to check event limits for sentry.io teams

raimund-s:
To test this script you need to obtain a sentry.io api key and organizaiton


src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+                else:
+                    events = None
+                    team['summed_events'] = None
+
+                if args.verbose:
+                    print('Team: {}, Project: {}, Key: {}, Limit: {}'.format(
+                        team['name'], project['name'], dsn['name'],
+                        dsn['rateLimit']))
+
+            '''Check if this team is over the team limit'''
```
TheFRedFox, line 91: These are comments and should not be python docstrings:
```suggestion
            # Check if this team is over the team limit
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+                args.organization, project['slug'], args.bearer)
+            for dsn in dsns:
+                if dsn['rateLimit']:
+                    if type(events) is int:
+                        events += int(dsn['rateLimit']['count'] * 60 /
+                                      (dsn['rateLimit']['window']))
+                    if type(team['summed_events']) is int:
+                        team['summed_events'] += int(
+                                dsn['rateLimit']['count'] * 60 /
+                                (dsn['rateLimit']['window']))
```
TheFRedFox, line 81: Is it needed to catch division by zero here?

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+                else:
+                    events = None
+                    team['summed_events'] = None
```
TheFRedFox, line 84: When we are here, can't we break out of the for loop already?

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
+
+from argparse import ArgumentParser
+import requests
```
kofrezo, line 4: Is it feasible to work without requests but use built-in urllib to keep dependencies to a minimum ?

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+import sys
+
+
+p = ArgumentParser()
+p.add_argument('-b', '--bearer', type=str, required=True, dest='bearer',
```
kofrezo, line 9: FYI: Required arguments are usually positional arguments.

kofrezo, line 9: FYI: The `dest` is implicit too. It defaults to the name (2nd Argument)

src/check_sentryio_limits.py
TheFRedFox, line 81: What is the reason to try to squeeze everything in one line:
```suggestion
                        rate_count = dsn['rateLimit']['count']
                        rate_window = dsn['rateLimit']['window']
                        rate_sum = int(rate_count * 60 / rate_window)
                        team['summed_events'] += rate_sum
```
Would make it much more readable.

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
+
```
kofrezo, line 2: Please add the license and a description/purpose here.

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+            '''Check if this team is over the team limit'''
+            if args.perteamlimit and not team['summed_events']:
+                exit = 1
+                print('WARNING: unlimited events configured for team: {}\
+                    '.format(team['name']))
+            elif args.perteamlimit and team['summed_events'] and \
+                    team['summed_events'] > args.perteamlimit:
+                exit = 1
+                print('WARNING: {} are configure of {} allowed for team: {}\
```
TheFRedFox, line 99:
```suggestion
                print('WARNING: {} are configured, but {} are allowed for team: {}\
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+                    '.format(team['summed_events'], args.perteamlimit,
```
TheFRedFox, line 100: The backslash is not needed:
```suggestion
                print('WARNING: {} are configure of {} allowed for team: {}'
                      .format(team['summed_events'], args.perteamlimit,
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+
+p = ArgumentParser()
+p.add_argument('-b', '--bearer', type=str, required=True, dest='bearer',
+               help='A sentry api token with at least read permissions')
+p.add_argument('-o', '--organization', type=str, required=True,
+               dest='organization', help='The organization slug for the\
+               sentry.io organization to be queried')
+p.add_argument('-t', '--team', type=str, action='append', dest='teams',
+               help='Only check this team, can be added repeatetly')
+p.add_argument('-g', '--globallimt', type=int, dest='globallimit',
```
kofrezo, line 16: `--global-limit` (dash as separator) is best practise.

Mind the typo.

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+                if args.verbose:
+                    print('Team: {}, Project: {}, Key: {}, Limit: {}'.format(
+                        team['name'], project['name'], dsn['name'],
+                        dsn['rateLimit']))
+
+            '''Check if this team is over the team limit'''
+            if args.perteamlimit and not team['summed_events']:
+                exit = 1
+                print('WARNING: unlimited events configured for team: {}\
+                    '.format(team['name']))
```
TheFRedFox, line 95: Uppercase after colon and backslash is not needed:

```suggestion
                print('WARNING: Unlimited events configured for team: {}'
                      .format(team['name']))
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+                    team['summed_events'] > args.perteamlimit:
+                exit = 1
+                print('WARNING: {} are configure of {} allowed for team: {}\
+                    '.format(team['summed_events'], args.perteamlimit,
+                    team['name']))
+
+    '''Check if global limit is reached'''
+    if args.globallimit and not events:
+        exit = -1
+        print('WARNING: unlimited events configured in total')
```
TheFRedFox, line 106: Uppercase after colon:
```suggestion
        print('WARNING: Unlimited events configured in total')
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+               help='A sentry api token with at least read permissions')
+p.add_argument('-o', '--organization', type=str, required=True,
+               dest='organization', help='The organization slug for the\
+               sentry.io organization to be queried')
+p.add_argument('-t', '--team', type=str, action='append', dest='teams',
+               help='Only check this team, can be added repeatetly')
+p.add_argument('-g', '--globallimt', type=int, dest='globallimit',
+               help='If the total amount if events per minute is higher \
+               than this limit the script will exit with a warning and \
+               a the exit code 1, for nrpe compatibility')
```
kofrezo, line 19: Extra `a`

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+p.add_argument('-p', '--perteamlimit', type=int, dest='perteamlimit',
+               help='If any teams\' projects\' keys summed up limits is \
```
kofrezo, line 21: When having apostrophes it is best practise to use double quotes for the string

src/check_sentryio_limits.py
TheFRedFox, line 19:
```suggestion
               help='If the total amount if events per minute is higher '
               'than this limit the script will exit with a warning and '
               'a the exit code 1, for nrpe compatibility')
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+               higher than this, or not set the script will exit with a \
+               warning and the exit code 1')
+p.add_argument('-v', '--verbose', action='store_true', dest='verbose',
+               help='Print detailed stats')
+args = p.parse_args()
+
+
+def get_teams(organization, bearer):
+    '''Return a list of all teams in the account'''
```
kofrezo, line 30: Use double quotes for doc strings (convention)

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    teams = []
+    headers = {'Authorization': 'Bearer  {}'.format(bearer)}
```
kofrezo, line 32:
```suggestion
    headers = {'Authorization': f'Bearer  {bearer}'}
```
With Python 3.7 we can use the more readable f-string format :-)

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    r = requests.get('https://sentry.io/api/0/organizations/{}/teams/'.format(
```
kofrezo, line 33: Is it worth to make the URL configurable (argparse) and default to this ?

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+p.add_argument('-p', '--perteamlimit', type=int, dest='perteamlimit',
+               help='If any teams\' projects\' keys summed up limits is \
+               higher than this, or not set the script will exit with a \
+               warning and the exit code 1')
+p.add_argument('-v', '--verbose', action='store_true', dest='verbose',
+               help='Print detailed stats')
+args = p.parse_args()
+
+
+def get_teams(organization, bearer):
```
kofrezo, line 29:
```suggestion
def get_teams(organization: str, bearer: str) -> str:
```

We can already use type hints :-)

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    dsns = r.json()
+
+    return(dsns)
+
+
+def main():
+
+    teams = get_teams(args.organization, args.bearer)
+
+    '''Filter teams if list is provided'''
```
TheFRedFox, line 54: Comments are always appreciated. :)

TheFRedFox (changes requested):
Thx for the new script. :)
Unfortunately, it does not follow our Python Scripts Structure Guide. There are also some typos and outdated python usages, which I highlighted. The overall code complexity in the `main` function is also quite high. For me the WTFPM is too high here, so that we should reduce the code complexity there.
So, have fun with the first review iteration. ;)

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+        organization), headers=headers)
+    teams = r.json()
+    return(teams)
+
+
+def get_dsns_from_project(organization, project, bearer):
+    '''Retun a list of DSNs for the project'''
+    dsns = []
+    headers = {'Authorization': 'Bearer {}'.format(bearer)}
+    r = requests.get('https://sentry.io/api/0/projects/{}/{}/keys/'.format(
```
kofrezo, line 43: See comment about URL in get_teams.

Surely we can have a central string for the domain and concatenate the URL

E.g.: `API_URL = 'https://sentry.io/api'`

kofrezo (changes requested):
@TheFRedFox was faster. I'll just submit what I have so far.

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
raimund-s: done

src/check_sentryio_limits.py
raimund-s: not sure, urllib seems kind of a hassle compared to requests, only used the urllib2 interface which was kind of a mess compared to request

src/check_sentryio_limits.py
raimund-s: fixed

src/check_sentryio_limits.py
raimund-s: done

src/check_sentryio_limits.py
raimund-s: changed

src/check_sentryio_limits.py
raimund-s: changed

src/check_sentryio_limits.py
raimund-s: do these work with builtins or do I need another module for these?

src/check_sentryio_limits.py
raimund-s: done

src/check_sentryio_limits.py
raimund-s: clarified it

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+
+from argparse import ArgumentParser
+import requests
+import sys
+
+
+p = ArgumentParser()
```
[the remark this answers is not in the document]
raimund-s: moved

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    dsns = []
+    headers = {'Authorization': 'Bearer {}'.format(bearer)}
+    r = requests.get('https://sentry.io/api/0/projects/{}/{}/keys/'.format(
+        organization, project), headers=headers)
+    dsns = r.json()
+
+    return(dsns)
+
+
+def main():
```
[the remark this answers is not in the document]
raimund-s: changed

src/check_sentryio_limits.py
raimund-s: done

src/check_sentryio_limits.py
raimund-s: is -> str correct? the return value should be a dict

src/check_sentryio_limits.py
raimund-s: I actually tried to avoid any positional arguments to be more explicit, but I'm not aware of any convention here, then I need to find the correct order of the two :)

src/check_sentryio_limits.py
kofrezo: My bad. No than it should be `dict`

raimund-s:
Added changes from most of the comments, thank you for all your feedback so far, @TheFRedFox any suggestion on how to split the main logic up properly?

src/check_sentryio_limits.py
kofrezo: I avoided the word convention on purpose. I think both is fine. I see your point. Lets keep it as it is then.

src/check_sentryio_limits.py
kofrezo: `requests` it is then. We use it anyway in many other scripts already.

src/check_sentryio_limits.py
raimund-s: added

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+p.add_argument('-o', '--organization', type=str, required=True,
+               dest='organization', help='The organization slug for the\
+               sentry.io organization to be queried')
+p.add_argument('-t', '--team', type=str, action='append', dest='teams',
+               help='Only check this team, can be added repeatetly')
+p.add_argument('-g', '--globallimt', type=int, dest='globallimit',
+               help='If the total amount if events per minute is higher \
+               than this limit the script will exit with a warning and \
+               a the exit code 1, for nrpe compatibility')
+p.add_argument('-p', '--perteamlimit', type=int, dest='perteamlimit',
```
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+               help='If any teams\' projects\' keys summed up limits is \
+               higher than this, or not set the script will exit with a \
+               warning and the exit code 1')
```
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+
+def get_dsns_from_project(organization, project, bearer):
+    '''Retun a list of DSNs for the project'''
+    dsns = []
+    headers = {'Authorization': 'Bearer {}'.format(bearer)}
+    r = requests.get('https://sentry.io/api/0/projects/{}/{}/keys/'.format(
+        organization, project), headers=headers)
+    dsns = r.json()
+
+    return(dsns)
```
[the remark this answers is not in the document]
raimund-s: canged

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    teams = []
+    headers = {'Authorization': 'Bearer  {}'.format(bearer)}
+    r = requests.get('https://sentry.io/api/0/organizations/{}/teams/'.format(
+        organization), headers=headers)
+    teams = r.json()
+    return(teams)
+
+
+def get_dsns_from_project(organization, project, bearer):
+    '''Retun a list of DSNs for the project'''
```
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def get_teams(organization, bearer):
+    '''Return a list of all teams in the account'''
+    teams = []
+    headers = {'Authorization': 'Bearer  {}'.format(bearer)}
+    r = requests.get('https://sentry.io/api/0/organizations/{}/teams/'.format(
+        organization), headers=headers)
+    teams = r.json()
+    return(teams)
```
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: changed

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+
+    '''Filter teams if list is provided'''
+    if args.teams:
+        filtered_teams = []
+        for team in teams:
+            if team['slug'] in args.teams or team['name'] in args.teams:
+                filtered_teams.append(team)
+        teams = filtered_teams
+
+    '''Set exit code and global event counter'''
```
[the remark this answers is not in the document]
raimund-s: changed

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    exit = 0
+    events = 0
+
+    '''Iterate over teams, their projects and sum up their keys' rates'''
```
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    teams = r.json()
+    return(teams)
+
+
+def get_dsns_from_project(organization, project, bearer):
+    '''Retun a list of DSNs for the project'''
+    dsns = []
+    headers = {'Authorization': 'Bearer {}'.format(bearer)}
+    r = requests.get('https://sentry.io/api/0/projects/{}/{}/keys/'.format(
+        organization, project), headers=headers)
```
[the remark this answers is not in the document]
raimund-s: broke out the api url it's not to readable now in both cases

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+    exit = 0
+    events = 0
+
+    '''Iterate over teams, their projects and sum up their keys' rates'''
+    for team in teams:
+        for project in team['projects']:
+            team['summed_events'] = 0
+
+            dsns = get_dsns_from_project(
+                args.organization, project['slug'], args.bearer)
```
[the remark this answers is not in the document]
raimund-s: I guess about 10 teams with 5-10 projects each, which need get checked for the api keys, this only needs to run once a day or so after the setup is done

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+            for dsn in dsns:
+                if dsn['rateLimit']:
+                    if type(events) is int:
```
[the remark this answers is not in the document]
raimund-s: I get a None in the dict from sentry if there is no rate limit set, if this is the case we always fail unless we just use the script with -v to get the limits but not check against any 

I need one global counter which should be a non-numeric value if any project has no limit set, otherwise it should be the sum of all the limits
The second counter needed is one per team for all the teams projects which should also be none if any project is limitless, otherwise the sum

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+            elif args.perteamlimit and team['summed_events'] and \
+                    team['summed_events'] > args.perteamlimit:
+                exit = 1
+                print('WARNING: {} are configure of {} allowed for team: {}\
+                    '.format(team['summed_events'], args.perteamlimit,
+                    team['name']))
+
+    '''Check if global limit is reached'''
+    if args.globallimit and not events:
+        exit = -1
```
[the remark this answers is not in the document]
raimund-s: fixed, this was a typo, should have been 1

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+        print('WARNING: unlimited events configured in total')
+    elif args.globallimit and events > args.globallimit:
+        exit = 1
+        print('WARNING: {} of {} events are configured in total'.format(
+            events, args.globallimit))
+    elif events and exit == 0:
+        print('OK: {} events are configured in total'.format(events))
+    elif not args.globallimit:
+        print('OK: no global event limit')
```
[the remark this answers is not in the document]
raimund-s: no we can either check for only the team limits, only the global or the combination, only when neither team nor global limit is to be checked the script will exit 0 in any case an just print the results if -v is used to get an overview about all projects and their limits

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: Is there an infinit int which is always bigger than any other and can be added to? Then I can get rid of these type checks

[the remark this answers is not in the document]
raimund-s: I split the check for unlimited out into a separate variable to make it more readable

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+
+    teams = get_teams(args.organization, args.bearer)
+
+    '''Filter teams if list is provided'''
+    if args.teams:
+        filtered_teams = []
+        for team in teams:
+            if team['slug'] in args.teams or team['name'] in args.teams:
+                filtered_teams.append(team)
+        teams = filtered_teams
```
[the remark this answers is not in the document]
raimund-s: changed and tested

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: I took another look at the api but could not find any other way to retrieve the limits for all dsns other then query each project separately. I might need to slow down the script if we run into an api rate limit here at some point on the sentry side.

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: probably not needed anymore with the unlimited_events bools added now

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: I renamed the global limit to organisation limit to avoid the term global in the script

raimund-s:
@TheFRedFox could you take another look at the man function, I'm not getting around the 3 level nesting but could split it to some degree to make it more readable. 
For the indentation level the 5th is now verbose mode printing only, do you think it is worth it to use a separate function to avoid that?

src/check_sentryio_limits.py
```diff
@@ -65,13 +65,10 @@ def main():
...
-    # Filter teams if provided via arguments
+    # Filter teams to the ones provided via arguments
     if args.teams:
-        filtered_teams = []
-        for team in teams:
-            if team['slug'] in args.teams or team['name'] in args.teams:
-                filtered_teams.append(team)
-        teams = filtered_teams
+        teams = list(filter(lambda t: t['slug'] in args.teams or t['name']
+                     in args.teams, teams))
```
TheFRedFox, line 72: You could try this style, which IMO is more readable:
``` python
        teams = list(filter(
            lambda t: t['slug'] in args.teams or t['name'] in args.teams,
            teams
        ))
```
or
``` python
        teams = list(
            filter(
                lambda t: t['slug'] in args.teams or t['name'] in args.teams,
                teams
            )
        )
```

src/check_sentryio_limits.py
```diff
@@ -46,10 +46,11 @@ def args_parse():
                    'to be queried')
     p.add_argument('-t', '--teams', action='append', dest='teams',
                    help='Only check this team, can be added repeatedly')
-    p.add_argument('-g', '--global-limit', type=int, dest='globallimit',
-                   help='If the total amount if events per minute is higher '
-                   'than this limit the script will exit with a warning and '
-                   'the exit code 1, for nrpe compatibility')
+    p.add_argument('-l', '--organization-limit', type=int,
+                   dest='organizationlimit', help='If the total amount of '
```
TheFRedFox, line 46: What is the reason, that you are not going with the default snake_casing? Which would also be the convention in python to break down compound words.

Would it be possible to get back to the default snake_casing?

src/check_sentryio_limits.py
```diff
@@ -70,10 +71,9 @@ def main():
         teams = list(filter(lambda t: t['slug'] in args.teams or t['name']
                      in args.teams, teams))
 
-    # Set exit code and global event counter
+    # Set exit code and organnization wide event counter
```
TheFRedFox: `organization`
btw.
are we going for British or American English? As American English would be `organisation`.

src/check_sentryio_limits.py
```diff
@@ -70,10 +71,9 @@ def main():
...
     exit = 0
-    global_summed_events = 0
-    global_unlimited_events = False
+    organization = {'summed_events': 0, 'unlimited_events': False}
```
TheFRedFox, line 76: Variable name is misleading, as it does hold any information about the organisation itself, but rather event.
Would it be possible to name the variable `event_data` or the like?

TheFRedFox: Like the idea to store it like that definitely more. 👍🏼

src/check_sentryio_limits.py
```diff
@@ -71,64 +71,87 @@ def main():
...
+    # Initiate exit code and organnization wide event counter
     exit = 0
     organization = {'summed_events': 0, 'unlimited_events': False}
 
-    # Iterate over teams, their projects and sum up their keys' rates
+    # Iterate over teams and their projects to sum up their keys' rates
     for team in teams:
+
+        if args.verbose:
+            print(f"\nTeam \"{team['name']}\", checking for projects")
```
TheFRedFox: What's the reason that you start with a new line here?

src/check_sentryio_limits.py
```diff
@@ -71,64 +71,87 @@ def main():
         teams = list(filter(lambda t: t['slug'] in args.teams or t['name']
```
TheFRedFox, line 71: Just general feedback to that commit as reviewer role:
That's too many changes around the whole space, so, will skip that commit and will check the new structure on the big change overview.

src/check_sentryio_limits.py
```diff
@@ -125,12 +125,12 @@ def main():
             if team['unlimited_events']:
                 exit = 1
                 print('WARNING: Unlimited events configured for team: {}'
-                      .format(team['name']))
+                      .format(team['slug']))
```
TheFRedFox: As you are already using the `f-string`-format:
``` python
print(f'WARNING: ...: {team[slug]}')
```
is possible to use.

src/check_sentryio_limits.py
TheFRedFox: Just to put my 2 cents here too:
I also think that `requests` is such a highly used and spread module, that we can freely use it here. :)

src/check_sentryio_limits.py
```diff
@@ -29,32 +29,38 @@
...
-               'the exit code 1, for nrpe compatibility')
-p.add_argument('-p', '--per-team-limit', type=int, dest='perteamlimit',
-               help="If any teams' projects' keys summed up limits is "
-               "higher than this, or not set the script will exit with a "
-               "warning and the exit code 1")
-p.add_argument('-v', '--verbose', action='store_true',
-               help='Print detailed stats')
-args = p.parse_args()
-
+def args_parse():
```
TheFRedFox, line 33: For almost all scripts we have we go for the function name `parse_args`. No blocker, but if you have the time, change it, so we are consistent throughout the scripts. ^^

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,177 @@
...
+
+            # Fetch all keys on the project
+            dsns = get_dsns_from_project(
+                args.api, args.organization, project['slug'], args.bearer)
+
+            # Fetch rateLimits for each key and add them to the totals
+            for dsn in dsns:
+
+                if args.verbose:
+                    print(f"  Key: \"{dsn['name']}\",", end=" ")
```
TheFRedFox, line 102:
```suggestion
                    print(f'  Key: "{dsn["name"]}",', end=' ')
```

TheFRedFox, line 102: What's the reason that it starts with whitespaces here?

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,177 @@
...
+
+                if dsn['rateLimit']:
+                    # Calculate events per minute and add them to the counters
+                    epm = int((dsn['rateLimit']['count'] * 60 /
+                               dsn['rateLimit']['window']))
+                    organization['summed_events'] += epm
+                    team['summed_events'] += epm
+                    if args.verbose:
+                        print(f'limited to {epm} events per mintue')
```
TheFRedFox, line 111:
```suggestion
                        print(f'limited to {epm} events per minute')
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,177 @@
...
+                else:
+                    # Set unlimtied events if no limt is given
+                    team['unlimited_events'] = True
+                    organization['unlimited_events'] = True
+                    if args.verbose:
+                        print('with unlimited events')
```
TheFRedFox, line 117: The whole printing process here is quite confusing. If it works, it works right now, I would say. In general I would highly question if we can't refactor it to first build the message string and then print it out completely. Or have different message templates to choose from at the end.

src/check_sentryio_limits.py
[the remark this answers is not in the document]
TheFRedFox: My idea would be to go with concurrent requests. Which would need quite a rewrite of the script, so nothing for now. With 10 * 10 teams and projects we will make 100 requests, which calculation with 1 sec per request we will end up in 100 seconds per run. So not even 2 minutes. Totally fair enough for a daily run.
Just that we have in mind that this will definitely be a bottleneck candidate while scaling.
Might be interesting to put a comment, maybe even a TODO comment, somewhere in the script to make that clear and easy for a future maintainer to directly jump to it.

src/check_sentryio_limits.py
TheFRedFox: With the newest changes it is clear that it is not possible. Therefore: Resolved.

src/check_sentryio_limits.py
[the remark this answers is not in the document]
raimund-s: added

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+        exit = 1
+        print('WARNING: {} of {} events are configured in total'.format(
+            events, args.globallimit))
+    elif events and exit == 0:
+        print('OK: {} events are configured in total'.format(events))
+    elif not args.globallimit:
+        print('OK: no global event limit')
+    else:
+        exit = 3
+        print('UKNOWN Contidion this shoud not happen')
```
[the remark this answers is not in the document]
raimund-s: gone

src/check_sentryio_limits.py
raimund-s: not for any expected case window will be 60, 300, 3600 ... if there is no rate limit there is neither count nor window

src/check_sentryio_limits.py
raimund-s: done

src/check_sentryio_limits.py
TheFRedFox: Fair enough.

src/check_sentryio_limits.py
```diff
@@ -117,6 +117,7 @@ def traverse_dsns(dsns: List[dict], team: dict) -> None:
                 print(f" Project \"{project['slug']}\", checking for keys")
 
             # Fetch all keys on the project
+            # @TODO use concurrent requests to scale out for tons of projects
```
TheFRedFox, line 120: I see that now for the second time. I actually kinda like the extra `@` in front of `TODO`, however, it would not be consistent to the other scripts.
Do you have something where you've got it from, that it was written with the `@`?

src/check_sentryio_limits.py
raimund-s: changed

src/check_sentryio_limits.py
raimund-s: changed and adapted to daily rates

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,123 @@
...
+                print('WARNING: unlimited events configured for team: {}\
+                    '.format(team['name']))
+            elif args.perteamlimit and team['summed_events'] and \
+                    team['summed_events'] > args.perteamlimit:
+                exit = 1
+                print('WARNING: {} are configure of {} allowed for team: {}\
+                    '.format(team['summed_events'], args.perteamlimit,
+                    team['name']))
+
+    '''Check if global limit is reached'''
```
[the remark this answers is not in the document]
raimund-s: done

src/check_sentryio_limits.py
TheFRedFox: So much better! Thx! 👍🏼

src/check_sentryio_limits.py
```diff
@@ -136,7 +137,7 @@ def traverse_dsns(dsns: List[dict], team: dict) -> None:
                       .format(team['slug']))
             elif team['summed_events'] > args.per_team_limit:
                 exit = 1
-                print('WARNING: {} are configure of {} allowed for team: {}'
+                print('WARNING: {} are configure, but {} allowed for team: {}'
```
TheFRedFox, line 140:
```suggestion
                print('WARNING: {} are configured, but {} allowed for team: {}'
```

src/check_sentryio_limits.py
```diff
@@ -149,8 +150,9 @@ def traverse_dsns(dsns: List[dict], team: dict) -> None:
         # If the organizaion wide limit is hit
         elif organization['summed_events'] > args.organization_limit:
             exit = 1
-            print('WARNING: {} of {} events are configured in total'.format(
-                  organization['summed_events'], args.organization_limit))
+            print('WARNING: {} events are configured, but {} allowd in total'
```
TheFRedFox, line 153:
```suggestion
            print('WARNING: {} events are configured, but {} allowed in total'
```

src/check_sentryio_limits.py
TheFRedFox: That might be better to use with logging. But we will skip it for this PR.

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,191 @@
...
+        for dsn in dsns:
+
+            if args.verbose:
+                print(f'  Key: "{dsn["name"]}",', end=' ')
+
+            if dsn['rateLimit']:
+                # Calculate events per day and add them to the counters
+                rate_count = dsn['rateLimit']['count']
+                rate_window = dsn['rateLimit']['window']
+                rate_daily = int(rate_count * 14400 / rate_window)
```
kofrezo, line 94: What is 14400 ?

It is best practise to avoid "magic numbers". Create a dedicated variable instead. For Example:
```
seconds_per_day = 60 * 60 * 24
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def get_dsns_from_project(api: str, organization_slug: str, project: str,
+                          bearer: str) -> dict:
+    """Return a list of DSNs for the passed project"""
+    headers = {'Authorization': f'Bearer {bearer}'}
+    res = requests.get(
+            f'{api}/0/projects/{organization_slug}/{project}/keys/',
+            headers=headers)
+    return res.json()
```
kofrezo, line 187:
```suggestion
    if res.status_code != 200:
        print(f'Expected HTTP 200 but got {res.status_code} while fetching dsns for project')
        sys.exit(3)  # Nagios code UNKNOWN

    return res.json()
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,191 @@
...
+    sys.exit(exit)
+
+
+def get_teams(api: str, organization_slug: str, bearer: str) -> dict:
+    """Return a list of all teams in the account"""
+    headers = {'Authorization': f'Bearer  {bearer}'}
+    res = requests.get(
+            f'{api}/0/organizations/{organization_slug}/teams/',
+            headers=headers)
+    return res.json()
```
kofrezo, line 177:
```suggestion
    if res.status_code != 200:
        print(f'Expected HTTP 200 but got {res.status_code} while fetching teams')
        sys.exit(3)  # Nagios code UNKNOWN
        
     return res.json()
```

src/check_sentryio_limits.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - sentry.io event limit check
+
+The script will exit with:
+ - 0 (OK) All the limits provided to this check are kept
+ - 1 (WARNING) Limits are either set to high or not at all
```
kofrezo, line 6: Let's not send some poor sysadmin debugging the script when there is network foo or the authentication failed.

I added to check for HTTP status code 200 further below.

```suggestion
 - 1 (WARNING) Limits are either set to high or not at all
 - 3 (UNKNOWN) Error while fetching team or project info via HTTP API
```

kofrezo (changes requested):
Just some minor improvements:
- Please check my comment about the magic number if we can improve that.
- Consider checking the HTTP status code if that makes sense.
- I consider the rest optional.

kofrezo (approved):

PR #250, opened 2022-04-07
check_cpu_steal_time: Compare correct types

seqizz:


src/check_cpu_steal_time.py
```diff
@@ -37,11 +37,13 @@ def get_parser():
     parser = ArgumentParser()
 
     parser.add_argument(
-        '--warning', '-w', help='warning threshold for steal time',
+        '--warning', '-w',
+        help='warning threshold for steal time, int or float',
```
lamaral, line 41: I think you can add `type=float` here to have the parser cast it for you.

src/check_cpu_steal_time.py
```diff
@@ -37,11 +37,13 @@ def get_parser():
...
         default=50
     )
     parser.add_argument(
-        '--critical', '-c', help='critical threshold for steal time',
+        '--critical', '-c',
+        help='critical threshold for steal time, int or float',
```
lamaral, line 46: I think you can add `type=float` here to have the parser cast it for you.

src/check_cpu_steal_time.py
seqizz: oh even better, doing it

lamaral (approved):

PR #251, opened 2022-04-08
check_https_domains: Set and handle socket timeout

lamaral:


src/check_https_domains.py
```diff
@@ -98,10 +104,10 @@ def get_domains(domains):
     return domains
 
 
-def fetch_cert_info(domain, ip, port):
+def fetch_cert_info(domain, ip, port, timeout):
     domain = domain.replace('*', 'www', 1)
 
-    conn = ssl.create_connection((ip, port))
+    conn = ssl.create_connection((ip, port), timeout)
```
TheFRedFox, line 110: In `get_check_result` you are catching the Timeout. Would it make sense to do it here too?

src/check_https_domains.py
```diff
@@ -120,12 +126,16 @@ def fetch_cert_info(domain, ip, port):
...
+def get_check_result(domains, ip, port, timeout):
     output = []
     expirations = []
 
     for domain in domains:
-        expirations.append(fetch_cert_info(domain, ip, port))
+        try:
+            cert_info = fetch_cert_info(domain, ip, port, timeout)
+            expirations.append(cert_info)
+        except TimeoutError:
```
TheFRedFox, line 137: I don't see no import for the `TimeoutError`. I doubt that it is something so builtin. Is the import maybe missing?

lamaral: TimeoutError is builtin. It's in the docs.

TheFRedFox: You are right, nevermind then. ^^

src/check_https_domains.py
lamaral: Catching the timeout here would require a rewrite of many parts of the check. 
Additionally, it would not break the loop on the first timeout. All connections are towards the same server. If the first connection timed out, it's safe to assume the other connections will timeout as well.

src/check_https_domains.py
```diff
@@ -60,6 +60,10 @@ def parse_args():
         '-p', '--port', type=int, default=443,
         help='Port on the host where Nagios will connect to. Defaults to 443.'
     )
+    parser.add_argument(
+        '-t', '--timeout', type=int, default=2,
+        help='Socket timeout for each SSL connection'
```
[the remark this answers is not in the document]
lamaral: Fixed.

src/check_https_domains.py
TheFRedFox: Fair enough.

TheFRedFox (approved):

PR #253, opened 2022-04-19
Check if all passed teams exists

kofrezo:


raimund-s (approved):

src/check_sentryio_limits.py
```diff
@@ -75,6 +75,10 @@ def main():
             teams
         ))
 
+        if len(args.teams) != len(teams):
+            print(f"Could not find all teams: {args.teams}! Typo ?")
```
TheFRedFox, line 79: Actually, the exit code prefix is missing here:
Shouldn't it be:
``` python
print(f"UNKNOWN: Could not find all teams: {args.teams}! Typo ?")
```

kofrezo: Fixed in https://github.com/innogames/igmonplugins/pull/254

PR #263, opened 2022-07-12
check_smartmon: Allow files starting with manufacturer name

seqizz:
New hardware has files starting directly with string INTEL

### DIFF
--- a/src/check_smartmon.py
+++ b/src/check_smartmon.py
@@ -64,7 +64,7 @@ for manufacturer in HDD_NAMES:
             exit_status = 1
         continue
 
-    for csvfile in glob.glob("*" + manufacturer + "*.csv"):
+    for csvfile in glob.glob(".*" + manufacturer + "*.csv"):
         with open(csvfile) as myfile:
             csv_last_line = (list(myfile)[-1])
         csv_array = csv_last_line.split("\t")


src/check_smartmon.py
```diff
@@ -64,7 +64,7 @@
             exit_status = 1
         continue
 
-    for csvfile in glob.glob("*" + manufacturer + "*.csv"):
+    for csvfile in glob.glob(".*" + manufacturer + "*.csv"):
```
TuxPowered42, line 67: The change says allow files starting with manufacturer name, but the change is files starting with dot and any string and manufacturer name. This is supposed to be glob, not regexp.

TuxPowered42 (changes requested):

src/check_smartmon.py
seqizz: Hmm, it doesn't act like glob though. In glob * means zero or more chars, which should've been OK in our case, but it is not :thinking:
I wonder if it is suddenly counting `./INTEL...` because of this change.
In any case, please advise, we need a change to grab strings starting with INTEL, and the current glob doesn't work for some reason.

TuxPowered42:
```
[13:24:17] positron ~/ # mkdir test
[13:24:21] positron ~/ # cd test
[13:24:22] positron test/ # touch INTEL
[13:24:24] positron test/ # ls *INTEL
INTEL
[13:24:26] positron test/ # ls .*INTEL
zsh: no matches found: .*INTEL
[13:24:31] positron test/ # ls ./*INTEL
./INTEL
```

PR #264, opened 2022-07-12
check_zfs: Request explicit columns from zpool command

seqizz:
New version seems to be returning one more column, which we don't use and it breaks the check.

### DIFF
--- a/src/check_zfs.py
+++ b/src/check_zfs.py
@@ -23,14 +23,14 @@ Copyright (c) 2020 InnoGames GmbH
 
 import argparse
 from subprocess import check_output, STDOUT, CalledProcessError
-import platform
-import re
 import sys
 
 
 def main():
     parser = argparse.ArgumentParser(
-        description=('Check raid state, fragmentation and utilisation of ZPools'),
+        description=(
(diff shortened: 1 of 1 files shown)


src/check_zfs.py
```diff
@@ -65,16 +65,20 @@ def main():
     args = parser.parse_args()
 
     try:
-        proc = check_output(
-            ['/sbin/zpool', 'list', '-H'], stderr=STDOUT
+        proc = check_output([
+            '/sbin/zpool',
+            'list',
+            '-H',
+            '-o', 'name,size,alloc,free,expandsz,frag,cap,dedup,health,altroot'
```
TuxPowered42, line 72: Not all of those columns are needed. Look at parse_zpool():
```
name = data[0]
frag = data[5]
cap = data[6]
health = data[8]
```

We can clean up the code and get rid of unused parameters.

TuxPowered42 (changes requested):

src/check_zfs.py
seqizz: Alright, pushed.

TuxPowered42 (approved):

PR #265, opened 2022-07-15
bird_protocols: Allow protocols to be ignored

lamaral:


TuxPowered42 (approved):

src/check_bird_protocols.py
```diff
@@ -36,21 +37,28 @@ def parse_args():
...
+        default=None,
+        help='Regular expression that matches protocols that should be ignored'
+    )
     return parser.parse_args()
 
 
 def main():
     args = parse_args()
     birdc_path = args.path
+    ignore_regex = re.compile(args.ignore_regex) if args.ignore_regex else None
```
TheFRedFox, line 53: First: Not a blocker.

I was wondering if it would make more sense to have the compilation of the regex string inside the `check_birdc_protocols` function. Especially as the parameter is called `ignore_regex` it sounds more like the raw string than the compiled object. For me it would also make more sense.
If you can agree with that, I would like to see the compilation inside the `check_birdc_protocols` function. :)

TheFRedFox (approved):
Early approval, as there are just 2 discussion points.

src/check_bird_protocols.py
lamaral: I moved the compilation inside `check_birdc_protocols` and gave a more meaningful name there. 
The if grew a little bigger, so I had to split it in two lines. Please check if the style is decent there.

PR #266, opened 2022-08-11
check_smartmon: Skip empty CSV files

seqizz:
Turns out we might have empty files generated. This is a simple check to pass in case.

### DIFF
--- a/src/check_smartmon.py
+++ b/src/check_smartmon.py
@@ -66,7 +66,12 @@ for manufacturer in HDD_NAMES:
 
     for csvfile in glob.glob("*" + manufacturer + "*.csv"):
         with open(csvfile) as myfile:
-            csv_last_line = (list(myfile)[-1])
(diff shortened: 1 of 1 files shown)


src/check_smartmon.py
```diff
@@ -66,7 +66,12 @@
 
     for csvfile in glob.glob("*" + manufacturer + "*.csv"):
         with open(csvfile) as myfile:
-            csv_last_line = (list(myfile)[-1])
+            # Get a snapshot in case file changes in the meantime
+            myfile_list = list(myfile)
```
TuxPowered42, line 70: The file can still change during this reading into a list. If you want to skip empty files the best idea would be to list the directory including file stats and only operate on files with size > 0. They still can change while being read, but at list you would properly exclude the 0-sized ones. Unless whatever deploys those files can do it in atomic way (>tmpfile && mv tmpfile realfile).

TuxPowered42 (approved):

PR #268, opened 2022-09-19
Parallelize requests for keys of project

raimund-s:
Not perfect yet but twice as fast and scaling
for more keys per project

old:
real 0m13.538s
user 0m0.620s
sys  0m0.043s

new:
real 0m6.178s
user 0m0.623s
sys  0m0.063s

There is still more room for improvment if teams
also get handled in parallel


seqizz (approved):
LGTM. Should be reasonably good for some time.

PR #272, opened 2022-10-26
Add check_secret for checking service restarts

zados:
check_secret checks if a service was restarted after secret rotation. The secret rotation is found in a timestamp file.

### DIFF
--- a/src/check_secret.py
+++ b/src/check_secret.py
@@ -30,12 +30,11 @@ from sys import exit
 
 
 def parse_args() -> Namespace:
-    parser = ArgumentParser()
-    parser.add_argument('-s', dest='service_name', type=str, required=True)
-    parser.add_argument('-w', dest='warning_days', type=int, required=True)
-    parser.add_argument('-c', dest='critical_days', type=int, required=True)
-    parser.add_argument('-p', dest='timestamp_file_path', type=str,
-                        required=True)
+    parser = ArgumentParser(__doc__)
+    parser.add_argument("-s", dest="service_name", type=str, required=True)
+    parser.add_argument("-w", dest="warning_days", type=int, required=True)
+    parser.add_argument("-c", dest="critical_days", type=int, required=True)
+    parser.add_argument("-p", dest="timestamp_file_path", type=str, required=True)
     return parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


src/check_secret.py
```diff
@@ -0,0 +1,107 @@
...
+    Gets restart time of service via systemctl
+
+    Args:
+        service_name: Systemd Service name
+
+    Returns: Datetime.
+    """
+    date_s = check_output(
+        f"systemctl show {service_name}.service -p ExecMainStartTimestamp",
+        shell=True).decode("utf-8").split('=')[1].rstrip('\n')
```
TuxPowered42, line 53: Please don't run things with `shell=True`.

src/check_secret.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    Gets restart time of service via systemctl
+
+    Args:
+        service_name: Systemd Service name
+
+    Returns: Datetime.
+    """
+    date_s = check_output(
+        f"systemctl show {service_name}.service -p ExecMainStartTimestamp",
```
TuxPowered42, line 52: Please don't run things with shell=True.

src/check_secret.py
TuxPowered42, line 53: To get the value only use `--value`, then you don't need to strip.

src/check_secret.py
TuxPowered42, line 52: Please don't run things with shell=True.

src/check_secret.py
TuxPowered42, line 53: To get the value only use --value, then you don't need to strip.

TuxPowered42 (changes requested):

src/check_secret.py
```diff
@@ -0,0 +1,107 @@
...
+# THE SOFTWARE.
+
+from argparse import ArgumentParser, Namespace
+from datetime import datetime, timedelta
+from subprocess import check_output
+from sys import exit
+
+
+def parse_args() -> Namespace:
+    parser = ArgumentParser()
```
kofrezo, line 33: A description would be nice.

src/check_secret.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    Gets restart time of service via systemctl.
+
+    Args:
+        service_name: Systemd Service name
+
+    Returns: Datetime.
+    """
+    date_s = check_output(
+        ['/bin/systemctl', 'show', f"{service_name}.service", '-p',
```
kofrezo, line 52: No need for double quotes here. f-strings work with single quotes as well. Use either consistent.

kofrezo (approved):
Early approve. Please decide for either double or single quotes and stick to one. I consider the description a nice to have.

TuxPowered42 (approved):

PR #275, opened 2023-01-27
Add check_offline_cpus for offline CPU monitoring

konstantinbest:
For some years we have had the problem that sometimes some CPUs are offline after we have added them with igvm. The problem can sometimes be solved by manually setting the CPU to online, but it is often necessary to restart the VM.

Recently, we noticed that the CPU thread on the hypervisor has 100% load when the CPU is offline in the VM. 

This check allows us to find out which VMs are affected. This enables us to save CPU time on the affected hypervisors and to use all potential CPUs on the affected VMs.

### DIFF
--- a/src/check_offline_cpus.py
+++ b/src/check_offline_cpus.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - Offline CPU check
+
+Copyright (c) 2023 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import subprocess
+import sys
+
+
+def get_cpus():
+    """
+    Get the number of total and offline CPU cores.
+
+    Returns:
+    tuple: A tuple containing the number of total CPU
+           cores and the list of offline CPU cores.
+    """
+
+    # Get the list of offline CPU cores
+    offline_cpus = subprocess.check_output(
+        "lscpu | grep '^Off-line CPU(s) list:' | awk -F: '{print $2}'",
+        shell=True
+    ).strip().decode()
+
+    # Get the number of total CPU cores
+    total_cpus = subprocess.check_output(
+        "lscpu | grep '^CPU(s):' | awk '{print $2}'",
+        shell=True
+    ).strip().decode()
+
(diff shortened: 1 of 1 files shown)


seqizz (approved):
LGTM. I found a problematic host and a normal one to test both outputs.

src/check_offline_cpus.py
```diff
@@ -0,0 +1,61 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import subprocess
+
+
+def check_offline_cpus():
```
kofrezo, line 27: It's more `get_cpus(offline=True)` instead of `check` no ?

Nothing is checked but offline and total cpus are returned.

src/check_offline_cpus.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Check if there are offline CPU cores and return the number
+    of offline CPU cores
+
+    """
+
+    # Get the number of total CPU cores
+    total_cpus = subprocess.check_output(
+        "lscpu | grep '^CPU(s):' | awk '{print $2}'", shell=True
```
kofrezo, line 36: You rely on the PATH variable being set here. You could consider using the full path to each binary.

src/check_offline_cpus.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def main():
+    offline_cpus, total_cpus = check_offline_cpus()
+
+    if offline_cpus:
+        print(f"WARNING - {offline_cpus} CPU(s) are offline")
+        exit(1)
+    else:
+        print(f"OK - All {total_cpus} CPUs are online")
```
kofrezo, line 55: I wouldn't consider this a no-go but in line 49 you are already executing the function to get all CPUs (online and offline) but print them only down here. 

You could use two functions or a parameterized one and call them right before or in the print statement.

src/check_offline_cpus.py
```diff
@@ -0,0 +1,61 @@
...
+
+    return offline_cpus, total_cpus
+
+
+def main():
+    offline_cpus, total_cpus = check_offline_cpus()
+
+    if offline_cpus:
+        print(f"WARNING - {offline_cpus} CPU(s) are offline")
+        exit(1)
```
kofrezo, line 53: Please use sys.exit see [Stackoverflow](https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python#:~:text=exit%20is%20a%20helper%20for,in%20namespace%20(e.g.%20exit%20).)

kofrezo (changes requested):
Please use sys.exit. I consider the other remarks debatable and IMHO style improvements. You may also consider using type hints.

kofrezo:
Useful check btw!

src/check_offline_cpus.py
konstantinbest: Relying on the PATH variable makes the code more readable and maintainable. By using the binary names instead of their full paths, the code is shorter, easier to understand, and less prone to breaking if the location of the executables changes

src/check_offline_cpus.py
konstantinbest: I changed it a bit, please check if it is now better.

kofrezo (approved):

src/check_offline_cpus.py
```diff
@@ -0,0 +1,66 @@
...
+    Get the number of total and offline CPU cores.
+
+    Returns:
+    tuple: A tuple containing the number of total CPU
+           cores and the list of offline CPU cores.
+    """
+
+    # Get the list of offline CPU cores
+    offline_cpus = subprocess.check_output(
+        "lscpu | grep '^Off-line CPU(s) list:' | awk -F: '{print $2}'",
```
TuxPowered42, line 39: There's no need to use both grep and awk.

src/check_offline_cpus.py
```diff
@@ -0,0 +1,66 @@
...
+        shell=True
+    ).strip().decode()
+
+    # Get the number of total CPU cores
+    total_cpus = subprocess.check_output(
+        "lscpu | grep '^CPU(s):' | awk '{print $2}'",
```
TuxPowered42, line 45: There's no need to use both grep and awk.

src/check_offline_cpus.py
TuxPowered42: To be honest there's no need to use shell for this at all. Handling such simple string splitting should be trivial in Python.

PR #276, opened 2023-02-08
Add support for file_paths in check_secret.

zados:
Not only services may need to be checked, but also cache files.

### DIFF
--- a/src/check_secret.py
+++ b/src/check_secret.py
@@ -2,8 +2,10 @@
 """InnoGames Monitoring Plugins - check_secret
 
 Check if service was restarted after timestamp in secret_timestamp file.
+OR
+Check if file is older than timestamp in secret_timestamp file.
 
-Copyright (c) 2022 InnoGames GmbH
+Copyright (c) 2023 InnoGames GmbH
 """
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


src/check_secret.py
```diff
@@ -91,30 +101,47 @@ def get_time_delta(service_time: datetime,
...
     args = parse_args()
-    delta = get_time_delta(
-        get_service_restart_time(args.service_name),
-        get_secret_file_time(args.timestamp_file_path),
+    if args.service_name:
+        check_delta=get_service_restart_time(args.service_name)
+        warning_name = args.service_name
+    else:
+        check_delta=get_file_mtime(args.file_path)
+        warning_name = 'Server'
```
TuxPowered42, line 126: It's rather if service name … else if file path … else fail "tell me what to check".

src/check_secret.py
```diff
@@ -91,30 +101,47 @@ def get_time_delta(service_time: datetime,
...
+    return datetime.fromtimestamp(path.getmtime(file_path))
 
 
 def main():
     args = parse_args()
-    delta = get_time_delta(
-        get_service_restart_time(args.service_name),
-        get_secret_file_time(args.timestamp_file_path),
+    if args.service_name:
+        check_delta=get_service_restart_time(args.service_name)
```
kofrezo, line 122: A little space for the equals sign here and in the following lines please. It's hard for it to breathe ;-)

kofrezo (approved):
Early approve as my comment only remarks style issue but please adjust them before merging.

src/check_secret.py
```diff
@@ -91,30 +101,47 @@ def get_time_delta(service_time: datetime,
...
+    """
+    return datetime.fromtimestamp(path.getmtime(file_path))
 
 
 def main():
     args = parse_args()
-    delta = get_time_delta(
-        get_service_restart_time(args.service_name),
-        get_secret_file_time(args.timestamp_file_path),
+    if args.service_name:
```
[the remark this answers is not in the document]
zados: That is automatically handled by argparse.

PR #277, opened 2023-02-13
Set correct file permissions on offline cpu check.

konstantinbest:
Fix wrong file permissions


seqizz (approved):
I don't understand why do we have different permissions for different files, all should be called by python binary but lgtm.

PR #287, opened 2023-07-03
Refactor check_systemd_units to make it faster

Woellchen:
When checking all units it finishes around ~4 times as fast and when checking a specific service unit it is even ~7 times as fast now (at least on my test VMs).
Unfortunately the D-Bus API does not provide a way to fetch multiple units at once, that is why `systemctl` is used here - which is slow, but not as slow as calling D-Bus in a loop. The `Dump` call is only meant for debugging and does not provide stable and complete information.


seqizz (approved):
Tested on few hosts which were painful due to many hardware-related units, works good.

thank_you_office

PR #289, opened 2023-08-01
Add check for CRL expiration

lamaral:


lamaral:
Turns out `check_ssl_cert` already has support for CRL, so no need for a new check.

PR #292, opened 2023-08-16
Handle redis masters with no or multiple slot ranges

Woellchen:
This fixes a crash when masters have multiple slot ranges. Also it adds a warning for masters without any slots assigned.


Phantuman (approved):
LGMT ... again

PR #293, opened 2023-08-28
Tiny fixes in mysql checks

v0devil:
1. src/check_mysql_table_status.py: MySQL 8.0 introduced the auto_increment field on mysql.host table where PK datatype is Char (??). Just skip such tables if key_datatype is not known.
2. src/check_mysql_conf.py: Empty files with only [mysqld] line are producing an error in pt-config-diff - checking if there is at least one non-empty line in the config file.


src/check_mysql_table_status.py
```diff
@@ -146,7 +146,10 @@ def get_messages(database, filter_tables, ignore_tables, attributes, outputs):
                         'We don\'t know how to run relative limits on "{}"'
                         .format(attribute)
                     )
-                value.scale(database.get_primary_key_datatype(table))
+                key_datatype = database.get_primary_key_datatype(table)
+                if not key_datatype:
+                    continue
```
Woellchen, line 151: The `mysql.host` table was removed a long time ago in 5.6. It should not be there in the first place. Lets revert this `continue` and fix the database instead

src/check_mysql_conf.py
```diff
@@ -60,6 +60,18 @@ def parse_args():
 
     return parser.parse_args()
 
+def check_config_file(conf_fd):
```
kofrezo, line 63:
```suggestion
def check_config_file(filename):
    # Import configparser module on top
    config = configparser.ConfigParser()
    config.read_file(conf_fd)
    return config.has_section('mysqld')
```

kofrezo (changes requested):
You can use the [configparser](https://docs.python.org/3.9/library/configparser.html) module here instead of building your own code :-)

PR #301, opened 2024-01-23
Add check for ambiguous apt package sources

Woellchen:
Monitoring script which checks that every known
package has only one available source to choose from

### DIFF
--- a/src/check_apt_ambiguous_source.py
+++ b/src/check_apt_ambiguous_source.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - APT ambiguous package sources
+
+Checks that every package has only one available source to choose from.
+
+Copyright (c) 2024 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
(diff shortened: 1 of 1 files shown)


src/check_apt_ambiguous_source.py
```diff
@@ -0,0 +1,108 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+
+import subprocess
+import sys
+import typing
+import re
+
+SOURCE_REGEX = re.compile(r'^\s+[0-9]+\s(\S+\s\S+\s\S+\s\S+)$')
```
TuxPowered42, line 32: Please add an example of a line matched by this.

src/check_apt_ambiguous_source.py
```diff
@@ -0,0 +1,108 @@
...
+
+    return ambiguous_sources
+
+
+def get_apt_cache_policy_output() -> str:
+    policy_cmd = subprocess.run(
+        ['apt-cache', 'policy', '.'],
+        check=True,
+        stdout=subprocess.PIPE,
+    )
```
TuxPowered42, line 101: The exception should be caught and Nagios error code 3 should be returned.

TuxPowered42 (changes requested):

src/check_apt_ambiguous_source.py
TuxPowered42: Why parse text output instead of using the `apt` module?

```
import apt

for pkg in apt.cache.Cache():
    if not pkg.is_installed:
        continue
    print(f'Package: {pkg}')
    for origin in pkg.installed.origins:
        if origin.archive == 'now':
            continue
        print(f'    Source: {origin.site}/{origin.archive}')
```

Woellchen: Because `python3-apt` pulls a lot of other dependencies into the mix and it would require future maintenance for breaking changes and possibly supporting multiple versions of it. I wanted a lightweight approach

TuxPowered42 (approved):

PR #305, opened 2024-03-12
Add seaweedfs check

seqizz:
This is the initial check for seaweedfs, only checking few things which (I assume) is important. We can add more if needed.


Phantuman (approved):
LGTM

PR #307, opened 2024-03-25
Fix type error in join()

seqizz:
Looks like we can't do `str.join(list[int])` in Python, so we create a `list[str]`


kofrezo (approved):
Makes sense.

PR #310, opened 2024-05-17
Use first of the matching HDD_NAMES for a disk

raimund-s:
We want to allow for multiple values to match since we only need one specific model to have different values. For this the first match should be applied so we can use the order to have more specific model defined while keeping a general fallback.

### DIFF
--- a/src/check_smartmon.py
+++ b/src/check_smartmon.py
@@ -62,7 +62,8 @@ except OSError as e:
         exit_status = 1
     sys.exit(exit_status)
 
-# find all csvfiles and match a HDD_NAME and use the first one
+# find all csvfiles and match a HDD_NAME, only the first match will be used
+# ordering in the config file matters now
 files = {}
 for csv in glob.glob("*.csv"):
     for model in HDD_NAMES:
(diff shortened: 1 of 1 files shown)


src/check_smartmon.py
```diff
@@ -52,116 +52,124 @@
...
 
-    try:
-        os.chdir(CSV_PATH)
-    except OSError as e:
-        senddata.append(str(e) + ' ')
+try:
+    os.chdir(CSV_PATH)
+except OSError as e:
+    print('Could not find {}'.format(CSV_PATH))
+    senddata.append(str(e) + ' ')
```
lamaral, line 61: Appending to senddata here is not needed, because we exit on line 64 before printing anything.

src/check_smartmon.py
```diff
@@ -52,116 +52,124 @@
...
+    if exit_status < 1:
+        exit_status = 1
+    sys.exit(exit_status)
+
+# find all csvfiles and match a HDD_NAME and use the first one
+files = {}
+for csv in glob.glob("*.csv"):
+    for model in HDD_NAMES:
```
lamaral, line 69: Please add a comment here that the ordering of HDD_NAMES in the configuration matters now. I hope to prevent some head-scratching in the future.

lamaral (changes requested):

lamaral (approved):

PR #315, opened 2024-06-25
new check_macos_update.py check

brainexe:
Warns if there are pending recommended ones for current major version

example: 
```
WARNING - Recommended updates are available for 13.4.1: Title: macOS Ventura 13.6.7, Version: 13.6.7, Recommended: YES
````

### DIFF
--- a/src/check_macos_update.py
+++ b/src/check_macos_update.py
@@ -30,18 +30,20 @@ from typing import List
 
 
 def main():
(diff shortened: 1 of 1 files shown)


kofrezo (approved):
Looks good. Only minor thing I noticed is that you sometimes use single quotes and sometimes double quotes. I would recommend to just once use ruff or black to format the file.

PR #317, opened 2024-06-27
check_macos_update.py: speedup and suggest update command

brainexe:
- only scan for MacOS updates
 - print more details and update command


kofrezo:
Whoops must have missed that one.

kofrezo (approved):

brainexe:
np, I also don't have it on my radar anymore.

PR #318, opened 2024-07-30
Add folder writability check

seqizz:
Generic but not-so-clever check to test writability of a folder.
Explicitly needs most of the parameters since this is originally intended to check Seaweedfs' mounted folders to be writable or not.

### DIFF
--- a/src/check_folder_writability.py
+++ b/src/check_folder_writability.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+#
+# InnoGames Monitoring Plugins - Check folder writeability
+#
+# Copyright © 2024 InnoGames GmbH
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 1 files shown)


kofrezo:
Maybe I am missing your case but can't you just for example:
```
>>> import os
>>> os.access("/home/kofrezo/Desktop", os.W_OK)
True
>>> os.access("/home/kofrezo/Desktop", os.X_OK)
True
```

seqizz:
I basically could not trust what remote filesystem reports to OS vs what is actually possible. Otherwise yeah, it'd be just that :(
Write is done in usual way but read is validated by an arbitrary command (in current case `weed filer.cat`, but might be anything like a call to an API of local app etc.)

src/check_folder_writability.py
```diff
@@ -0,0 +1,135 @@
...
+            [
+                'Check folder writeability for Nagios.',
+                '',
+                'This script writes a random word to a file in a directory and then reads it back.',
+                'The read word is compared to the original word.',
+                '',
+                'It wants all options explicitly and does not assume file paths etc. since it',
+                'can be used in environments with network mounts and other special cases.',
+            ]
+        ),
```
kofrezo, line 56:
```suggestion
        description="""
Check folder writeability for Nagios.

This script writes a random word to a file in a directory and then reads it back.
The read word is compared to the original word.

It wants all options explicitly and does not assume file paths etc. since it
can be used in environments with network mounts and other special cases."""
),
```

src/check_folder_writability.py
```diff
@@ -0,0 +1,135 @@
...
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
+    parser.add_argument(
+        '--external-cmd', required=True, help='The command to confirm the written data.'
```
kofrezo, line 60: If arguments are not optional you can also remove the `--` and remove the `required=True` then they are positional which are required by default.

src/check_folder_writability.py
```diff
@@ -0,0 +1,135 @@
...
+        default=False,
+    )
+    return parser.parse_args()
+
+
+def main():
+    args = parse_arguments()
+
+    if not os.path.isdir(args.directory):
+        print('UNKNOWN - Directory does not exist')
```
kofrezo, line 87:
```suggestion
        print(f'UNKNOWN - Directory "{args.directory}" does not exist')
```

src/check_folder_writability.py
```diff
@@ -0,0 +1,135 @@
...
+        sys.exit(3)
+
+    if args.enforce_different_fs:
+        if not is_on_different_filesystem(args.directory):
+            print('CRITICAL - Directory is not mounted')
```
kofrezo, line 92:
```suggestion
            print(f'CRITICAL - Directory "{args.directory}" is not mounted')
```

src/check_folder_writability.py
seqizz: Yeah, but it's not looking explicit enough imho, with strict ordering and need to check what comes in which order each time. For this one I'll pass :crossed_fingers:

kofrezo (approved):

kofrezo:
Approved it. None of my comments is blocking and I think you applied those you want.

PR #319, opened 2024-08-01
Add redis difference check

seqizz:
When a config change explicitly set to not restart the service, this can help finding out configuration differences between running system and redis.conf.

### DIFF
--- a/src/check_redis_differences.py
+++ b/src/check_redis_differences.py
@@ -67,9 +67,9 @@ def parse_redis_config(file_path: str) -> dict[str, Union[str, list[str]]]:
                         config[current_key].append(line)
                     current_key = parts[0] if parts else None
         return config
-    except Exception as e:
+    except (FileNotFoundError, PermissionError, IOError, OSError) as error:
         # Maybe permission error? But let's catch more broadly
-        print(f'UNKNOWN - Error: {str(e)}')
+        print(f'UNKNOWN - Error: {error}')
         sys.exit(UNKNOWN)
 
 
@@ -79,14 +79,14 @@ def normalize_value(value: Union[str, list[str]]) -> str:
     """
     if isinstance(value, list):
         return ' '.join(value)
-    return str(value).lower()
+    return value.lower()
 
 
 def parse_memory_size(size_str: str) -> Union[int, str]:
     """
     Parses a memory size string and returns the size in bytes or the original string.
     """
-    size_str = str(size_str).upper()
+    size_str = size_str.upper()
     if size_str.isdigit():
         return int(size_str)
 
@@ -219,7 +219,7 @@ def main():
         sys.exit(exit_code)
 
     except Exception as e:
-        print(f'UNKNOWN - Error: {str(e)}')
+        print(f'UNKNOWN - Error: {e}')
         sys.exit(UNKNOWN)
 
 
--- a/src/check_redis_differences.py
+++ b/src/check_redis_differences.py
@@ -1,0 +1,227 @@
+#!/usr/bin/env python3
+#
+# InnoGames Monitoring Plugins - Redis difference check
+#
+# This script checks for differences between a Redis configuration file and a
+# running Redis instance. It compares the configurations based on key-value
+# pairs and reports any discrepancies found.
+#
+# Copyright (c) 2024 InnoGames GmbH
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+                            config[key].append(value)
+                        else:
+                            config[key] = value.strip('"')
+                    elif current_key in multi_line_keys:
+                        config[current_key].append(line)
+                    current_key = parts[0] if parts else None
+        return config
+    except Exception as e:
+        # Maybe permission error? But let's catch more broadly
+        print(f'UNKNOWN - Error: {str(e)}')
```
kofrezo, line 72:
```suggestion
        print(f'UNKNOWN - Error: {error}')
```

src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+                            if key not in config:
+                                config[key] = []
+                            config[key].append(value)
+                        else:
+                            config[key] = value.strip('"')
+                    elif current_key in multi_line_keys:
+                        config[current_key].append(line)
+                    current_key = parts[0] if parts else None
+        return config
+    except Exception as e:
```
kofrezo, line 70:
```suggestion
    except (FileNotFoundError, PermissionError, IOError, OSError) as error:
```

src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+        sys.exit(UNKNOWN)
+
+
+def normalize_value(value: Union[str, list[str]]) -> str:
+    """
+    Joins list values into a single string and converts to lowercase.
+    """
+    if isinstance(value, list):
+        return ' '.join(value)
+    return str(value).lower()
```
kofrezo, line 82: Given your type hints you already expect only strings so the type cast is redundant.
```suggestion
    return value.lower()
```

src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def parse_memory_size(size_str: str) -> Union[int, str]:
+    """
+    Parses a memory size string and returns the size in bytes or the original string.
+    """
+    size_str = str(size_str).upper()
```
kofrezo, line 89: Same here. 
```suggestion
    size_str = size_str.upper()
```

src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+    if size_str.isdigit():
+        return int(size_str)
+
+    multipliers = {
+        'B': 1,
+        'KB': 1024,
+        'MB': 1024 * 1024,
+        'GB': 1024 * 1024 * 1024,
```
kofrezo, line 97: I prefer `1024 ** 3` so you can change to that if you also like it better but feel free to keep this.

src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+    if isinstance(file_parsed, int) and isinstance(running_parsed, int):
+        return OK if file_parsed == running_parsed else CRITICAL
+    # If only one of them is an integer, return UNKNOWN (bug?)
+    if isinstance(file_parsed, int) or isinstance(running_parsed, int):
+        return UNKNOWN
+    # They are strings, compare them case-insensitively
+    if file_parsed.lower() == running_parsed.lower():
+        return OK
+    # When here, means they are different
+    return CRITICAL
```
kofrezo, line 123: Do we really wanna go critical or is a warning maybe enough. I mean it does not require immediate action does ?

src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def compare_client_output_buffer_limit(
```
kofrezo, line 126: How about renaming this to something like `is_client_output_buffer_limit_equal` to make it easier understandable ?

src/check_redis_differences.py
```diff
@@ -0,0 +1,227 @@
...
+                print(diff)
+        else:
+            print(
+                'OK - Redis is running with the declared configuration in the file.'
+            )
+
+        sys.exit(exit_code)
+
+    except Exception as e:
+        print(f'UNKNOWN - Error: {str(e)}')
```
kofrezo, line 222:
```suggestion
        print(f'UNKNOWN - Error: {e}')
```

kofrezo (approved):
Cannot see any obvious bugs or questionable behavior so approving it right away.

Just added some thoughts and suggestions that I think make the code more readable or improve it but can be seen as opinion.

src/check_redis_differences.py
seqizz: As far as this check is concerned, it should be critical imho. Mapping to actual business process might soften it in case of need.

seqizz:
Thanks for the review. I've applied almost all. :rocket:

PR #326, opened 2024-09-19
postgres: Introduce indexes validity check

abeglu:


kofrezo:
We have this check already: 
```
$ mon -c postgres_invalid_indexes_eu1_game
postgres_invalid_indexes_eu1_game - OK - All indexes are valid
```

abeglu:
Currently, we're creating a separate command for each database, but by using this script, we can check multiple databases with a single command.

seqizz (changes requested):
I guess single check is a lot cleaner on monitoring side, so LGTM. But please process this with a formatter like ruff. We should start using f-strings at least at this point.

seqizz (approved):
Tested, sadly (or thankfully) there was no easy way to test actual invalid case.

PR #330, opened 2024-11-14
Lower the fping amount to 9 pings

raimund-s:
A default of 9 pings and thereby 9 seconds makes more sense for most nagios nrpe installations. As the default nrpe timeout is 10s an unreachable host would cause this check to fail with a nrpe timeout rather then the correct output of the unreachble target.


TuxPowered42:
Commit 138bb3ea7edefcb1a7c8a58b9c81823ca552bbce implements this feature already.

PR #332, opened 2025-01-02
Add ceph check

seqizz:
Ref: BOP-49251

Tested on as much fault state as I could trigger without breaking the whole thing.

### DIFF
--- a/src/check_ceph_health.py
+++ b/src/check_ceph_health.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+#
+#  Modified version of check_ceph_health from:
+#  https://github.com/ceph/ceph-nagios-plugins/blob/0010cf88f8a8c8b670f3ecf59631eccb4342168e/src/check_ceph_health
+#
+#  Only supports Ceph Luminous (12) and newer versions.
+#
+#  Copyright (c) 2013-2016 SWITCH http://www.switch.ch
+#  Copyright (c) 2024 Innogames GmbH https://www.innogames.com
+#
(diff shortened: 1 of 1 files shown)


src/check_ceph_health.py
```diff
@@ -0,0 +1,178 @@
...
+STATUS_ERROR = 2
+STATUS_UNKNOWN = 3
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="'ceph health' nagios plugin."
+    )
+    parser.add_argument(
+        '-e', '--exe', help=f'ceph executable [{CEPH_COMMAND}]'
```
kofrezo, line 45: You can use the `default` keyword here to set the default to `CEPH_COMMAND`. Then ou can safe yourself line 72 and do not have to manually add it to the help.

src/check_ceph_health.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+    parser.add_argument(
+        '-s', '--skip-muted', help='skip muted checks', action='store_false'
+    )
+    args = parser.parse_args()
+
+    # validate main executable
+    ceph_exec = args.exe if args.exe else CEPH_COMMAND
+
+    if not os.path.exists(ceph_exec):
```
kofrezo, line 74: Is it worth to also check if it is executable ?

kofrezo (approved):

src/check_ceph_health.py
seqizz: Good call, used shutil.which

PR #337, opened 2025-03-03
Add new check for MySQL InnoDB buffer pool usage

konstantinbest:
Introduced a new monitoring check named `check_mysql_buffer_pool_usage` that calculates the percentage of the InnoDB buffer pool currently being utilized. The check triggers a warning alert if the usage reaches 90%.

### DIFF
--- a/src/check_mysql_buffer_pool_usage.py
+++ b/src/check_mysql_buffer_pool_usage.py
@@ -4,9 +4,9 @@
 This script monitors the usage of the InnoDB buffer pool by issuing a query that calculates
 the percentage of the buffer pool currently utilized. It raises alerts based on defined thresholds.
 
-Warning at 90% usage.
+The Warning threshold is set to 90% by default.
 
-Copyright (c) 2020 InnoGames GmbH
+Copyright (c) 2025 InnoGames GmbH
 """
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -26,20 +26,23 @@ Copyright (c) 2020 InnoGames GmbH
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-from argparse import ArgumentParser
+from argparse import ArgumentParser, Namespace
+from contextlib import closing
+
 from mysql.connector import (
     connect,
     DatabaseError,
 )
 
 
-def parse_args():
-    """Get argument parser -> ArgumentParser"""
+def parse_args() -> Namespace:
+    """Get argument parser"""
 
     parser = ArgumentParser()
     parser.add_argument('--user', default='root')
     parser.add_argument('--host', default='localhost')
-    parser.add_argument('--password', default='password')
+    parser.add_argument('--password', default='')
+    parser.add_argument('--threshold', type=float, default=90.0)
     parser.add_argument(
         '--unix-socket',
         default='/var/run/mysqld/mysqld.sock',
@@ -48,8 +51,8 @@ def parse_args():
     return parser.parse_args()
 
 
-def main():
-    """The main program"""
+def main() -> None:
+    """Main function"""
     args = parse_args()
 
     try:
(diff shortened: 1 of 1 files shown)


src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
...
+
+from argparse import ArgumentParser
+from mysql.connector import (
+    connect,
+    DatabaseError,
+)
+
+
+def parse_args():
+    """Get argument parser -> ArgumentParser"""
```
v0devil, line 37: In general we should use typings in this script, so such comments with -> ArgumentParser could be avoided, or even auto generated

kofrezo, line 37: You are not returning the ArgumentParser but the parsed arguments. What you are actually returning is a [Namespace](https://docs.python.org/3/library/argparse.html#argparse.Namespace) object.

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
...
+# THE SOFTWARE.
+
+from argparse import ArgumentParser
+from mysql.connector import (
+    connect,
+    DatabaseError,
+)
+
+
+def parse_args():
```
kofrezo, line 36:
```suggestion
def parse_args() -> argparse.Namespace:
```
Mind that the import must be added or adjusted.

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
...
+    """Get argument parser -> ArgumentParser"""
+
+    parser = ArgumentParser()
+    parser.add_argument('--user', default='root')
+    parser.add_argument('--host', default='localhost')
+    parser.add_argument('--password', default='password')
```
kofrezo, line 42: Maybe just leave it empty to avoid somebody using a default password such as password.

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
...
+    args = parse_args()
+
+    try:
+        buffer_pool_usage = get_buffer_pool_usage(
+            args.user, args.password, args.host, args.unix_socket)
+    except DatabaseError as error:
+        print('UNKNOWN: ' + str(error))
+        exit(3)
+
+    if buffer_pool_usage >= 90:
```
kofrezo, line 62: Let make the threshold configureable.

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
...
+        host=host,
+        unix_socket=unix_socket,
+    )
+
+    cur = cnx.cursor()
+
+    cur.execute(query)
+
+    res = cur.fetchall()
+    result = res[0][0]  # Expecting a single result
```
kofrezo, line 126:
```suggestion
    res = cur.fetchone() # Error if there is more than one result
    if res is None:
        raise Error("Failed to query buffer pool usage - no results for query!")
```

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
...
+
+    cur.close()
+    cnx.close()
```
kofrezo, line 129: You can use a context manager (or two) here then they will be close for sure and you do not have to take care of it manually.

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
...
+
+    return result
```
kofrezo, line 131: A type hint for the return type could be helpful here.

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - MySQL InnoDB Buffer Pool Check
+
+This script monitors the usage of the InnoDB buffer pool by issuing a query that calculates
+the percentage of the buffer pool currently utilized. It raises alerts based on defined thresholds.
+
+Warning at 90% usage.
+
+Copyright (c) 2020 InnoGames GmbH
```
kofrezo, line 9:
```suggestion
Copyright (c) 2025 InnoGames GmbH

```

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,138 @@
...
+    with closing(connect(**conn_info)) as cnx:
+        with closing(cnx.cursor()) as cur:
+            cur.execute(query)
+            row = cur.fetchone()
+            if row is None:
+                raise ValueError("Query returned no results")
+
+            result = float(row[0])
+
+    return result
```
kofrezo, line 134:
```suggestion
            return float(row[0])
```

src/check_mysql_buffer_pool_usage.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def parse_args() -> Namespace:
+    """Get argument parser"""
+
+    parser = ArgumentParser()
+    parser.add_argument('--user', default='root')
+    parser.add_argument('--host', default='localhost')
+    parser.add_argument('--password', default='')
+    parser.add_argument('--threshold', default=90)
```
kofrezo, line 45:
```suggestion
    parser.add_argument('--threshold', type=float, default=90.0)

```

kofrezo (approved):

PR #342, opened 2025-06-30
Fix oss api call

LindseyDurst:


zados:
Maybe rename `index` to `index_name` for better clarity.

zados (approved):

PR #344, opened 2025-07-29
new check_repmgr command

brainexe:
Add a new check to monitor repmgr cluster and service status, verifying all nodes are connected, repmgrd processes are running, and services are not paused (which would prevent a automated failover). 
Reports CRITICAL for any detected issue, with some output for troubleshooting.


kofrezo (approved):
This will work. I recommend using the CSV module as given by the suggestions it's more concise and robust IMHO.

PR #346, opened 2025-09-11
mysql_innodbcluster: Introduce InnoDB Cluster check

mpeltekis:
This PR introduces a new monitoring check for MySQL InnoDB Cluster.

### DIFF
--- a/src/check_mysql_innodbcluster.py
+++ b/src/check_mysql_innodbcluster.py
@@ -4,12 +4,20 @@
 This script checks the health of a MySQL InnoDB Cluster by examining
 member states, quorum, and replication status with role-aware lag detection.
 
+Authentication: This script is designed to use MySQL socket authentication.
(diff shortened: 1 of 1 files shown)


zados:
- Avoid passing passwords via `--password`, as CLI args are visible via ps. Using a config file (like .my.cnf) would be recommended. Environment variables are an alternative but are already considered deprecated.

- L214: Returning `[]` on failure may mask errors and yield false OK states. Raise the exception (or wrap and re-raise with context) so the check fails loudly.

- L347: Same issue. Also clarify whether cluster_name can legitimately be unknown. Is there any use case for that?

- Use a context manager for the MySQL connection so resources are closed even on exceptions, and to set connect/read timeouts. Setting the timeouts will prevent the check from hanging.

Optional: Add type hints and brief docstrings for functions to clarify variables, types and exceptions.

PR #347, opened 2025-09-18
Introduce Mailgun ESP blocks check

abeglu:


src/check_mailgun_esp_blocks.py
```diff
@@ -0,0 +1,308 @@
...
+        "dimensions": ["domain", "recipient_provider"],
+        "metrics": [
+            "permanent_failed_esp_block_count",
+            "temporary_failed_esp_block_count",
+        ],
+        "pagination": {
+            "limit": 50,
+            "skip": 0,
+            "sort": "permanent_failed_esp_block_count:desc",
+        },
```
kofrezo, line 95: I don't think this endpoint supports pagination: https://documentation.mailgun.com/docs/inboxready/api-reference/optimize/mailgun/metrics/post-v1-analytics-metrics

Only other endpoints specify this as valid field in the request.

src/check_mailgun_esp_blocks.py
```diff
@@ -0,0 +1,308 @@
...
+        "resolution": "hour",
+        "filter": {"AND": []},
+        "include_aggregates": True,
+        "include_subaccounts": False,
+    }
+
+    return json.dumps(payload).encode("utf-8")
```
kofrezo, line 102: What if there is pagination and there are more pages ? You might return an incomplete result here is this okay ?

src/check_mailgun_esp_blocks.py
```diff
@@ -0,0 +1,308 @@
...
+def process_response_data(response_data):
+    """Process API response and calculate totals."""
+    try:
+        # Handle case where response_data might be a list or have different structure
+        if isinstance(response_data, list):
+            items = response_data
+        elif isinstance(response_data, dict):
+            items = response_data.get("items", [])
+        else:
+            raise Exception(f"Unexpected response data type: {type(response_data)}")
```
kofrezo, line 152: Is this a "thing" ? According to the API docs it will always return an JSON object containing an items key (and value) as response.

src/check_mailgun_esp_blocks.py
```diff
@@ -0,0 +1,308 @@
...
+
+        if not items:
+            return 0, 0, []
```
kofrezo, line 155: Same here in the [docs](https://documentation.mailgun.com/docs/inboxready/api-reference/optimize/mailgun/metrics/post-v1-analytics-metrics) this response field is marked as required.

src/check_mailgun_esp_blocks.py
```diff
@@ -0,0 +1,308 @@
...
+
+        total_permanent = 0
+        total_temporary = 0
+        domain_issues = []
+
+        for item in items:
+            # Ensure item is a dictionary
+            if not isinstance(item, dict):
+                continue
```
kofrezo, line 164: I think this is also not strictly necessary. According to the docs it should always return a dictionary/object - any experience that it does not?

src/check_mailgun_esp_blocks.py
```diff
@@ -0,0 +1,308 @@
...
+
+            dimensions_list = item.get("dimensions", [])
+            metrics = item.get("metrics", {})
+
+            # Ensure metrics is a dictionary
+            if not isinstance(metrics, dict):
+                metrics = {}
```
kofrezo, line 171: Same here. It should always be a dictionary/object.

src/check_mailgun_esp_blocks.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # Sort by total blocks descending
+        domain_issues.sort(key=lambda x: x["total"], reverse=True)
+
+        return total_permanent, total_temporary, domain_issues
+
+    except (KeyError, TypeError, AttributeError) as e:
+        raise Exception(
+            f"Error processing response data: {e}. Response type: {type(response_data)}"
+        )
```
kofrezo, line 214: Same here. I would rather say we attempt to implement everything correct and if not let it raise the error with all it's details some sysadmin/developer has to check anyways if it does not work - it's not customer facing.

kofrezo (approved):
Please double check the pagination to not skip results. 

The rest I consider improvements but not strictly necessary so I give you an approval already.

PR #359, opened 2025-12-05
Ignore SEL entries related to booting the OS

TuxPowered42:


src/check_dell.py
```diff
@@ -330,9 +330,20 @@ def check_ipmi_sel(res):
...
+            # Cleaning the SEL.
+            "Log area reset/cleared",
+            "SEL has no entries",
+            # Booting an OS which has iDRAC agent installed.
+            "OEM record",
+            "OS Boot",
+            "OS Critical Stop OS graceful shutdown Asserted",
+        ):
+            if ignore_str in linestr:
+                ignore = True
```
lamaral, line 344: Why loop over all of the strings before making a decision?
Just call continue here and get rid of the ignore variable altogether.

PR #366, opened 2026-04-07
check_alloy: Rewrite to auto-discover all supported components and report drops

konstantinbest:
Example output:

```
OK - 3 component(s) checked | loki_write_default_dropped=0.0;10485760.0;52428800.0;0 loki_write_default_sent=9356.0 prometheus_remote_write_mimir_dropped=0.0;10485760.0;52428800.0;0 prometheus_remote_write_mimir_sent=4000.0 pyroscope_write_default_dropped=0.0;10485760.0;52428800.0;0 pyroscope_write_default_sent=0.0
Component                      Status  Dropped    Sent
-----------------------------  ------  ---------  ------------
loki.write.default             OK      0 bytes    9356 bytes
prometheus.remote_write.mimir  OK      0 samples  4000 samples
pyroscope.write.default        OK      0 bytes    0 bytes
```

### DIFF
--- a/src/check_alloy.py
+++ b/src/check_alloy.py
@@ -1,9 +1,13 @@
 #!/usr/bin/env python3
 """InnoGames Monitoring Plugins - Grafana Alloy Check
 
-This script checks Grafana Alloy metrics to detect when logs are piling up
-instead of being shipped. It monitors the difference between encoded bytes
-(queued) and sent bytes (shipped) via Alloy's Prometheus metrics endpoint.
+This script checks Grafana Alloy metrics to detect when data is being dropped
+instead of being shipped. It automatically discovers all supported component
+instances (loki.write, pyroscope.write, otelcol.exporter.otlp,
+prometheus.remote_write) and reports on each of them.
+
+A state file is used to compute per-interval deltas so that alerts clear as
+soon as drops stop occurring.
 
 Copyright (c) 2026 InnoGames GmbH
 """
@@ -26,16 +30,56 @@ Copyright (c) 2026 InnoGames GmbH
 # THE SOFTWARE.
 
 import argparse
-import sys
-from typing import Dict, List
-
+import json
+import re
 import requests
+import sys
+
+from typing import Dict, List, Optional, Tuple
+
+
+# Mapping from component type prefix to (dropped_metric, sent_metric, unit).
+# dropped_metric may be None if the component does not expose a drop counter.
+COMPONENT_METRICS: Dict[str, Tuple[Optional[str], str, str]] = {
+    'loki.write': (
+        'loki_write_dropped_bytes_total',
+        'loki_write_sent_bytes_total',
+        'bytes',
+    ),
+    'pyroscope.write': (
+        'pyroscope_write_dropped_bytes_total',
+        'pyroscope_write_sent_bytes_total',
+        'bytes',
+    ),
+    'otelcol.exporter.otlp': (
+        'otelcol_exporter_send_failed_spans_total',
+        'otelcol_exporter_sent_spans_total',
+        'spans',
+    ),
+    'prometheus.remote_write': (
+        'prometheus_remote_storage_samples_failed_total',
+        'prometheus_remote_storage_samples_total',
+        'samples',
+    ),
+}
+
+# All metric names we care about across all component types
+ALL_METRICS = {
+    metric
+    for dropped, sent, _ in COMPONENT_METRICS.values()
+    for metric in (dropped, sent)
+    if metric is not None
+}
+
+STATUS_MAP = {0: 'OK', 1: 'WARNING', 2: 'CRITICAL'}
+
+COL_HEADERS = ('Component', 'Status', 'Dropped', 'Sent')
 
 
 def parse_args() -> argparse.Namespace:
     """Parse command-line arguments"""
     parser = argparse.ArgumentParser(
-        description='Check Grafana Alloy metrics for log backlog'
+        description='Check Grafana Alloy metrics for dropped data across all components'
     )
 
     parser.add_argument(
@@ -44,21 +88,21 @@ def parse_args() -> argparse.Namespace:
         help='Alloy metrics endpoint URL (default: http://localhost:12345/metrics)',
     )
     parser.add_argument(
-        '--component',
-        default='loki.write.default',
-        help='Component to monitor (default: loki.write.default)',
-    )
-    parser.add_argument(
         '--warning', '-w',
         required=True,
         type=float,
-        help='Warning threshold in bytes',
+        help='Warning threshold — number of dropped units per interval',
     )
     parser.add_argument(
         '--critical', '-c',
         required=True,
         type=float,
-        help='Critical threshold in bytes',
+        help='Critical threshold — number of dropped units per interval',
+    )
+    parser.add_argument(
+        '--state-file',
+        default='/run/check_alloy_backlog.json',
+        help='Path to state file (default: /run/check_alloy_backlog.json)',
     )
 
     return parser.parse_args()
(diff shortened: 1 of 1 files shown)


src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
-                break
-
-        if not line_has_needed_metric:
+        # Quick pre-filter: skip lines that don't start with a metric we want
+        if not any(line.startswith(m + '{') for m in ALL_METRICS):
             continue
 
         try:
-            # Parse metric line: metric_name{label="value",...} numeric_value
             if '{' not in line:
```
Phantuman: This check is already fullfilled by the change above

src/check_alloy.py
```diff
@@ -26,39 +30,82 @@
...
-        '--component',
-        default='loki.write.default',
-        help='Component to monitor (default: loki.write.default)',
-    )
     parser.add_argument(
         '--warning', '-w',
         required=True,
         type=float,
-        help='Warning threshold in bytes',
+        help='Critical threshold — number of dropped units per interval',
```
Phantuman: This should be warning

src/check_alloy.py
```diff
@@ -26,39 +30,82 @@
...
+# All metric names we care about across all component types
+ALL_METRICS = {
+    metric
+    for dropped, sent, _ in COMPONENT_METRICS.values()
+    for metric in (dropped, sent)
+    if metric is not None
+}
+
+# Reverse map: metric name -> component type
+METRIC_TO_COMPONENT_TYPE = {
```
YannikSc: This seems to be unused

src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+        if not any(line.startswith(m + '{') for m in ALL_METRICS):
             continue
 
         try:
-            # Parse metric line: metric_name{label="value",...} numeric_value
             if '{' not in line:
                 continue
 
             metric_name = line[:line.index('{')]
+            if metric_name not in ALL_METRICS:
```
Phantuman: This should also be already checked above

src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+        worst_code = 0
+        perfdata_parts: List[str] = []
+        first_run_only = True
+
+        # Collect rows: (component_id, status_label, code, dropped_str, sent_str)
+        rows: List[Tuple[str, str, int, str, str]] = []
+
+        for component_id, metrics in sorted(supported_components.items()):
+            ctype = component_type(component_id)
+            assert ctype is not None
```
Phantuman: Asserts could be removed, this should not be used to test buissness logic

src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
-        else:
-            status = 'OK'
-            code = 0
-
-        # Build performance data
-        perfdata_parts = [
-            f'{value_label}={check_value};{args.warning};{args.critical};0',
-            f'encoded={encoded_bytes}',
-            f'sent={sent_bytes}'
+            dropped = metrics.get(dropped_metric, 0.0) if dropped_metric else 0.0
```
YannikSc: I think the `if dropped_metric...` is not needed. 
That should've been already covered by the `.get`

src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+            if not prev:
+                # First time seeing this component — baseline only, no delta
+                rows.append((component_id, 'BASELINE', 0, '-', '-'))
+                continue
+
+            first_run_only = False
+            delta_dropped = max(dropped - prev.get('dropped', 0.0), 0.0)
+            delta_sent = max(sent - prev.get('sent', 0.0), 0.0)
+
+            safe_id = component_id.replace('.', '_').replace('/', '_')
```
YannikSc: Issues might be better prevented by using a set of safe characters, like a regex replace with `/[^a-zA-Z0-9_]+/ => _`

src/check_alloy.py
YannikSc: Assert makes sure, that the code works, but this will still break the script. Maybe a `if .. continue` is good enough for a monitoring script?

src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+                f'{delta_sent:.0f} {unit}',
+            ))
+
+        save_state(args.state_file, new_state)
+
+        if first_run_only:
+            print('OK - Collecting baseline data (first run)')
+            sys.exit(0)
+
+        status_map = {0: 'OK', 1: 'WARNING', 2: 'CRITICAL'}
```
YannikSc: As you already made use of global values, this might also be a good candidate to "promote"

src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+        overall = status_map[worst_code]
+
+        # Build aligned table
+        col_headers = ('Component', 'Status', 'Dropped', 'Sent')
```
YannikSc: Maybe also a global?

src/check_alloy.py
```diff
@@ -80,120 +127,183 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+                code = 1
+                label = 'WARNING'
+            else:
+                code = 0
+                label = 'OK'
+
+            worst_code = max(worst_code, code)
+            rows.append((
+                component_id,
+                label,
```
YannikSc: With the status_map being globally available, you can replace the label with a `STATUS_MAP[code]` to cut down on variables

src/check_alloy.py
Phantuman: The ternary operator is not needed here, as get can already handle None input.

src/check_alloy.py
YannikSc: "Critical"? Why is it warning then?

src/check_alloy.py
```diff
@@ -26,39 +30,82 @@
...
     parser.add_argument(
         '--critical', '-c',
         required=True,
         type=float,
-        help='Critical threshold in bytes',
+        help='Critical threshold — number of dropped units per interval',
+    )
+    parser.add_argument(
+        '--state-file',
+        default='/var/tmp/check_alloy_backlog.json',
```
YannikSc: On our systems this is still hitting the drive. You might want to checkout `/run` or the `$XDG_RUNTIME_DIR` environment variable for a proper place to but temporary files

Phantuman (approved):
LGTM

src/check_alloy.py
```diff
@@ -26,39 +30,79 @@
...
+    ),
+    'prometheus.remote_write': (
+        'prometheus_remote_storage_samples_failed_total',
+        'prometheus_remote_storage_samples_total',
+        'samples',
+    ),
+}
+
+# All metric names we care about across all component types
+ALL_METRICS = {
```
YannikSc: Would be nice to have consistent type hints for all these globals

src/check_alloy.py
```diff
@@ -80,120 +124,173 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+                component_id,
+                STATUS_MAP[code],
+                code,
+                f'{delta_dropped:.0f} {unit}',
+                f'{delta_sent:.0f} {unit}',
+            ))
+
+        save_state(args.state_file, new_state)
+
+        if first_run_only:
```
YannikSc: I'm not sure if this is really necessary. Sure without it the first run might not be okay (probably only relevant for existing servers anyway), but on the second run it's already fine?

The cost of it on the other hand is to think through the whole logic of this `first_run_only` variable. I'm not sure if this is necessary.

src/check_alloy.py
```diff
@@ -80,120 +124,173 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+            max(len(COL_HEADERS[0]), max(len(r[0]) for r in rows)),
+            max(len(COL_HEADERS[1]), max(len(r[1]) for r in rows)),
+            max(len(COL_HEADERS[2]), max(len(r[3]) for r in rows)),
+            max(len(COL_HEADERS[3]), max(len(r[4]) for r in rows)),
         ]
-        perfdata = ' '.join(perfdata_parts)
 
-        # Output result
-        print(f'{status} - {value_label} {check_value:.0f} bytes | {perfdata}')
+        def fmt_row(cols: Tuple[str, str, str, str]) -> str:
```
YannikSc: Do we do inline functions? Also I don't think this is really necessary here. You're not really gaining anything by replacing a single line with a single statement method call

src/check_alloy.py
```diff
@@ -80,120 +124,173 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+        with open(state_file, 'r') as f:
+            return json.load(f)
+    except (FileNotFoundError, json.JSONDecodeError, OSError):
+        return {}
+
+
+def save_state(state_file: str, state: Dict) -> None:
+    """Persist counter state to disk"""
+    with open(state_file, 'w') as f:
+        json.dump(state, f)
```
YannikSc: You do logging at all kinds of places in this script, maybe this should be at least logged, if not actually throwing a proper error if the file could not be saved.

src/check_alloy.py
```diff
@@ -80,120 +124,173 @@ def fetch_metrics(url: str, timeout: int) -> str:
...
+    return result
+
+
+def load_state(state_file: str) -> Dict:
+    """Load persisted counter state from disk"""
+    try:
+        with open(state_file, 'r') as f:
+            return json.load(f)
+    except (FileNotFoundError, json.JSONDecodeError, OSError):
+        return {}
```
YannikSc, line 173: The file might not be found on the first run, but with encoding or other errors, you might want to still print some error message

PR #370, opened 2026-05-19
Allow for ignoring BIRD protocols by protocol type

TuxPowered42:


src/check_bird_protocols.py
```diff
@@ -44,21 +44,29 @@ def parse_args():
         action='append',
         help='Protocols containing this expression will be ignored'
     )
+    parser.add_argument(
+        '-p',
+        '--ignore_proto',
```
lamaral: This is now quite ambiguous with the existing `--ignore`, especially because Bird calls the instances protocols.

I think it would be better to call this one `--ignore_type` and the previous one `--ignore_proto`.

src/check_bird_protocols.py
```diff
@@ -44,21 +44,29 @@ def parse_args():
...
+        help='Protocols of this type will be ignored',
+    )
     return parser.parse_args()
 
 
 def main():
     args = parse_args()
     birdc_path = args.path
     ignored_terms = args.ignore
+    ignored_protos = args.ignore_proto
```
lamaral: Here the ambiguity gets even worse, as we are ignoring a protocol type and not a protocol itself.

lamaral (changes requested):

src/check_bird_protocols.py
TuxPowered42: I've fully renamed the options to `--ignore-name` and `--ignore-type`. Is that better?

src/check_bird_protocols.py
TuxPowered42: See the other comment.

lamaral (approved):

PR #371, opened 2026-05-20
Add check for Redfish sensors

lamaral:
Checks health of Chassis, Sensors and Redundancies as well as threshold values for sensors.

Works for Dell and Supermicro machines.

### DIFF
--- a/src/check_redfish_sensors.py
+++ b/src/check_redfish_sensors.py
@@ -1,0 +1,533 @@
+#!/usr/bin/env python3
+"""InnoGames Monitoring Plugins - Redfish Sensors Check
+
+This script queries the BMC of a Supermicro or Dell server via the Redfish
+HTTP API and reports the health of all sensors as well as power and thermal
+redundancies.
+
+Copyright (c) 2026 InnoGames GmbH
+"""
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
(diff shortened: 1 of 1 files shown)


src/check_redfish_sensors.py
```diff
@@ -0,0 +1,529 @@
...
+    location = sensor.get('PhysicalContext', 'Unknown')
+    name = sensor.get('Name') or sensor.get('Id') or 'sensor'
+    label = sanitize_perfdata_label(f'{location}_{name}')
+
+    # Any unit other than % is replaced with an empty string as Nagios perfdata
+    # can only handle a very specific set of units.
+    perf_unit = '%' if sensor.get('ReadingUnits') == '%' else ''
+
+    thresholds = sensor.get('Thresholds', {})
+    upper_warn = thresholds.get('UpperCaution', {}).get('Reading')
```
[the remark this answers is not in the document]
lamaral: This increases complexity massively and we don't care about the perfdata in Nagios that much anyway.

src/check_redfish_sensors.py
```diff
@@ -0,0 +1,529 @@
...
+    output = summary + '\n\n' + body if body else summary
+    if perfdata:
+        output += '|' + ' '.join(perfdata)
+    return output
+
+
+class RedfishClient:
+    def __init__(self, hostname, username, password, timeout, verify_ssl):
+        self.base_url = f'https://{hostname}'
+        self.username = username
```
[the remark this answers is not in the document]
lamaral: Works with the current check, so we keep as is for simplicity.

src/check_redfish_sensors.py
```diff
@@ -0,0 +1,529 @@
...
+    )
+
+    def handle_sigterm(_signum, _frame):
+        redfish_client.logout()
+        print('UNKNOWN: terminated by signal')
+        sys.exit(ExitCodes.unknown)
+
+    signal.signal(signal.SIGTERM, handle_sigterm)
+
+    try:
```
TuxPowered42, line 118: That's a lot of code in try/except block. Can we have it split into functions, or even just one function, for cleaner code?

src/check_redfish_sensors.py
```diff
@@ -0,0 +1,529 @@
...
+        response = self.session.post(
+            url,
+            json={'UserName': self.username, 'Password': self.password},
+            timeout=self.timeout,
+        )
+        response.raise_for_status()
+        token = response.headers.get('X-Auth-Token')
+        location = response.headers.get('Location')
+        if not token or not location:
+            raise requests.RequestException(
```
TuxPowered42, line 456: Will this exception be properly caught by `main()` and cause an `UNKNOWN` Nagios result?

lamaral: `except requests.RequestException as e:` on line 173 will handle it.

src/check_redfish_sensors.py
lamaral: Sure, I extracted it to a separate function.

TuxPowered42 (approved):

